PR #2, opened 2025-02-20
more configuration options for redis and database

kappa118:
Added ability to configure database hostname and redis database to use


Dispatcharr (approved):
I should’ve done that from the start. Thanks

PR #3, opened 2025-02-22
Adds the ability to run everything in one docker container.

SergeantPanda:
Some more work might need to be done but this should get us started on an entry-point that is compatible with everything being inside one Docker container.


SergeantPanda:
Noticed I hadn't added a launch to celery.

PR #4, opened 2025-02-23
Updated entrypoint

SergeantPanda:
Now migrations will be done on every startup and a default account will be created with a username/password of admin/admin if admin doesn't already exist.


SergeantPanda:
Fixed up permissions so Dispatcharr now runs as the user context specified with Docker. Also added a compose for AIO builds.

Issue #22, opened 2025-03-14
generate_m3u does not respect new channel numbers


neonworker:
https://discord.com/channels/1340492560220684331/1349789334038904964

teisenm:

> There looks to be an issue when changing the Channel Number after a channel has been created. If I create a new channel, its automatically assigns the first channel "Channel 1" if I then load the M3U URL into Emby channel 1 shows up and I am able to open the channel without an issue. If I change the channel number to something like "280" and then try to open channel, 280 shows up in Emby but it gives me a "Playback Error Internal Server Error" Also in log looks like there is a stream mapping issue. The issue may be channel number was changhed to 280 but the stream is still looking for the original channel 1. The Error Log I attached is using a channel that was originally channel 4, but I changed it to 280.

![Image](https://github.com/user-attachments/assets/bab5788c-2c7d-4c42-821d-f49e0704cbd3)

![Image](https://github.com/user-attachments/assets/bb50df25-e084-49f2-bc08-3f91ccb9d121)

I did some digging and I think I found the fault in the "generate_m3u" function (apps/output/views.py).

From my understanding of the code in line 25 
stream_url = request.build_absolute_uri(reverse('output:stream', args=[channel.id]))
the args should not contain channel.id but rather channel_number as it is defined on line 19:
stream_url = request.build_absolute_uri(reverse('output:stream', args=[channel_number]))

Issue #21, opened 2025-03-14
docker-compose mapping wrong directory


neonworker:

neonworker:
I did some digging and I think I found the fault in the "generate_m3u" function (apps/output/views.py).

From my understanding of the code in line 25 
stream_url = request.build_absolute_uri(reverse('output:stream', args=[channel.id]))
the args should not contain channel.id but rather channel_number as it is defined on line 19:
stream_url = request.build_absolute_uri(reverse('output:stream', args=[channel_number]))

Take this with a grain of salt as this my first time reading something made with django, so I might be wrong 🤷
From my understanding the channel numbers in UI are actually correct but the m3u is just generated wrong

SergeantPanda:
This has been updated!

PR #23, opened 2025-03-14
small fixes

neonworker:
fixes for #22 and #21


SergeantPanda:
@dekzter can you look over the changes for views.py. You are more familiar with that code. From what I can tell, we should merge the dockerfile. Those are important changes.

dekzter:
@neonworker can you resolve your conflicts? Also, I think you can revert the views.py change as we've done modifications for how stream URLs are built. Thanks for the PR!

SergeantPanda:
@neonworker I am going to close this as both issues should be resolved now. Unfortunately we were not able to use your pull request do to the conflicts.

PR #74, opened 2025-04-28
Fixed minor typo in UserAgentsTable.jsx

LoudSoftware:


SergeantPanda (approved):
I had to read that a few times to see the problem. Lol.

Good catch - thank you!

PR #114, opened 2025-05-31
add git to enable macos node build to function

Vibemetheus:
Without git in base image the dev build is impossible on macos due to node module sourcing files from git using the base image. I tested by writing extension dockerfile like below which proved adding GIT enabled me to continue moving forward on having a single dev process that works inclusively with apple silicon. This is first step in a few other changes that will be required.

FROM ghcr.io/dispatcharr/dispatcharr:base

RUN apt-get update && apt-get install -y git

Not sure who needs to review and test this, not trying to break anything for windows users.

### DIFF
--- a/docker/DispatcharrBase
+++ b/docker/DispatcharrBase
@@ -16,6 +16,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     libpcre3 libpcre3-dev libpq-dev procps \
     build-essential gcc pciutils \
     nginx streamlink \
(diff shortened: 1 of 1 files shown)


SergeantPanda:
I think instead of adding git to the base, if it is needed for MAC's we should add it to the entrypoint if running in dev mode it can be installed afterwards. We have a few things like that already so it shouldn't take much. 
Otherwise, if we add this to the base, it will be there for all containers.

PR #167, opened 2025-06-08
core: api_views.py: add fallback IP geo provider

maluueu:
- add ip-api.com as fallback geo provider
- fix silent JSONException by parsing only if HTTP 200 OK
- add logger and log error if IP geo can't be fetched

### DIFF
--- a/core/api_views.py
+++ b/core/api_views.py
@@ -1,5 +1,6 @@
 # core/api_views.py
 
+import logging
 from rest_framework import viewsets, status
 from rest_framework.response import Response
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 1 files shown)


core/api_views.py
```diff
@@ -77,14 +80,32 @@ def environment(request):
...
+
+            if r.status_code == requests.codes.ok:
+                geo = r.json()
+                country_code = geo.get("country_code") # e.g. "US"
+                country_name = geo.get("country_name") # e.g. "United States"
+
+            else:
+                # If ipapi.co fails, fallback to ip-api.com
+                # only supports http requests for free tier
+                r = requests.get("http://ip-api.com/json/", timeout=5)
```
aahmyu, line 97: You are not passing the public ip here.

maluueu: @aahmyu not needed, ip-api.com detects it automatically, try cURLing the URL manually.

PR #178, opened 2025-06-11
Fix process communicate() deadlock when epg match data overfill subprocess.PIPE buffer

xham3:
When epg_match returns a large match set, the subprocess stdout pipe gets clogged and prevents the child process from finishing. Reading data from the stdout/stderr fd as they become available will prevent the deadlock and allow auto match to complete for large match sets.

### DIFF
--- a/apps/channels/tasks.py
+++ b/apps/channels/tasks.py
@@ -1,6 +1,7 @@
 # apps/channels/tasks.py
 import logging
 import os
+import select
 import re
 import requests
 import time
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Ooooh  nice find. Do you have a good method to test this?

xham3:
You can trigger the deadlock probably having epg_match print a dummy line that's long enough (probably 64k+ characters). Or you can find a large m3u list and a large epg then do automatch. If the json returned from epg_match is large enough, then it'll deadlock.

SergeantPanda:
Awesome. I've been suspecting it was something like that but haven't had time to check it out. I'll look into this tomorrow and merge it if all is good!

SergeantPanda:
Well I tried to fix it on my end but couldn't. Could you please remove the changes to the issue templates.

xham3:
I rebase on the dev branch. @SergeantPanda what did you try from your end and what didn't work?

SergeantPanda:
Thank you! No sorry I tried to remove the changes to the templates but it didn't work.
I'll try your changes tomorrow but suspect by looking at it I'll be merging at the same time.

PR #256, opened 2025-07-22
Fix for views and rebuild debian script

deku-m:
When running the debian script i had an error when installing and it pointed to views.py by using '' instead of "" creating syntax-error in python. 

Also re-created the script a bit with functions and some checks.
Provided me with running setup on debian.


SergeantPanda:
@deku-m Thank you for the PR. Could you please remove the changes to update_views so it is more targeted to the debian_script?

@OkinawaBoss can you please check this out and make sure you are OK with it since this is your script?

deku-m:
> @deku-m Thank you for the PR. Could you please remove the changes to update_views so it is more targeted to the debian_script?
> 
> @OkinawaBoss can you please check this out and make sure you are OK with it since this is your script?

Sure i can or should i make a new pull request for only the script?
The views.py still gives me a error when installing/updating with script if i dont have the quotes "" at channel-id, will this be fixed?

SergeantPanda:
Sorry I'm not sure why that would stop you from running the script. That is odd. 
You don't need to create a new PR, you can simply remove the change from your branch and it will update the PR accordingly.

SergeantPanda:
I went ahead and made the change to the views as that function isn't used currently anyway.

PR #337, opened 2025-08-30
Update debian_install.sh

n9lvc5:
ran into many error trying to install. fixed code.

### DIFF
--- a/debian_install.sh
+++ b/debian_install.sh
@@ -58,12 +58,16 @@ configure_variables() {
   PYTHON_BIN=$(command -v python3)
   SYSTEMD_DIR="/etc/systemd/system"
   NGINX_SITE="/etc/nginx/sites-available/dispatcharr"
+  ARCH=$(uname -m)
+  if [[ "$ARCH" != "aarch64" ]]; then
+    echo "[ERROR] This script expects an ARM64 system (aarch64), but found $ARCH. The Rollup module may not be compatible."
+    exit 1
+  fi
 }
 
 ##############################################################################
 # 2) Install System Packages
 ##############################################################################
-
 install_packages() {
   echo ">>> Installing system packages..."
   apt-get update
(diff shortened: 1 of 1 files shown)


SergeantPanda:
I'm curious on a couple of things here and excuse my ignorance on it, I did not write this script, have never used it and have only briefly looked at it.

I see that you rolled back to version 20 of nodejs from 23 I am curious as to what issues the newer version was causing.

Also, it looks like you made it so the script only runs on arm64? I just briefly looked at the script but didn't see why it wouldn't work on amd64 so if you could let us know what roadblocks there were specifically I'd appreciate it!

n9lvc5:
Error: Cannot find module @rollup/rollup-linux-arm64-gnu. npm has a bug related to optional dependencies (https://github.com/npm/cli/issues/4828). It was actually there.

The script installs Node.js v23.x via curl -sL https://deb.nodesource.com/setup_23.x | bash -. Since Node.js v23.11.1 is very new, there could be compatibility issues with Rollup or its dependencies. I switched to an older one because it's more stable.

I'm using arm64 instead of amd64 because I'm using it on a Raspberry Pi 5 - Which works great!!!!

pantherale0:
I think perhaps rollup itself needs updating in this project.. I haven't really done much with nodejs, but looking at this, its currently on 4.35.0.. 4.50.0 was released a few days ago and the above NPM issue has been closed already. The installer script works fine on amd64 systems.

https://github.com/Dispatcharr/Dispatcharr/blob/58446fdd331f8031a057c345748a3c1f82a7a4ea/frontend/package-lock.json#L3195

n9lvc5:
Perhaps a new file `debian_install_arm64.sh`

SergeantPanda:
I believe this PR conflicted with another PR and this one was supposed to be closed.

PR #346, opened 2025-09-03
Update views.py

lasharor:
add EPG URL url-tvg


jesmannstl:
Should it be `"{base_url}/epg"` or `"{base_url}/output/epg"`

SergeantPanda:
Thank you for the PR. We don't allow PR's into main though, they need to go into dev first for testing (also, that is where the most recent code is). Could you please rebase with dev to get rid of the conflict with the debian install script?

SergeantPanda (commented):
Got a chance to look at the code and play with it a bit after meging latest changes. Sorry about the delay.

I think we'll add this (it only makes sense, we know the epg exists and where) but I would like to go about it a different way slightly.

We have a different function we use to build the url depending on environment followed by an educated guess if the ips/ports aren't in the headers. We'd use that instead but also, we need to grab the channel profile that the device is accessing it from too.

If you want to do that, let me know, otherwise I can and close this PR.

Thanks again!

lasharor:
I cannot at the moment, do not have access to a pc.

SergeantPanda:
Thank you for your submission. The feature has been added to dev.

PR #361, opened 2025-09-07
Issue 324 dev

crunchingcode:
Removing localhost and replacing with my environment variables.  this is so that all the many services that dispatcharr uses this can be done in its own containers.


SergeantPanda:
Looking through this PR again, I know it's been a while but this is something we need to tackle. Unfortunately I see that it is changing some of the fallback logic to "redis" (or similar) instead of localhost where localhost is a valid fallback if the environmental variables are not populated. This would cause issues for most users.

I do appreciate the effort here and do agree we need to make changes but we need to be very careful with these changes.

PR #374, opened 2025-09-11
Issue # 373 - XC: apply Filter Manager for Xtream ingest

Blarm1959:
On Dispatcharr (Debian LXC), **Filter Manager rules are not applied for Xtream (XC) accounts.
The same filters work correctly for M3U (Standard) accounts.
The Groups UI selection does work for XC.

On checking the code in apps\m3u\tsks.py I can see that the filter code is included in
process_m3u_batch
but not included in
process_xc_category


Blarm1959:
Replaced by #420.  
This new PR is rebased cleanly onto `dev` and contains only the two intended commits.
Closing this one.

Issue #373, opened 2025-09-11
[Bug]: Filters ignored for Xtream (XC) accounts; Groups selection works, M3U filters OK


Blarm1959:
### Describe the bug

On Dispatcharr (Debian LXC), **Filter Manager rules are not applied for Xtream (XC) accounts. 
The same filters work correctly for **M3U (Standard)** accounts. 
The **Groups** UI selection does work for XC.

On checking the code in apps\m3u\tsks.py I can see that the filter code is included in
process_m3u_batch
but not included in
process_xc_category

I will open a PR with the filter code added to process_xc_category which I have tested

### How can we recreate this bug?

Add filters to an Xtream (XC) account
They will not be used

### Dispatcharr Version

v0.8.0

### Docker Version

N/A (Debian LXC under PVE)

### What's in your Docker Compose file?

N/A

### Client Information

Dispatcharr: v0.9.0 (installed via `debian_install.sh`)
Host: Debian LXC under PVE

Blarm1959:
I have create a PR named 'Issue # 373 - XC: apply Filter Manager for Xtream ingest'

PR #420, opened 2025-09-18
Fix: apply XC filters (issue #373) plus fixes for M3U filters in going from v0.8.0 to v0.9.1

Blarm1959:
On Dispatcharr (Debian LXC), **Filter Manager rules are not applied for Xtream (XC) accounts.
The same filters work correctly for M3U (Standard) accounts.
The Groups UI selection does work for XC.

There are 2 commits

Commit 1:
On checking the code in apps\m3u\tsks.py I can see that the filter code is included in
process_m3u_batch
but not included in
process_xc_category

Commit 2:
Also going from v0.8.0 to v0.9.1 the filters were erroring on M3U accounts
This change was made for v0.9.0/v0.9.1 in m3u/tasks.py:
                    if (f.custom_properties or {}).get(
I changed it back to:
                    if json.loads(f.custom_properties or "{}").get(
as i it was giving an error and then it worked again for M3U accounts

A line in M3UFilter.jsx was also changed similarly:
        case_sensitive: (filter.custom_properties || {}).case_sensitive ?? true,
I changed it back to:
        case_sensitive:
          JSON.parse(filter.custom_properties || '{}').case_sensitive ?? true,
as it was giving an error and the case_sensitive was always showing as ON in the GUI form even when set to OFF


SergeantPanda:
Closing after our discussion in Discord.

PR #423, opened 2025-09-18
Add 'UK' region

bigpandaaaa:
I'm not sure if this was intentional, but the UK seems to be missing from the region list.

### DIFF
--- a/frontend/src/constants.js
+++ b/frontend/src/constants.js
@@ -303,6 +303,7 @@ export const REGION_CHOICES = [
   { value: 'tz', label: 'TZ' },
   { value: 'ua', label: 'UA' },
   { value: 'ug', label: 'UG' },
+  { value: 'uk', label: 'UK' },
   { value: 'um', label: 'UM' },
   { value: 'us', label: 'US' },
   { value: 'uy', label: 'UY' },


pantherale0:
It is there, the list seems to be based on ISO3166-1 which makes "UK" an invalid country code. What you are looking for instead is "GB"

https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#GB

SergeantPanda:
I'll need to investigate this to see if we should be switching to UK instead. Thank you for bringing this to our attention.

Megamannen:
> I'll need to investigate this to see if we should be switching to UK instead. Thank you for bringing this to our attention.

That is not how standards works :)

SergeantPanda:
> > I'll need to investigate this to see if we should be switching to UK instead. Thank you for bringing this to our attention.
> 
> That is not how standards works :)

You'd be correct if anything iptv related was a standard. My life would also be a lot easier.

SergeantPanda (approved):
So, this has been a long time but I finally had time to dig into it and can confirm a lot of provider do in fact use .uk in their channel lineups for tvg-ids. I think this is a good thing to add, thank you for the PR!

PR #427, opened 2025-09-19
Add option to add a logo by a URL in Channel editor/creator

csmith1210:
Added an option to upload and set a logo by an external URL. A few things were changed:
1. Refactored the new EPG logo code to reuse for this component
2. Added UI element to set url with separate submit button
3. Had to remove `style={{ flex: 1 }}` from the TextInput stack for the channel name as the spacing was messed up after adding the new UI elements

image

### DIFF
--- a/frontend/src/components/forms/Channel.jsx
+++ b/frontend/src/components/forms/Channel.jsx
@@ -79,6 +79,7 @@ const ChannelForm = ({ channel = null, isOpen, onClose }) => {
   const [selectedEPG, setSelectedEPG] = useState('');
   const [tvgFilter, setTvgFilter] = useState('');
   const [logoFilter, setLogoFilter] = useState('');
+  const [logoUrl, setLogoUrl] = useState('');
 
   const [groupPopoverOpened, setGroupPopoverOpened] = useState(false);
   const [groupFilter, setGroupFilter] = useState('');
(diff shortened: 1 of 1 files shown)


EmeraldPi:
I've also added an option in the batch channel editor to set a logo for multiple channels at once, or remove the logos for multiple channels. I believe these changes bring the logo features in line with services like IPTVEditor and give a better UX.

image

EmeraldPi:
> Can you pull the recent dev into your branch. There were some major changes to the channel edit form since this branch was originally created.

I've pulled the changes into my dev.

> I like the idea of adding bulk channel editing for the logo but I would like to see something a little cleaner for the logo editing in the channel edit form.
>
> Maybe we should rethink the addition of a logo in this area all together and just have a button to add a logo that opens the same form that we use in the logo manager.

Should I split this into two pull requests, one for the batch editing changes and one for the channel editing?

Do you have ideas about what you mean for making the channel edit form cleaner? Do you mean to add a button on the channel edit to pop up the logo manager modal? 

I've been using this to edit 300+ channels and the workflow is pretty simple with how it is now, having to click through to a separate modal would probably be annoying for a large amount of channels.

Also you'd have to add a logo selector to the logo manager modal too if you went that route.

SergeantPanda:
Yeah I having a text field for adding a url that's always visible feels off that's all.

PR #441, opened 2025-09-20
fix: Don't serialize custom_properties for m3u filters

harryyoud:
Serialising results in text strings being stored in the backend database, and then errors like this on sync:

`Error in XC thread batch 0: 'str' object has no attribute 'get'`

### DIFF
--- a/frontend/src/components/forms/M3UFilter.jsx
+++ b/frontend/src/components/forms/M3UFilter.jsx
@@ -52,7 +52,7 @@ const M3UFilter = ({ filter = null, m3u, isOpen, onClose }) => {
       filter ? filter.custom_properties : {},
       'case_sensitive',
       values.case_sensitive,
-      true
+      false
     );
 
     delete values.case_sensitive;


Blarm1959:
I have told SargeantPanda on Discord that this is a problem
See my fix which reverts code back to v0.8.0 (Issue #373, Pull Request #420)
The change made to V0.9.0 and carried forward stops the M3U account  filters running and also in the GUI always shows case_sensitive as ON when it has been set to OFF
That change is in the 2nd commit on the PR

In the first commit in the PR:
I also added code so that XC accounts would use the filters as the filters code was missing form that section

harryyoud:
Right, but the issue described above is that the frontend is serialising the dictionary to JSON string, which my pull request would fix. Then you wouldn't need to JSON parse the result from the database (which should already be a JSON object)

harryyoud:
I've tested my pull request, and it sorts out the case sensitive toggle not being saved as well as allowing filters to run on standard and XC accounts

Blarm1959:
Good I'll remove my 2nd commit code and use your code instead

For the XC accounts if you look in apps/m3u/tasks.py you will see that the compiled_filters code used in function process_m3u_batch_direct for the M3U type accounts is missing from the function process_xc_category_direct which is used for the XC type accounts
My 1st commit just takes the compiled_filters code from process_m3u_batch_direct and adds it into process_xc_category_direct

SergeantPanda:
Thank you for this @harryyoud! 
I did send a patch to dev that fixes this issue.

PR #457, opened 2025-09-23
Allow batch editing channel logos

EmeraldPi:
@SergeantPanda I've split the previous request.

Here is the changes to the batch editor to allow for logo changes.

Adds a section to the batch editor, allows for "no change", "remove logo" (sets to logo_id to null), and logos in logo manager.


SergeantPanda (commented):
Here is what I propose. Slight change to how default looks (shows the default logo)

EmeraldPi:
Looks good to me 👍

PR #458, opened 2025-09-23
Add option to add a logo by a URL in Channel editor/creator

csmith1210:
This pull request revolves around adding an option to set the logo for a channel by a url directly in the channel editor/creator Channel.jsx form.

My original proposal is to simply add textbox for the url to set the logo above the current dropzone. The logo name will be automatically set to the filename.

A second option would be to have a separate button that then links to the Logo.jsx form to set the logo.

image
image

A third option would be to remove the logo portion entirely and have a separate button to bring up the Logo.jsx form. This would require probably creating a second version of this form that includes the existing logo selector so as to not alter the original form for the Logo Manager.

image

### DIFF
--- a/frontend/src/components/forms/Channel.jsx
+++ b/frontend/src/components/forms/Channel.jsx
@@ -11,6 +11,7 @@ import logo from '../../images/logo.png';
 import { useChannelLogoSelection } from '../../hooks/useSmartLogos';
 import useLogosStore from '../../store/logos';
 import LazyLogo from '../LazyLogo';
+import LogoForm from './Logo';
 import {
   Box,
   Button,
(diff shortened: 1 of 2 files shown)


EmeraldPi:
Personally, I like the first option, having a direct url submission in the Channel form. When modifying a large amount of channels, the easily accessible option is nice. Also I think that when creating a channel, it makes sense to present all of the options needed for a channel in one form (including creating a logo).

SergeantPanda:
Yes just using the logo form would be my preferred method. All of the code is already there.

You need to update your branch though to get the new channel edit form before adding it.

EmeraldPi:
Branch is now updated to Dispatcharr:dev and I pushed the changes to make the channel edit form like the second example above to link to Logo.jsx under the logo selector.

SergeantPanda:
This was just what I was looking for. Thank you!

PR #515, opened 2025-10-04
[Enhancement/Fix] Set logo name from url

EmeraldPi:
Previously, the logo form would automatically set the logo name when a file was uploaded, using the file name without the extension. However, when adding a logo via URL, the user had to manually enter a name.

This update adds automatic naming for URL-based logos:
- When the URL input loses focus, the logo name is set based on the URL.
- If the user changes the URL, the logo name updates accordingly.
- Existing behavior for file uploads remains unchanged.


EmeraldPi:
Looks like the checks are failing because of dockerhub credentials by the way.

SergeantPanda (commented):
Looks perfect. Thank you!

PR #530, opened 2025-10-06
Fix: gestion sûre des valeurs 'rating' non numériques dans xc_get_vod…

Lesthat:
…_streams

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -1038,6 +1038,11 @@ def xc_get_vod_categories(user):
 
     return response
 
+def safe_float(value):
+    try:
+        return float(value)
+    except (ValueError, TypeError):
+        return 0.0
 
 def xc_get_vod_streams(request, user, category_id=None):
     """Get VOD streams (movies) for XtreamCodes API"""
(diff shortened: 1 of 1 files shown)


Lesthat:
Pull Request Explanation for Dispatcharr: Safe Conversion of VOD Movie Rating Field
Summary
This pull request fixes an issue where the  /player_api.php?action=get_vod_streams  endpoint returns a 500 Internal Server Error when processing VOD items with non-numeric rating values such as  'n/a' . The failure was caused by an unhandled exception during the conversion of the  movie.rating  field to a floating-point number.
Details
The original code attempted to convert  movie.rating  directly using  float(movie.rating) . However, this raises a  ValueError  when  movie.rating  contains strings that cannot be parsed as floats (e.g.,  'n/a' , empty strings).
Proposed Fix
	•	Introduce a helper function  safe_float(value)  that attempts to convert the input to a float but returns  0.0  if the conversion fails due to a  ValueError  or  TypeError .
	•	Replace the direct conversion with a call to  safe_float(movie.rating)  to ensure robust handling of invalid or missing rating data.
	•	This prevents the exception, allowing the API to respond correctly and maintain service availability.
Code snippet of the fix

def safe_float(value):
    try:
        return float(value)
    except (ValueError, TypeError):
        return 0.0

rating_5based = round(safe_float(movie.rating) / 2, 2)

Impact
	•	Eliminates crashes caused by invalid rating values in VOD metadata.
	•	Improves API stability and user experience when VOD content includes non-standard or missing ratings.
	•	Maintains backward compatibility with valid float rating values.
This PR contributes to the robustness and reliability of Dispatcharr’s VOD streaming feature by handling real-world data inconsistencies gracefully.

SergeantPanda:
Sorry, just saw this and I fixed it a little while ago. I went a different route with the fix. I decided to sanitize the values before adding them to the database that way we don't have extra calculations during every client pull. The VOD endpoint can already get a bit slow to return when lots of VODs are in the database.

If you don't mind testing out the dev branch my fix is live. I would appreciate you letting me know if it fixes it or not. You will need to refresh your vod sources so it fixes the values in the database.

I'll leave this open until I hear back from you. I do appreciate the PR though!

Lesthat:
Hi,
I just tested the fix from the dev branch and it works perfectly!

SergeantPanda:
Thank you for letting me know!

PR #531, opened 2025-10-07
Fix: Add IPv6 CIDR validation in Settings

jordandalley:
Fixed form validation in Settings.jsx to include IPv6 CIDR's

Also changed the new install default from '0.0.0.0/0' to '0.0.0.0/0,::0/0' which is effectively the same thing in both stacks and is a good example of comma separation.

### DIFF
--- a/frontend/src/pages/Settings.jsx
+++ b/frontend/src/pages/Settings.jsx
@@ -94,14 +94,16 @@ const SettingsPage = () => {
   const networkAccessForm = useForm({
     mode: 'controlled',
     initialValues: Object.keys(NETWORK_ACCESS_OPTIONS).reduce((acc, key) => {
-      acc[key] = '0.0.0.0/0';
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you for the PR.  Can you confirm that the backend works correctly with IPv6? I don't have a really good way to test it out at the moment.

jordandalley:
Hey @SergeantPanda I can confirm the backend works perfectly with IPv6 CIDR's. The API has always accepted IPv6 CIDR's. I think I've been running IPv6 on the backend for around 4 months?

SergeantPanda:
Thank you for confirming. I'll play with it a bit later and merge if my testing works. Just want to make sure nothing bad happens with IPV4 as that's what most people use.

jordandalley:
No worries at all :)

SergeantPanda (approved):
Testing appears to be going well. Thank you for your contribution!

PR #594, opened 2025-10-29
Channel name truncation when containing apostrophes or special characters

ghost:
Channel names are being truncated at apostrophes and certain special characters when added to the main playlist.

Examples:

    UK: That's 60s → truncates to UK: That
    UK: 80'S - DIRECT ᴴᴰ → truncates to UK: 80

The bug is in apps/m3u/tasks.py at line 453 in the parse_extinf_line() function. The regex pattern used for parsing M3U EXTINF attributes incorrectly stops matching when it encounters an apostrophe inside quoted values:

    attrs = dict(re.findall(r'([^\s]+)=["\']([^"\']+)["\']', attributes_part))

Updated the regex to properly handle both single and double quotes by matching each quote type separately:

    attrs = dict(re.findall(r'([^\s]+)="([^"]+)"', attributes_part) + re.findall(r"([^\s]+)='([^']+)'", attributes_part))

### DIFF
--- a/apps/m3u/tasks.py
+++ b/apps/m3u/tasks.py
@@ -450,7 +450,7 @@ def parse_extinf_line(line: str) -> dict:
     if len(parts) != 2:
         return None
     attributes_part, display_name = parts[0], parts[1].strip()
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you for this PR! Looks good.

PR #600, opened 2025-10-30
Check for "SERVER_PORT" header before falling back to 9191

lasharor:
I moved on to LXC install on PVE. The x-tvg-url defaults back to 9191 port whilst in reality the port is defined as :80, therefore this obviously brakes this functionality. 

Check request.META.get("SERVER_PORT") before falling back to defaults. This should capture the actual server port from the request.

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -2645,12 +2645,17 @@ def get_host_and_port(request):
     if port:
         return host, port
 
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you for this. I've been meaning to take a look at this for a while.

PR #635, opened 2025-11-10
feat: Configurable DVR Retry Settings

### LINKED ISSUE #474: [Feature]: Configurable DVR Retry Settings (Max Retries & Retry Frequency)

The DVR only attempts to start a recording once - if a channel is offline, the DVR will not attempt again to start a recording. This is an issue for things like team-dedicated sports and event channels where the stream does not go live until the exact moment, or some time later (depending on broadcast and provider delays)

My proposed solution for this would be to potentially add two new fields to the DVR Settings pane.
1. Max Retries
   How many times should the DVR attempt to start recording before reporting failure?
3. Retry Frequency
   How long should the DVR wait between each attempt to start the stream.

Image

### Additional context

Nothing Extra

sethwv:
Closes #474 

*Disclaimer: Much of the python in this PR was written with the assistance of generative AI.*

Add configurable retry system for DVR recordings that fail to start due to offline channels or unavailable streams.

- Backend: retry scheduling with MPEG-TS validation, timezone-aware datetimes
- Database: migration 0017 adds retry settings (default: 3 retries @ 30s intervals)
- Frontend: "Pending Recordings" section with countdown timers, retry config in Settings
- WebSocket: recording_pending_retry event handler for real-time updates
- Cleanup: removes files from failed retry attempts, preserves interrupted recordings

### DIFF
--- a/apps/channels/tasks.py
+++ b/apps/channels/tasks.py
@@ -2062,7 +2062,13 @@ def run_recording(recording_id, channel_id, start_time_str, end_time_str):
         try:
             if temp_ts_path and os.path.exists(temp_ts_path):
                 subprocess.run([
-                    "ffmpeg", "-y", "-i", temp_ts_path, "-c", "copy", final_path
+                    "ffmpeg", "-y",
+                    "-fflags", "+genpts+igndts",  # Regenerate PTS, ignore DTS to fix sync issues
+                    "-i", temp_ts_path,
+                    "-c", "copy",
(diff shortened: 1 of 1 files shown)


sethwv:
Recordings that are scheduled to have begun will briefly enter a new "Pending" group.
image
Streams that successfully start will begin recording and follow the existing flow. If the stream fails to produce good bytes, we will begin scheduling retries and reporting success/failure.
image
image
image

PR #649, opened 2025-11-14
Features added - MAC Support + HTTP Proxy & FailOver + BASIC Auth or M3u

sbeimel:
Failover, if Stream Profil 1 not will work try to run with Profil 2 (if set) otherwise take the backup stream

MAC Integration with proxy support - Handshake + Stream 

M3u & EPG secured by basic Auth

### DIFF
--- a/frontend/src/components/forms/User.jsx
+++ b/frontend/src/components/forms/User.jsx
@@ -174,7 +174,6 @@ const User = ({ user = null, isOpen, onClose }) => {
               description="Used for UI authentication"
               {...form.getInputProps('password')}
               key={form.key('password')}
-              disabled={form.getValues().user_level == USER_LEVELS.STREAMER}
             />
 
             {showPermissions && (


SergeantPanda:
Fail-over deliberately does not utilize profile connections because generally it does not make sense to do so. Dispatcharr is intended to be used as the sole client to a provider so it should have accurate counts of connections used.

Now, I could see an option in the future to allow such a thing for users that really want it but to me this is a low priority.

Also, adding MAC support is OK to me but this seems like a LOT of changes to for that. Some of which don't seem to be even in scope for these features.

Please only submit PR's for 1 feature at a time so they can be properly evaluated and tested.

sbeimel:
ok thanks for feedback. I have different opinion about profile failover. but i can understand  your point. MAC have a lot of changes i guess you can implement it much smoother and faster than me + the proxy topic. all the best anyways BR

SergeantPanda:
Understood. Feel free to submit a targeted PR for MAC support but please make sure it doesn't make unnecessary modifications to other code. 

It had looked like AI had gone a little crazy on some of that PR, not that AI is a bad thing, we all use it for sure but we need to be careful with it as well.

Issue #323, opened 2025-08-21
[Feature]: Username on channel card in Stats section.


agryts:
### Is your feature request related to a problem?

No.

### Describe the solution you'd like

A nice addition would be to have the Username of the connected clients to be shown alongside the IP-address on the Stats channel cards for the active channels.

### Additional context

Nothing Extra

krinkuto11:
Yeah I think that this would be a nice upside to using the XC API instead of sharing the m3u file.

mjcocat:
upvote!

PR #676, opened 2025-11-19
Enhancement: Added User Field to Stream Stats cards

krinkuto11:
This Pull Request introduces the User field in the stats block to better identify streaming sessions.

# User Field (Stats Block) — Implementation Summary

## Backend
- Source: channel statistics API that returns active channel objects containing `clients` arrays (see backend endpoints serving active channel stats).
- Data shape (per-client): each client object includes fields such as `client_id`, `ip_address`, `username` (when available from XC API), `access_type` (e.g., `XC`, `M3U`, `HDHR`), `connected_at`/`connected_since`, and other telemetry (duration, bytes, seeks).
- Responsibility: populate `username` when XC API metadata is available; otherwise return `access_type` or omit username. API returns channel-level data with nested `clients` for frontend aggregation.

## Frontend
- Location: `frontend/src/pages/Stats.jsx`
- Data fetching: `API.fetchActiveChannelStats()` is called by `fetchChannelStats()` (invoked on mount and at polling intervals). Result stored via `useChannelsStore` into `channelStats`.
- Aggregation: `channelStats.channels` is processed into `channelHistory` and `clients` (flattened list) in a `useEffect` — each client object is augmented with its parent `channel` for context.
- Table implementation:
  - Table component: `CustomTable` (wrapper) + table utilities (`useTable`) create `channelClientsTable`.
  - Column definition for "User": in `clientsColumns` within `Stats.jsx` — `accessorFn` returns:
    - `username` if present,
    - otherwise `access_type` (M3U/HDHR/etc.),
    - otherwise `'Unknown'`.
  - `cell` renderer wraps the result in a `Tooltip` and a `Text` element; tooltip message adjusts depending on access type (`XC API User: <username>`, `M3U Playlist Access`, `HDHomeRun Access`, or `'Unknown Access Type'`).
- Styling & behavior: the table cell uses `size="xs"` text and is placed into the `CustomTable` which handles sizing/scrolling. The `User` cell is read-only display only (no inline editing).

## Notes / Rationale
- Backend provides canonical client metadata (username only when available); frontend chooses a safe fallback order (username → access_type → 'Unknown').
- UI focuses on clarity: tooltip provides context about how the user value was determined.
- To change labeling/behavior adjust `accessorFn` and `cell` in `frontend/src/pages/Stats.jsx` and/or enhance backend to always resolve usernames.

Screenshot 2025-11-20 at 00 06 25

Addresses #323

### DIFF
--- a/frontend/src/pages/Stats.jsx
+++ b/frontend/src/pages/Stats.jsx
@@ -284,13 +284,13 @@ const VODCard = ({ vodContent }) => {
       padding="md"
       radius="md"
       withBorder
-      style: {
-        color: '#fff',
-        backgroundColor: '#27272A',
-        maxWidth: '760px',
-        width: '100%',
-      }
-    >
+        style={{
+          color: '#fff',
+          backgroundColor: '#27272A',
+          maxWidth: '760px',
+          width: '100%',
+        }}
+      >
       <Stack style={{ position: 'relative' }}>
         {/* Header with poster and basic info */}
         <Group justify="space-between">
@@ -1298,7 +1298,11 @@ const ChannelCard = ({
           </Group>
         </Group>
 
-        <CustomTable table={channelClientsTable} />
+        <Box style={{ width: '100%', overflowX: 'auto', display: 'flex', justifyContent: 'center' }}>
+          <Box style={{ width: '100%' }}>
+            <CustomTable table={channelClientsTable} />
+          </Box>
+        </Box>
       </Stack>
     </Card>
   );


SergeantPanda:
This is a well done PR and we do appreciate it. 

Unfortunately, we have decided not to allow user information in the api. This is to prevent Dispatcharr from being used for resellers (which is against our license)

We may revisit having this information in the stats in the future but this is our current stance. It has been deliberately left out.

PR #701, opened 2025-11-30
Added IPv6 port bind to nginx.conf

jordandalley:
Updates the default nginx.conf to include portbind for IPv6


SergeantPanda (commented):
Looks good. Thank you!

PR #725, opened 2025-12-09
Removing unreachable code

DawtCom:
## Summary
   Removed unreachable code for code cleanliness
## Testing
   Code ran locally with breakpoints and runs as expected.


SergeantPanda:
Yep good catch, thank you!

PR #750, opened 2025-12-14
Parallel EPG processing experiment

stlalpha:
Parallel EPG Processing Experiment

  I was playing around with ways to speed up EPG program parsing and wanted to see if this was useful.

  Parallel processing: Modified the program parsing to use Celery groups so channels process concurrently. Detects autoscale config and splits work into batches (--autoscale=6,1 creates 6 concurrent batches).

  Database queries: Found N+1 patterns in channel parsing and EPG matching. Added prefetching and in_bulk() to reduce queries from ~16K to ~14.

  Progress tracking: Built a ProgressTracker utility with Redis counters for real-time progress across workers.

  Testing with a 4250-channel EPG:
  - Queries: 16,000 down to 14
  - About 6x faster with multiple workers

  The parallel approach works with the atomic delete+insert pattern - each worker handles its chunk atomically.

  Not sure if this fits with what you're working on, but figured I'd share. Happy to adjust or scrap it if it doesn't fit the direction.

### DIFF
--- a/apps/epg/tasks.py
+++ b/apps/epg/tasks.py
@@ -1601,16 +1601,12 @@ def track_parallel_epg_progress(self, source_id, group_result_id, total_channels
             return
 
         last_progress = 0
-        stall_count = 0
-        max_stalls = 30  # Exit after 60 seconds of no progress (30 * 2s)
 
         while not group_result.ready():
             try:
                 completed = group_result.completed_count()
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Shall I close this after our conversation?

stlalpha:
Aye capn

PR #746, opened 2025-12-12
feat: Add VLC stream profile with codec detection and streamlink parsing

sethwv:
This pull request adds support for using VLC (`cvlc`) as a stream profile alongside existing streaming backends. It introduces new logic for parsing VLC and Streamlink output to detect codecs and resolutions, updates the database fixtures and migrations to include a VLC stream profile, and ensures VLC is installed and configured for headless operation in the Docker environment.

**VLC/Streamlink integration and parsing:**

* Added logic to `ChannelService.parse_and_store_stream_info` to handle `vlc` and `streamlink` stream types, enabling detection of video/audio codecs from VLC TS demux output and extraction of resolution/quality from Streamlink output.
* Updated `StreamManager._log_stderr_content` to invoke the new parsing logic for VLC and Streamlink based on process output.
* Clarified docstring in `parse_and_store_stream_info` to reflect support for FFmpeg/VLC/Streamlink.

**Database and fixtures:**

* Added a migration (`0019_add_vlc_stream_profile.py`) to create a locked, read-only VLC stream profile in the database, using the default user agent.
* Updated `core/fixtures/initial_data.json` and `fixtures.json` to include the VLC stream profile for initial setup and tests. [[1]](diffhunk://#diff-86b8c92b0ead2540e50663cd2f126679f8cfb7932bb8bb1d812db0dfb418308bR43-R53) [[2]](diffhunk://#diff-323f1fce237026deac65a7f2f0422d2882b9c1a29878cc0f9527bbe693ef4e6cR56-R65)

**Docker environment:**

* Installed VLC and required plugins in the Docker base image and configured environment variables for headless operation. [[1]](diffhunk://#diff-37c17d7a8c67a652b9fa00cebe2ed520f5c7b6741e119b2a7a1c397f2406acddR19) [[2]](diffhunk://#diff-37c17d7a8c67a652b9fa00cebe2ed520f5c7b6741e119b2a7a1c397f2406acddR29-R33)

PR #757, opened 2025-12-16
Add VLC, Refactor Output Log Parsing

sethwv:
This pull request supersedes #746, instead of trying to do git magic I've just cherry picked my changes on to a fresh pull of dev.

---

This pull request adds support for using VLC (`cvlc`) as a stream profile alongside existing streaming backends. It introduces new logic for parsing VLC and Streamlink output to detect codecs and resolutions, updates the database fixtures and migrations to include a VLC stream profile, and ensures VLC is installed and configured for headless operation in the Docker environment.

**VLC/Streamlink integration and parsing:**

* Added logic to `ChannelService.parse_and_store_stream_info` to handle `vlc` and `streamlink` stream types, enabling detection of video/audio codecs from VLC TS demux output and extraction of resolution/quality from Streamlink output.
* Updated `StreamManager._log_stderr_content` to invoke the new parsing logic for VLC and Streamlink based on process output.
* Clarified docstring in `parse_and_store_stream_info` to reflect support for FFmpeg/VLC/Streamlink.

**Database and fixtures:**

* Added a migration (`0019_add_vlc_stream_profile.py`) to create a locked, read-only VLC stream profile in the database, using the default user agent.
* Updated `core/fixtures/initial_data.json` and `fixtures.json` to include the VLC stream profile for initial setup and tests. [[1]](diffhunk://#diff-86b8c92b0ead2540e50663cd2f126679f8cfb7932bb8bb1d812db0dfb418308bR43-R53) [[2]](diffhunk://#diff-323f1fce237026deac65a7f2f0422d2882b9c1a29878cc0f9527bbe693ef4e6cR56-R65)

**Docker environment:**

* Installed VLC and required plugins in the Docker base image and configured environment variables for headless operation. [[1]](diffhunk://#diff-37c17d7a8c67a652b9fa00cebe2ed520f5c7b6741e119b2a7a1c397f2406acddR19) [[2]](diffhunk://#diff-37c17d7a8c67a652b9fa00cebe2ed520f5c7b6741e119b2a7a1c397f2406acddR29-R33)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Advanced filtering for Channels table: Filter menu now allows toggling disabled channels visibility (when a profile is selected) and filtering to show only empty channels without streams (Closes #182)
 - Network Access warning modal now displays the client's IP address for better transparency when network restrictions are being enforced - Thanks [@damien-alt-sudo](https://github.com/damien-alt-sudo) (Closes #778)
+- VLC streaming support - Thanks [@sethwv](https://github.com/sethwv)
+  - Added `cvlc` as an alternative streaming backend alongside FFmpeg and Streamlink
+  - Log parser refactoring: Introduced `LogParserFactory` and stream-specific parsers (`FFmpegLogParser`, `VLCLogParser`, `StreamlinkLogParser`) to enable codec and resolution detection from multiple streaming tools
+  - VLC log parsing for stream information: Detects video/audio codecs from TS demux output, supports both stream-copy and transcode modes with resolution/FPS extraction from transcode output
+  - Locked, read-only VLC stream profile configured for headless operation with intelligent audio/video codec detection
+  - VLC and required plugins installed in Docker environment with headless configuration
 
 ### Changed
 
 - Fixed event viewer arrow direction (previously inverted) — UI behavior corrected. Thanks [@drnikcuk](https://github.com/drnikcuk) (Closes #772)
+- Stream log parsing refactored to use factory pattern: Simplified `ChannelService.parse_and_store_stream_info()` to route parsing through specialized log parsers instead of inline program-specific logic (~150 lines of code removed)
+- Stream profile names in fixtures updated to use proper capitalization (ffmpeg → FFmpeg, streamlink → Streamlink)
 
 ### Fixed
 


sethwv:
@SergeantPanda speaking of the log parse for transcode/commands - I've sort of implemented parsing for both streamlink and VLC at the same time here. The problem with VLC is that unless you're transcoding and have a high verbosity flag on the output, there is no video information that comes through that we can parse.

I think a lot of that detailed information is going to be very specific to ffmpeg - another option depending on the goal of the stats that I initially played around with was creating a custom wrapper for the binary (vlc + streamlink) that runs an ffmpeg probe on the video being output.

SergeantPanda (commented):
I think we should refactor this if we are going to implement it (really should be refactored anyway to be honest).

I'd like to see the channel service for parsing the logs be separated out for ffmpeg, streamlink and VLC probably. 

Hopefully I'm making sense!

sethwv:
I've spent the better part of 3 days with this, and with a little (or a lot) of help from claude turning my pseudocode-grade python into actual python.

> Top-Left: Streamlink
> Top-Right: FFMpeg
> Bottom-Left: VLC Without Video Transcode
> Bottom-Right: VLC With Video Transcode
> image

1. Created log_parsers.py
    - BaseLogParser - Abstract base class defining the parser interface
    - FFmpegLogParser - Handles FFmpeg log parsing (input format, video/audio streams with full metadata)
    - VLCLogParser - Handles VLC log parsing with two modes:
      - Stream-copy mode: Codec detection from TS demux lines
      - Transcode mode: Full metadata (codec, resolution, FPS, audio channels/sample rate) from decoder and transcode output
    - StreamlinkLogParser - Handles Streamlink quality/resolution parsing
    - LogParserFactory - Factory class to route to the appropriate parser based on stream_type
2. Refactored channel_service.py
    - Simplified parse_and_store_stream_info() to use factory pattern
    - Single line routing: `LogParserFactory.parse(stream_type, stream_info_line)`
    - Removed ~150 lines of program-specific parsing logic
3. Enhanced stream_manager.py
    - Detects VLC TS demux lines for codec extraction
    - Detects VLC decoder output for audio metadata (channels, sample rate)
    - Detects VLC transcode output for video metadata (resolution, FPS)
    - Routes to vlc_video or vlc_audio stream types appropriately
4. VLC metadata capabilities:
    - Stream-copy mode: video codec, audio codec, channels, sample rate
    - Transcode mode: all above + resolution, FPS (at cost of CPU)
---
These are the limitations for VLC's Stats:
- VLC Can only output information it is affecting, for example video if it is transcoding. Otherwise the information will not be present.
    - Audio Information can be shown with the baseline command included as a default. This command parameter profile fetches the stream, remuxes to TS and then writes to stdout.
      `-vv -I dummy --no-video-title-show --http-user-agent {userAgent} {streamUrl} --sout #standard{access=file,mux=ts,dst=-}`
    - Video Information comes through when you add transcoding. This command parameter is the same as above, except it decodes both audio/video and re-encodes them to H.264/MPEG.
      `-vv -I dummy --no-video-title-show --http-user-agent {userAgent} {streamUrl} --sout #transcode{vcodec=h264,acodec=mpga}:standard{access=file,mux=ts,dst=-}`

SergeantPanda (commented):
I pushed a small refactor to it to first parse if we konw what command it is instead of trying to determine with each line. Should be a bit more efficient while still giving the same functionality.

sethwv:
Thanks for the additions, I'm not seeing any issues after getting it spun up and running my tests again.

PR #761, opened 2025-12-17
Enhancement/component cleanup

nick4810:
Started updating the components within `pages/` for readability and SoC.

Summary of changes:
- Moved business logic into corresponding `utils/` files
- Introduced `ErrorBoundary` component with generic "Something went wrong" message
- Lazy loading components that aren't needed for initial page load with "Loading..." fallback
- Removed unused `Dashboard` and `Home` pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,12 +17,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - VLC log parsing for stream information: Detects video/audio codecs from TS demux output, supports both stream-copy and transcode modes with resolution/FPS extraction from transcode output
   - Locked, read-only VLC stream profile configured for headless operation with intelligent audio/video codec detection
   - VLC and required plugins installed in Docker environment with headless configuration
+- ErrorBoundary component for handling frontend errors gracefully with generic error message
 
 ### Changed
 
 - Fixed event viewer arrow direction (previously inverted) — UI behavior corrected. - Thanks [@drnikcuk](https://github.com/drnikcuk) (Closes #772)
 - Stream log parsing refactored to use factory pattern: Simplified `ChannelService.parse_and_store_stream_info()` to route parsing through specialized log parsers instead of inline program-specific logic (~150 lines of code removed)
 - Stream profile names in fixtures updated to use proper capitalization (ffmpeg → FFmpeg, streamlink → Streamlink)
+- Frontend component refactoring for improved code organization and maintainability - Thanks [@nick4810](https://github.com/nick4810)
+  - Extracted large nested components into separate files (RecordingCard, RecordingDetailsModal, RecurringRuleModal, RecordingSynopsis)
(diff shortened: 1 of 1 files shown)


SergeantPanda (commented):
I'm not necessarily opposed to a lot of this cleanup/refactoring as I know a lot of it needs to be done. 
That being said, When first logging in I am only getting a screen that says "Something went wrong" and I need to navigate back and forth to the channels page to see channels.

Mind taking a look at that?

nick4810:
Reverted the change on the Channels page to prevent throwing an error on the initial render. Also suggesting to set the User before performing the fetch settings requests.

nick4810:
@SergeantPanda Do you mean display a loading spinner instead of the ChannelsTableOnboarding component while the channels are being fetched?

SergeantPanda:
> @SergeantPanda Do you mean display a loading spinner instead of the ChannelsTableOnboarding component while the channels are being fetched?

Yes exactly. Like the stream table does when it's loading.

nick4810:
Updated ChannelsTable to have similar loading behavior as StreamsTable.

SergeantPanda:
I don't see the message anymore so that is good but it definitely takes longer to load the channels table than it did before (after initial load).

In order for this to be approved we'd want to make sure we aren't slowing anything down.

nick4810:
There shouldn't be any additional processing causing slowness but it might be due to fetchData being called multiple times and showing the loading overlay when it doesn't need to. I reverted the overlay part but left the logic change for displaying the default message.

SergeantPanda:
Ok I can take a look later. Basically I could see slowness when navigating to the channels page even if it had already been loaded once. Prior to the refactor it loads almost instantly but after it takes maybe a half second to 1 second.

nick4810:
> Ok I can take a look later. Basically I could see slowness when navigating to the channels page even if it had already been loaded once. Prior to the refactor it loads almost instantly but after it takes maybe a half second to 1 second.

Ah ok I think I see what you mean. I'll try to adjust the logic a little bit.

nick4810:
Ok @SergeantPanda lmk if that looks better.

SergeantPanda:
That definitely improved the loading speed after initial load. It is VERY close.

The only thing that I'd like to see improved would be the order in which the data loads.

With this refactor, the streams and linked epg don't populate until after all of the logos load. Could we have logos load last?

SergeantPanda:
I see some changes were made, is this ready to be tested?

nick4810:
Yeah I reverted the lazy load on the StreamsTable so it shouldn't be waiting on anything.

SergeantPanda:
Thank you for your work so far on this. I went ahead and made some changes to how the logos work (I've been wanting to do this anyway)
Now, logos should be the last thing to load in which was causing slowness for the streams table loading.

Please review and let me know if you are OK with everything.

nick4810:
👍 Changes look good

SergeantPanda:
Looks good to me too. Thank you!

PR #784, opened 2025-12-26
Fix VOD category filtering for names containing pipe "|" characters

Vitekant:
Use rsplit('|', 1) instead of split('|', 1) to split from the right, preserving any pipe characters in category names (e.g., "PL | BAJKI", "EN | MOVIES"). This ensures the category_type is correctly extracted as the last segment while keeping the full category name intact.

Fixes MovieFilter, SeriesFilter, and UnifiedContentViewSet category parsing.


SergeantPanda (approved):
This looks like a good change. Thank you!

PR #787, opened 2025-12-28
Enhancement/component cleanup

nick4810:
- Guide.jsx
- Plugins.jsx

Summary of changes:
- Added Plugins store to handle fetching plugins and loading state


SergeantPanda:
There are some conflicts now with the other changes from your other PR it looks like. Mind fixing that?

nick4810:
> Playing around with this and I think a lot of these are good changes. For the TV guide, I've been playing with that and I notice when navigating to the guide, it shows "no program data" while it's being fetched, could we have it show that it's loading instead to avoid any confusion for the user?

@SergeantPanda Lmk if the loading logic looks correct.

SergeantPanda (approved):
Yes, this looks good. Thank you!

nick4810:
Yep looks good

SergeantPanda:
Thank you for your work on this!

PR #789, opened 2025-12-30
Signaling ready when no StreamTable rendered

nick4810:
With the logo signaling changes introduced in https://github.com/Dispatcharr/Dispatcharr/commit/f5c6d2b576f0b6c050b704ef63cc38bbac406215 Standard Users don't see logos on the Channels page because StreamsTable doesn't get rendered.

This change signals ready for the StreamsTable within the conditional render block.


SergeantPanda:
Good catch!

PR #794, opened 2025-12-31
Fix root-owned __pycache__ by running Django commands as non-root user

sethwv:
This pull request focuses on improving the Docker setup and entrypoint scripts for the application, mainly to address file permission issues and ensure proper environment variable handling. The changes enhance how static files are managed and how Django commands are executed during container startup.

**Docker and Entrypoint Improvements:**

* Changed the execution of Django management commands (`migrate` and `collectstatic`) to run as the non-root user specified by `$POSTGRES_USER`, preventing permission issues with generated files.
* Added `/app/static` to the list of application directories created during initialization, ensuring the directory exists before static files are collected.

**Environment Variable Handling:**

* Included `PYTHONDONTWRITEBYTECODE` in the list of variables exported to the environment profile, which helps prevent Python from writing `.pyc` files and keeps container filesystems cleaner.

**Dockerfile Adjustments:**

* Removed the `collectstatic` step from the Docker build process in `Dockerfile`, delegating this responsibility entirely to the entrypoint script at container startup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - M3U and EPG manager page no longer crashes when a playlist references a deleted channel group (Fixes screen blank on navigation)
 - Stream validation now returns original URL instead of redirected URL to prevent issues with temporary redirect URLs that expire before clients can connect
 - XtreamCodes EPG limit parameter now properly converted to integer to prevent type errors when accessing EPG listings (Fixes #781)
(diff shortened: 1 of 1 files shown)


sethwv:
> Ok, so I think I'd rather keep the PYC's generating personally but the rest are OK changes.

We can definitely remove `PYTHONDONTWRITEBYTECODE` from the global variables array. I just want to mention that in this PR, it’s not actually being set. The addition just allows it to be passed down to downstream users instead of ending at the first level of `su -`.

SergeantPanda (approved):
Ok you're right this all looks like it will be good changes. Thank you!

PR #795, opened 2025-12-31
Enhancement/component cleanup logos settings

nick4810:

- Logos.jsx
- Settings.jsx

Summary of changes:

- Each panel in Settings is using its own form state with the parent component handling active=true/false


SergeantPanda (commented):
This looks like a good refactor of the settings. Thank you!

PR #811, opened 2026-01-05
Extracted component and util logic

nick4810:

- Stats.jsx
- VODs.jsx

Summary of changes:


SergeantPanda:
Looks good. I've been wanting to separate out the VOD stuff for a while.. Thanks for doing that!

Issue #813, opened 2026-01-05
[Bug]: bulk_create() prohibited due to unsaved related object during episode refresh


kaputti1o1:
### Describe the bug

This Bug might be similar to the issue #770 but it is not solved with the bug fix in version 0.16.

When refreshing the VODs the series are not created. Only the last batch chunk is processed correctly.

In the logs it looks like this (in Debug mode no more info is available beside the series names):

`2026-01-04 08:57:23,141 INFO apps.vod.tasks Refreshing series for account Domain1

2026-01-04 08:57:23,143 INFO apps.vod.tasks Fetching all series from provider...

2026-01-04 08:57:23,857 INFO apps.vod.tasks Processing series chunk 1/4 (1000 series)

2026-01-04 08:57:23,857 INFO apps.vod.tasks Processing series batch of 1000 series for account Domain1

2026-01-04 08:57:23,933 INFO apps.vod.tasks Executing batch operations: 977 series to create, 0 to update

2026-01-04 08:57:24,660 ERROR apps.vod.tasks Series batch processing failed: bulk_create() prohibited to prevent data loss due to unsaved related object 'series'.

2026-01-04 08:57:24,663 INFO apps.vod.tasks Processing series chunk 2/4 (1000 series)

2026-01-04 08:57:24,663 INFO apps.vod.tasks Processing series batch of 1000 series for account Domain1

2026-01-04 08:57:24,729 INFO apps.vod.tasks Executing batch operations: 960 series to create, 1 to update

2026-01-04 08:57:25,400 ERROR apps.vod.tasks Series batch processing failed: bulk_create() prohibited to prevent data loss due to unsaved related object 'series'.

2026-01-04 08:57:25,403 INFO apps.vod.tasks Processing series chunk 3/4 (1000 series)

2026-01-04 08:57:25,403 INFO apps.vod.tasks Processing series batch of 1000 series for account Domain1

2026-01-04 08:57:25,473 INFO apps.vod.tasks Executing batch operations: 995 series to create, 1 to update

2026-01-04 08:57:26,178 ERROR apps.vod.tasks Series batch processing failed: bulk_create() prohibited to prevent data loss due to unsaved related object 'series'.

2026-01-04 08:57:26,181 INFO apps.vod.tasks Processing series chunk 4/4 (50 series)

2026-01-04 08:57:26,181 INFO apps.vod.tasks Processing series batch of 50 series for account Domain1

[...]

PR #814, opened 2026-01-05
Fix bulk_create and bulk_update errors for VOD content refresh

patchy8736:
This commit addresses the issues described in GitHub issue #813 where bulk_create and bulk_update operations were failing with errors like:

- "bulk_update() prohibited to prevent data loss due to unsaved related object 'movie'"
- "bulk_create() prohibited to prevent data loss due to unsaved related object 'series'"

The changes include:

1. Fixed movie batch processing (lines 652-678):
   - Properly handle relations_to_update to reference saved movies
   - Filter out relations with unsaved movies (no PK) before bulk operations
   - Add warning logs when relations are skipped due to unsaved movies

2. Fixed series batch processing (lines 1007-1023):
   - Properly handle relations_to_update to reference saved series
   - Filter out relations with unsaved series (no PK) before bulk operations
   - Add warning logs when relations are skipped due to unsaved series

3. Root cause resolution:
   - The errors occurred because bulk_create doesn't immediately set primary keys on created objects in all cases
   - Relations were being created with references to objects that didn't have PKs yet
   - When bulk_update was called on relations, Django detected that the related objects (movies/series) didn't have primary keys

4. Solution implementation:
   - All relations now reference objects that have been successfully saved to the database
   - Relations with unsaved objects are filtered out before bulk operations
   - Proper warnings are logged when relations are skipped
   - The transaction ensures atomicity of the operations

These changes ensure data integrity while maintaining performance during VOD content refresh operations.

### DIFF
--- a/apps/vod/tasks.py
+++ b/apps/vod/tasks.py
@@ -654,12 +654,26 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
                 if id(relation.movie) in created_movies:
                     relation.movie = created_movies[id(relation.movie)]
 
-            # Handle relations
-            if relations_to_create:
-                M3UMovieRelation.objects.bulk_create(relations_to_create, ignore_conflicts=True)
+            # Also update relations_to_update to reference saved movies
+            for relation in relations_to_update:
+                if id(relation.movie) in created_movies:
+                    relation.movie = created_movies[id(relation.movie)]
 
-            if relations_to_update:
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you for this PR. I have been wanting to revisit this code a little bit more for a while to see if I can get some more performance out of it. I decided to go a different route in attempting to fix the bug.

PR #823, opened 2026-01-07
Bug Fix: Add socket-timeout to uWSGI production config to prevent VOD stream timeouts/streams vanishing from stats page

patchy8736:
---
  Summary

  - Added socket-timeout = 600 to docker/uwsgi.ini to match the existing http-timeout = 600 value
  - Fixes issue where VOD streams disappear from the stats page even though the stream is still playing

  Problem

  The production uWSGI configuration was missing the socket-timeout directive, causing it to default to 4 seconds. When clients (e.g., VLC) buffer VOD streams and temporarily stop reading from the HTTP socket, uWSGI's write operations timeout after 4 seconds, triggering premature stream cleanup.

  Root Cause

  - Default socket-timeout in uWSGI is 4 seconds
  - Client buffering (e.g., VLC filling its buffer) is normal behavior, not a disconnect
  - uWSGI times out on write operations → OSError: write error → GeneratorExit → cleanup triggered
  - This causes VOD streams to vanish from the stats page even though the stream continues playing normally

  Fix

  Add socket-timeout = 600 to production config, matching http-timeout = 600. This gives uWSGI sufficient time to wait for clients to resume reading from buffered sockets.

  Impact

  Prevents:
  - uwsgi_response_write_body_do() TIMEOUT !!! errors in logs
  - GeneratorExit exceptions and premature stream cleanup
  - VOD streams vanishing from the stats page even though the stream continues playing

  Notes

  - Debug config already had socket-timeout = 3600 (which is why the issue wasn't observed in debug mode)
  - This fix aligns production behavior with debug config while maintaining the production-appropriate 10-minute timeout duration
  - Most users use uwsgi.ini (when DISPATCHARR_ENV != "dev" and DISPATCHARR_DEBUG != "true")

  ---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Fixed VOD streams disappearing from stats page during playback by adding `socket-timeout = 600` to production uWSGI config. The missing directive caused uWSGI to use its default 4-second timeout, triggering premature cleanup when clients buffered content. Now matches the existing `http-timeout = 600` value and prevents timeout errors during normal client buffering - Thanks [@patchy8736](https://github.com/patchy8736)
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Looks good. Thank you!

PR #838, opened 2026-01-10
fix(ts-proxy): eliminate TOCTOU race condition and connection leaks

patchy8736:
This PR fixes two critical issues in the TS proxy connection management that caused false positive "at capacity" errors under high concurrency and connection counter leaks during cleanup.

### 1. TOCTOU Race Condition (Time-Of-Check to Time-Of-Use) The original code checked profile capacity and then incremented the counter in two separate Redis operations. Under high concurrency, multiple requests could pass the capacity check simultaneously, leading to:
- False positive "All active M3U profiles have reached maximum connection limits"
- Actual connections exceeding profile limits

### 2. Connection Counter Leaks
The release_stream() method had no fallback mechanism when Redis keys were missing. If cleanup happened in an unexpected order, the profile connection counter would never be decremented, causing permanent "at capacity" errors.

### Part 1: Atomic Slot Reservation
- Added _check_and_reserve_profile_slot() using a Lua script
- Check and increment happen in a single atomic Redis operation
- Eliminates the TOCTOU race condition entirely

### Part 2: Defensive Release Stream
- Channel.release_stream() now has metadata hash fallbacks
- Returns bool for success/failure tracking
- Detailed logging for debugging
- All 7 call sites now log failures for monitoring

### Part 3: Consistency & Logging Improvements
- Fixed Stream.release_stream() to return bool (was returning None)
- Standardized all logs to use channel UUID instead of database ID
- Downgraded expected cleanup messages from WARNING to INFO (prevents log spam during normal disconnect flow)

## Changes
- apps/channels/models.py: Atomic Lua reservation, defensive release, bool returns
- apps/proxy/ts_proxy/server.py: Return value checking
- apps/proxy/ts_proxy/services/channel_service.py: Return value checking
- apps/proxy/ts_proxy/stream_generator.py: Return value checking
- apps/proxy/ts_proxy/views.py: Return value checking

## Testing
- Tested under high concurrency - no more false positive errors
- Verified connection counters decrement correctly in all scenarios
- Logs are cleaner and easier to correlate across components

### DIFF
--- a/apps/channels/models.py
+++ b/apps/channels/models.py
@@ -16,6 +16,10 @@ logger = logging.getLogger(__name__)
 # If you have an M3UAccount model in apps.m3u, you can still import it:
 from apps.m3u.models import M3UAccount
 
+# Additional imports for connection leak fix
+from apps.proxy.ts_proxy.redis_keys import RedisKeys
+from apps.proxy.ts_proxy.constants import ChannelMetadataField
+
 
 # Add fallback functions if Redis isn't available
 def get_total_viewers(channel_id):
@@ -123,13 +127,13 @@ class Stream(models.Model):
     stream_stats = models.JSONField(
         null=True,
         blank=True,
-        help_text="JSON object containing stream statistics like video codec, resolution, etc."
+        help_text="JSON object containing stream statistics like video codec, resolution, etc.",
     )
     stream_stats_updated_at = models.DateTimeField(
         null=True,
         blank=True,
         help_text="When stream statistics were last updated",
-        db_index=True
+        db_index=True,
     )
 
     class Meta:
@@ -154,7 +158,13 @@ class Stream(models.Model):
         if use_stream_id:
             effective_url = stream_id
 
-        stream_parts = {"name": name, "url": effective_url, "tvg_id": tvg_id, "m3u_id": m3u_id, "group": group}
+        stream_parts = {
+            "name": name,
+            "url": effective_url,
+            "tvg_id": tvg_id,
+            "m3u_id": m3u_id,
+            "group": group,
+        }
 
         hash_parts = {key: stream_parts[key] for key in keys if key in stream_parts}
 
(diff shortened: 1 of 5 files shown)


SergeantPanda:
Please rebase this to our dev branch.

patchy8736:
Done :)

SergeantPanda:
Thank you. I am going to be very cautious with this PR. I generally don't like PR's that touch the proxy since it is such a crucial part of Dispatcharr and easy to implement bugs with.

I do agree with some of this though so I'll play with it.

patchy8736:
That's totally fine. :) Thanks for looking at it!

CodeBormen:
@patchy8736 Good work, I have been working on a complimentary branch that definitely has some overlap with this one (#949). I have been working on fixes for connection leaks but also have gone back to address issues where there are double decrements, negative counters, counter drift, stream exception, and scheduling cleanup. I don't see a clean way to merge the branches together without some overlap, do you mind if I work your changes into the branch that I am working on so that it contains the changes of both?

patchy8736:
@CodeBormen Of course, go for it!
That's why we're all here - to share code :)
I've been using Dispatcharr for quite a while now with this fix and it's been working really good. I'm excited to see what you've come up with.

CodeBormen:
Awesome, I will take a look tonight at merging these changes together after some testing. I did want to ask if there was a particular reason you used a Lua script to correct the TOCTOU issue rather than using native Redis tooling? Just to be clear this is not a criticism, I am just wondering if there is a significant advantage to your approach? I have not worked with Lua a whole lot, so always down to learn something new!

CodeBormen:
Hey @patchy8736, thanks for allowing me to pull in your changes. I hope you don't mind, but I did end up translating your Lua script into native Redis commands for the TOCTOU race conditions. I wanted to make sure that it stayed consistent with the project, and more frankly wanted to make sure that I understood the change.

Otherwise I did incorporate all of the other changes in that you had if they weren't already.

If you have the chance give it a look and let me know what you think! I did test the changes as comprehensively as I could, but of course no better evidence than actual end testing.

patchy8736:
Hey @CodeBormen, thanks for the question! I went with Lua scripting for a few reasons:

  1. True atomicity - The Lua script handles both the check and increment in a single Redis operation. With the INCR-first approach, the counter technically goes over the limit before being rolled back. In most cases this doesn't matter, but it means monitoring could momentarily show 6/5 connections, and other requests might incorrectly see "at capacity" during that rollback window.
  2. No rollback risk - If anything crashes between the INCR and DECR in the native approach, that counter slot is permanently leaked. Lua runs atomically so it's all-or-nothing.
  3. Performance at capacity - When profiles are maxed out, Lua is a single call that returns failure immediately vs two calls (INCR + rollback DECR).
  4. It's what Redis docs recommend for conditional operations like this.

That said, I totally get wanting to keep it consistent with the rest of the project. Your translation achieves the same result and is more readable for anyone not familiar with Lua.
Both approaches fix the TOCTOU issue properly - Lua's just the more robust pattern for this specific use case.

Appreciate you taking the time to work through the change and incorporate everything else from the PR!

SergeantPanda:
Thank you two for working this out. It sounds like it will be incorporated in #949 so I will close this one.

PR #841, opened 2026-01-11
Tests/frontend unit tests

nick4810:
- Components within pages/
- JS files within utils/

Also added workflow to run the tests
- .github/workflows/frontend-tests.yml

There's a couple of failing tests that appear to be valid:
- src/utils/__tests__/networkUtils.test.js
  - 'should not match invalid IPv4 CIDR notation' 
    - expect(networkUtils.IPV4_CIDR_REGEX.test('192.168.1.0/33')).toBe(false);
  - 'should match valid IPv6 CIDR notation'
    - expect(networkUtils.IPV6_CIDR_REGEX.test('2001:0db8:85a3:0000:0000:8a2e:0370:7334/64')).toBe(true);
- src/utils/__tests__/dateTimeUtils.test.js
  - 'should return null for null value'
    - TypeError: default.invalid is not a function


SergeantPanda:
Wow, you've been busy! Thank you for this.

You're correct on the network utils, originally we did not support ipv6.

PR #865, opened 2026-01-17
fix: stabilize build-dev paths and push behavior

JeffreyBytes:
### Description:
This updates `docker/build-dev.sh` so it works from any cwd by anchoring paths to the script and repo root, builds with explicit --platform args when provided, and fixes push behavior and messaging. Also normalizes formatting and quoting to avoid word-splitting issues.

### Changes:

  - Resolve Dockerfile and build context relative to the script location.
  - Convert -a to --platform args safely.
  - Fix push logic to honor -p and correct push hints.

### DIFF
--- a/docker/build-dev.sh
+++ b/docker/build-dev.sh
@@ -1,65 +1,69 @@
-#!/bin/bash 
+#!/bin/bash
 set -e
 
 # Default values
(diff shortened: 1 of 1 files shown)


SergeantPanda (approved):
Thank you for this. I'm not sure how many use this script but it doesn't hurt to improve it regardless.

PR #867, opened 2026-01-18
Fix streams requery to respect table filters

JeffreyBytes:
### Bug Description:
Removing streams with “Only Unassociated” on left deleted rows visible, pointing to a stale query cache.

### Changes: 
  - Fix stale Streams table results by requerying with the table’s latest query params.
  - Store the last streams table params in the streams table store and reuse them in API.requeryStreams().
  - Ensures deletes and websocket refreshes reflect active filters like “Only Unassociated.”

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -715,6 +715,7 @@ export default class API {
   static async queryStreamsTable(params) {
     try {
       API.lastStreamQueryParams = params;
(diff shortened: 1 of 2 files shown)


SergeantPanda:
So, I am unable to duplicate the behavior you're seeing but I might be misunderstanding and testing incorrectly.

That being said, I don't see anything wrong with this and it looks good to me.

Thank you!

PR #868, opened 2026-01-18
Move caching to client to remove burden on dispatch server 

DawtCom:
This should improve performance as it caches long lived images / logos on the client side.  Reducing round trips to the host.


DawtCom:
This also respects the proxied source image cache control headers.

DawtCom:
Based on a few tests, this could reduce a few seconds for logo loads where people have large number of channels based on the waterfall in debugging tools.

SergeantPanda (approved):
This looks really good. I've been wanting to get logos loading faster in dev and this did the trick!

PR #879, opened 2026-01-21
Feature: Support for external PostgreSQL and Redis services

s0len:
## Summary
Add ability to use external PostgreSQL and Redis instead of the internal services bundled in the container. This enables proper Kubernetes deployments and other scenarios where you want to use managed database/cache services.

## New Environment Variables
| Variable | Default | Description |
|----------|---------|-------------|
| `EXTERNAL_POSTGRES` | auto-detect | Set to `true` to skip internal PostgreSQL startup |
| `EXTERNAL_REDIS` | auto-detect | Set to `true` to skip internal Redis startup |
| `REDIS_PORT` | `6379` | Redis port (was missing from exports) |

## Auto-detection Behavior
- If `POSTGRES_HOST` is not `localhost` or `127.0.0.1`, automatically treats PostgreSQL as external
- If `REDIS_HOST` is not `localhost` or `127.0.0.1`, automatically treats Redis as external
- Explicit `EXTERNAL_POSTGRES`/`EXTERNAL_REDIS` environment variables override auto-detection

## Changes
- Skip internal PostgreSQL initialization (`02-postgres.sh`) and startup when using external
- Wait for external PostgreSQL to be ready before proceeding with migrations
- Remove `redis-server` daemon from uwsgi config when using external Redis
- Add service configuration logging at startup showing internal/external status
- Export `REDIS_PORT` variable for consistency

### Kubernetes with external services
```yaml
env:
  - name: POSTGRES_HOST
    value: "postgres-service"
  - name: REDIS_HOST  
    value: "redis-service"
  # Auto-detects external services based on host != localhost
```

### Explicit override (use external even with localhost via port-forward)
```yaml
env:
  - name: POSTGRES_HOST
    value: "localhost"
  - name: EXTERNAL_POSTGRES
    value: "true"
```

## Test plan
- [ ] Run container with default settings (internal PostgreSQL + Redis)
- [ ] Run container with `POSTGRES_HOST=external-db` (should skip internal PostgreSQL)
- [ ] Run container with `REDIS_HOST=external-redis` (should skip internal Redis)
- [ ] Run container with explicit `EXTERNAL_POSTGRES=true` override
- [ ] Verify migrations run successfully against external database
- [ ] Verify Celery connects to external Redis

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #881, opened 2026-01-21
feat: Add DISPATCHARR_ENV=modular support for multi-container deployment

CodeBormen:
- docker/entrypoint.sh: Conditional PostgreSQL init/startup for modular mode
- docker/entrypoint.celery.sh: Dedicated entrypoint for celery container - prevents race condition in modular deployment
- docker/uwsgi.modular.ini: New uWSGI config without Redis/Celery daemons
- docker/Dockerfile: Add line ending fixes and chmod for entrypoints - adds cross-platform support for image builds
- docker/docker-compose.yml: Configure modular mode with "DISPATCHARR_ENV=modular" - uses new celery entrypoint and adds dependency for web service to prevent race condition; also moves volume mount from "../:/app" to "./data:/data" for the celery service to create consistency with AIO, source is copied in at the new celery entrypoint

Tested AIO, modular, and dev deployments all with success

### DIFF
--- a/docker/init/02-postgres.sh
+++ b/docker/init/02-postgres.sh
@@ -192,7 +192,7 @@ ensure_utf8_encoding() {
 }
 
 check_external_postgres_version() {
-    # Only check for modular deployments using external databases
+    # Only check for modular deployments
     if [[ "$DISPATCHARR_ENV" != "modular" ]]; then
         return 0
     fi
@@ -200,7 +200,7 @@ check_external_postgres_version() {
     echo "🔍 Checking external PostgreSQL version compatibility..."
 
     # Get minimum required version from base image (set in entrypoint.sh)
-    # PG_VERSION is the version installed in DispatcharrBase
+    # PG_VERSION is from DispatcharrBase
     MIN_REQUIRED_VERSION=$PG_VERSION
 
     # Query external PostgreSQL version
@@ -223,7 +223,7 @@ check_external_postgres_version() {
         echo "  Required Version:  PostgreSQL $MIN_REQUIRED_VERSION or higher"
         echo ""
         echo "  Your external PostgreSQL database is too old for Dispatcharr."
-        echo "  Please upgrade to PostgreSQL $MIN_REQUIRED_VERSION or later."
+        echo "  Please upgrade to PostgreSQL $MIN_REQUIRED_VERSION or higher."
         echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
         echo ""
         return 1


SergeantPanda:
This is pretty close to what I have been wanting to do. One thing that is actually pretty important is the database running in UTF8 encoding though or the user will eventually get some errors.

Mind making that change?

timothystewart6:
Really looking forward to this.  I would like to break up my monolith deployment since I already run redis and postgres individually.

CodeBormen:
@SergeantPanda Yes, I'll add UTF8 in and validate through testing. Will take this PR to draft and mark for review after integration/testing!

SergeantPanda:
Thank you!

CodeBormen:
Alright, I'm back with my changes.

---------------- UTF8 Encoding Check ----------------

The UTF8 encoding check was really only designed to work on the AIO version, it would have failed on modular deployments. 02-postgres.sh which contains the UTF8 function could not be safely run on unmanaged postgres instances. 

The entrypoint.sh script was modified to allow the 02-postgres.sh to run on all types of deployment (to make the function available to modular deployments).

The 02-postgres.sh script was modified to include if/else logic for external databases that should not have setup run on them. The modification allows the encoding check to be run safely on external databases, and use input variables to run the same set of database repairs (dump/drop/recreate/restore) if there is an encoding issue.

Validation logic was also included as part of the encoding check so that if there is a failure at any step of the repair phase, it will not continue to corrupt the db.

---------------- Environment Variables ----------------

With the above change making external db instances viable (whether same- or external-stack), there was no way to include a custom redis host/port without changing the hardcoded string in the compose file. The entrypoint was modified so that Django will dynamically build it at runtime with default or user-defined variables. Postgres already included native support for a custom port, but it was not included in the compose.

With the change, the environment variables POSTGRES_HOST, POSTGRES_PORT, REDIS_HOST, and REDIS_PORT are all now user-accessible in the compose file, which should bring proper support for both the modular deployment type as well as completely external db services.

@timothystewart6 - the only caveat is that with external dbs the end user is responsible for creating the database in their application and making sure the db user has the correct permissions, since it's not being done at initialization. I think this is fairly common for most applications that support external dbs, and shouldn't be a surprise. If it ends up being a significant issue, perhaps in the future support could be added to do a readiness check on the external db.

---------------- Other ----------------

Since this change brings support for external db instances, logic and verbose logging were added to the entrypoint to ensure that dispatcharr can connect to the specified service, as well as output to notify the user of successful connections for them on $HOST:$PORT

timothystewart6:
> @timothystewart6 - the only caveat is that with external dbs the end user is responsible for creating the database in their application and making sure the db user has the correct permissions, since it's not being done at initialization. I think this is fairly common for most applications that support external dbs, and shouldn't be a surprise. If it ends up being a significant issue, perhaps in the future support could be added to do a readiness check on the external db.
> 
Sounds good to me.  Agreed that user should be responsible for creating their own user/database/permissions.  As long as the init process creates the schema and seeds the database, that's about all we can ask for.  I assume it will considering it would need to do this for the monolith anyway.   Thank you!

docker/docker-compose.yml
```diff
@@ -60,11 +61,12 @@ services:
     environment:
       - DISPATCHARR_ENV=modular
       - POSTGRES_HOST=db
+      - POSTGRES_PORT=5432
       - POSTGRES_DB=dispatcharr
       - POSTGRES_USER=dispatch
       - POSTGRES_PASSWORD=secret
       - REDIS_HOST=redis
-      - CELERY_BROKER_URL=redis://redis:6379/0
+      - REDIS_PORT=6379
```
timothystewart6, line 69: Any reason to not just add `REDIS_PASSWORD` in this PR for those that use a password?

CodeBormen: Hey Tim,

Yeah, I took a look at adding this and the reason was because REDIS_HOST and REDIS_PORT were already supported  through the codebase so I just needed to get the variables in the right place.

Currently, the redis connection is hardcoded across several files and there is no support for auth anywhere, so I'd need to change about ~10 files to add proper redis auth support.

I should be able to add this in the following week or so but will need some time to make changes and have the devs review. I guess the good news is that since redis doesn't need to be persistent you could always run redis in the local stack and then when there is proper auth support, switch to external and it will rebuild its cache.

s0len:
Hey @CodeBormen, great work on the modular deployment support!

I had a similar PR (#879) focused on external service support with some features that might complement this nicely. Would you be open to incorporating these additions?

## 1. Auto-detection of external services

This allows users to skip setting `DISPATCHARR_ENV=modular` when the host is clearly external (useful for Kubernetes where you just set `POSTGRES_HOST=postgres-service`):

```bash
# Add after REDIS_PORT export in entrypoint.sh (around line 51)

# Auto-detect external services based on hostname
# Can be explicitly overridden with EXTERNAL_POSTGRES/EXTERNAL_REDIS env vars
if [ -z "$EXTERNAL_POSTGRES" ]; then
    if [ "$POSTGRES_HOST" != "localhost" ] && [ "$POSTGRES_HOST" != "127.0.0.1" ]; then
        EXTERNAL_POSTGRES="true"
    else
        EXTERNAL_POSTGRES="false"
    fi
fi
export EXTERNAL_POSTGRES

if [ -z "$EXTERNAL_REDIS" ]; then
    if [ "$REDIS_HOST" != "localhost" ] && [ "$REDIS_HOST" != "127.0.0.1" ]; then
        EXTERNAL_REDIS="true"
    else
        EXTERNAL_REDIS="false"
    fi
fi
export EXTERNAL_REDIS
```

Then update the conditionals to use these flags OR modular mode:

```bash
# Change this:
if [[ "$DISPATCHARR_ENV" != "modular" ]]; then

# To this (supports both explicit modular mode AND auto-detected external):
if [[ "$DISPATCHARR_ENV" != "modular" ]] && [[ "$EXTERNAL_POSTGRES" != "true" ]]; then
```

## 2. Service configuration logging at startup

Helps users confirm their setup is correct:

```bash
# Add after the auto-detection block

# Log service configuration
echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
echo "Service Configuration:"
if [ "$EXTERNAL_POSTGRES" = "true" ] || [ "$DISPATCHARR_ENV" = "modular" ]; then
    echo "  🗄️  PostgreSQL: external (${POSTGRES_HOST}:${POSTGRES_PORT})"
else
    echo "  🗄️  PostgreSQL: internal"
fi
if [ "$EXTERNAL_REDIS" = "true" ] || [ "$DISPATCHARR_ENV" = "modular" ]; then
    echo "  📮 Redis: external (${REDIS_HOST}:${REDIS_PORT})"
else
    echo "  📮 Redis: internal"
fi
echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
```

## 3. Export the new variables for profile.d

```bash
# Update the envs_to_set array to include:
EXTERNAL_POSTGRES EXTERNAL_REDIS
```

## Benefits of combining

- **Your PR**: Comprehensive modular mode with Celery entrypoint, uwsgi config, healthchecks
- **My additions**: Auto-detection for Kubernetes/external DB scenarios without requiring explicit `DISPATCHARR_ENV=modular`

Users get the best of both worlds - explicit modular mode for docker-compose, and auto-detection for Kubernetes/external services.

Let me know if you'd like me to submit these as commits to your branch, or if you prefer to incorporate them yourself!

---

*Copy everything above and post as a comment on PR #881*

CodeBormen:
Hey @s0len,

Thanks, spent a good amount of time on it this week, just happy to help bring modular support!

As far as your first suggested change goes, I don't think that adding auto-detection of external services brings enough value in comparison to the risk it introduces. Allowing autodetection by evaluating hostname adds ambiguity and hidden logic to the script, whereas DISPATCHARR_ENV=modular is a very explicit decision from the end user and is not much effort to include. This also works correctly for Kubernetes deployments. Detection of external services by using localhost or 127.0.0.1 is a fragile mechanism as it makes assumptions about loopback addresses and hostname resolution. For example, in an instance where the app refers to the internal postgres instance as 'POSTGRES_HOST=/var/run/postgresql' this is something that would trigger external detection validation. I will admit that scenarios like these are unlikely, but the change is not solving a significant problem.

Regarding your second change around logging verbosity for external services, I have actually already added this as part of this PR. Here is an example of logging output at startup:

🔗 Modular mode: Using external PostgreSQL at db:5432
✅ External PostgreSQL is ready

🔗 Modular mode: Using external Redis at redis:6379
✅ External Redis is ready

s0len:
Thanks for your work and I totally get you. I rest my case :)

SergeantPanda (approved):
Well done. Thank you for all of the time you spent on this. 

I tested to verify it does not break the aio deployments and your code looks good for modular. I trust it's been tested thoroughly at this point so we should be good to merge.

The only thing we *may* want to improve is version checking of postgres version for modular deployments. 

I made two very minor changes to this. 
1) Indented if statement for modular check in postgres setup script
2) Bumped Postgres version from 14 to 17 for modular compose.

Karbust:
Just checking in to see when this will be merged into main. I'm new to the project and want to use the modular approach.

Thank you @CodeBormen for the changes.

SergeantPanda:
> Just checking in to see when this will be merged into main. I'm new to the project and want to use the modular approach.
> 
> Thank you @CodeBormen for the changes.

Most likely in the next release. You can run AIO and switch later. No reason to wait.

SergeantPanda:
@CodeBormen did you want to look at the Postgres version issue or do you think it isn't anything to worry about?

CodeBormen:
@SergeantPanda You're right, I should add a postgres version check in. I don't think it will be a problem in the near future but it will eventually become an issue down the line. I will find a sensible way to add a check into 02-postgres.sh that references the pg version defined in DispatcharrBase. Should have this tested and pushed in a little bit.

CodeBormen:
Here is example output of the new version check:

postgres_match

postgres_newer

postgres_out_of_date

CodeBormen:
@SergeantPanda should be good to go now with a version check

SergeantPanda:
Merged, thank you for the work on this!

docker/docker-compose.yml
timothystewart6: Looks like this was not added.  All good, maybe I will take a swing at it sometime

CodeBormen: Hey Tim,

I wanted to get this one pushed through so default instances will work, but I will come back around and add redis auth support this week!

timothystewart6: Oh cool, thank you!  No worries!

CodeBormen:
@timothystewart6 have added support for Redis auth in https://github.com/Dispatcharr/Dispatcharr/pull/937 - you'll see why I wanted it to be a separate PR

mrdynamo:
Just having a look through the files modified. It doesn't look like this adds TLS auth support for Postgres right? Just password auth only?

CodeBormen:
Correct, both this PR and my PR for Redis add authentication support only, not any transport encryption. Perhaps TLS support is something that could be added in the future but the addition is not trivial. It would fairly doable to add certification validation for an external host, but I expect that many users would have an expectation of Dispatcharr to generate cert files and make them available for export which adds considerable complexity.

PR #882, opened 2026-01-22
Fix "Invalid page" error when filtering empty channels

### LINKED ISSUE #864: [Bug]: "Invalid page" when trying to show empty channels
### Describe the bug

A clear and concise description of what the bug is. What did you expect to happen?

When on page 2+ of the channels list and select Empty Channels, an error pops up that says "Invalid page". Go back to page 1 and the Empty Channels are there

I would expect that when you click that in this condition, it would go back to page 1.

This is a low priority bug, not seriously impactful.

When the following conditions are true, there is an "Invalid page" error that pops up:
1. not on page 1 of channels list
2. select Only Empty Channels
3. Invalid page error pops up

Image

### Dispatcharr Version

v0.17.0

### Docker Version

29.1.3, build f52814d

services:
  dispatcharr:
    container_name: dispatcharr-dev
    image: ghcr.io/dispatcharr/dispatcharr:0.17.0
    network_mode: xxxxx   
    volumes:
      - /file/docker/appdata/dispatcharr_dev:/data
      - /file/docker/appdata/dispatcharr_data/epgs:/data/epgs/
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
    deploy:
      resources:
          reservations:
              devices:
                  - driver: nvidia
                    count: all
                    capabilities: [gpu]

### Client Information

n/a, UI issue.

CodeBormen:
Reset pagination to page 1 when toggling "Only Empty Channels" filter to prevent requesting non-existent pages after filtering reduces results.

Fixes #864

### DIFF
--- a/frontend/src/components/tables/ChannelsTable.jsx
+++ b/frontend/src/components/tables/ChannelsTable.jsx
@@ -1413,6 +1413,8 @@ const ChannelsTable = ({ onReady }) => {
             setShowDisabled={setShowDisabled}
             showOnlyStreamlessChannels={showOnlyStreamlessChannels}
             setShowOnlyStreamlessChannels={setShowOnlyStreamlessChannels}
+            pagination={pagination}
+            setPagination={setPagination}
           />
 
           {/* Table or ghost empty state inside Paper */}
--- a/frontend/src/components/tables/ChannelsTable/ChannelTableHeader.jsx
+++ b/frontend/src/components/tables/ChannelsTable/ChannelTableHeader.jsx
@@ -109,6 +109,8 @@ const ChannelTableHeader = ({
   setShowDisabled,
   showOnlyStreamlessChannels,
   setShowOnlyStreamlessChannels,
+  pagination,
+  setPagination,
 }) => {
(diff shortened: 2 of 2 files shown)


CodeBormen:
FYI it is likely that a similar fix for the same error is going to be needed for the following as well:

- ChannelsTable - showDisabled
- ChannelsTable - Delete
- StreamsTable - Delete
- VODLogosTable - Filters
- VODLogosTable - Delete
- LogosTable - Delete

If you want me to incorporate pagination fixes for the rest of them as well just let me know and I will move this PR to draft, make my changes, and resubmit for review

SergeantPanda:
Thank you for submitting this.
I was playing with it a bit because I know we need to handle the filters/pagination better. I realized we should probably handle this at the api level and wrote some logic for that instead. 

Sorry about that but I just think it will be better long term as more filters are added to just catch it earlier in the code.

CodeBormen:
No worries a fix at the api level would be a better approach than what my fix was, just glad to see it fixed!

PR #885, opened 2026-01-23
Refactor: Migrate to UV

tobimichael96:
Hi all,

I'm a big fan of UV! It makes things faster and most of the time easier to maintain. 

This is only a draft but it should be funtional. Feel free to take this over or let me know what you would expect!

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -27,18 +27,6 @@ echo_with_timestamp() {
     echo "$(date '+%Y-%m-%d %H:%M:%S') - $1"
 }
 
-# --- NumPy version switching for legacy hardware ---
-if [ "$USE_LEGACY_NUMPY" = "true" ]; then
-    # Check if NumPy was compiled with baseline support
-    if $VIRTUAL_ENV/bin/python -c "import numpy; numpy.show_config()" 2>&1 | grep -qi "baseline"; then
-        echo_with_timestamp "🔧 Switching to legacy NumPy (no CPU baseline)..."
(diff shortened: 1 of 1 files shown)


uv.lock
tobimichael96, line 1: Just as FIY: it's recommended to add this file to git, see https://docs.astral.sh/uv/guides/projects/#uvlock and https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile.

SergeantPanda: Yeah, I did look into it. Not sure we want to use it with the way the images are built though. We could have a github workflow to update the lockfile but I'm not sure I want to worry about messing with that all the time at this point.

SergeantPanda:
There were several 'gotchas' I ran into while getting this to work. Probably spent way too much time on it. 

It does seem to be faster on my local machine to build the image.

uv.lock
SergeantPanda: Open to suggestions on how this should be handled but I'd rather not worry about having to run a command before building the base container everytime. We didn't do that when we used PIP.

tobimichael96: I'm not really sure what speaks against the lock file being in Git to be honest. Can you explain again what makes you think it could lead to problems?

That being said, I don't have a hard opinion on it. For my part, I can follow the logic outlined in the documentation, but that's it.

tobimichael96:
> There were several 'gotchas' I ran into while getting this to work. Probably spent way too much time on it.
> 
> It does seem to be faster on my local machine to build the image.

I see you really did a lot of stuff. Didn't do such a deep dive, but yeah UV should be significantly faster! I experienced that on a lot of projects now. 

I'm glad you like it! :)

Issue #888, opened 2026-01-24
[Feature]: Add option to disable the main profile for an M3U account


dvdblg:
First of all thank you for Dispatcharr, it's really a great project and it's been very useful for managing IPTV.

I love the ability to add multiple profiles for an M3U account. However, I would like the ability to disable the main profile (or, alternatively, set a secondary profile as the main one) for an M3U account. The two use cases I can think of for this are:

- Case 1: I need to change the URL or credentials for a M3U account. As far as I know, I would have to migrate the main credentials to a another set of credentials using [this procedure](https://dispatcharr.github.io/Dispatcharr-Docs/troubleshooting/#i-got-new-credentials-or-url-from-my-provider-what-should-i-do). I think that it would be easier to setup a second profile with the regex and disable the main one.
- Case 2: This is actually my use case. One of my providers right now has problems with `.ts` streams but the streams work fine if I replace `.ts` with `.m3u8?output=hls` in the URL. As far as I know, the only way to do this is through M3U Account's Profile feature, but I can't disable the default profile and use the profile with the regex replace as the main one.

Both of these use cases could also be tackled by adding the "regex replace" also for the default profile, instead of allowing to disable it.

SergeantPanda:
Firstly, thank you for the kind words!

This is not something that we'd add simply because it makes more sense to just update the username and password for the main account. There would be too many other issues to worry about if we disabled the main account.

With the changes to XC account stream hashing a couple of version ago the procedure is very simple.

PR #889, opened 2026-01-24
Enable regex pattern replacement for default M3U profiles

dvdblg:
This PR adds the ability to enable regex pattern replacement for the default profile for M3U accounts.

More info on the use case for this here: #888

### DIFF
--- a/apps/m3u/serializers.py
+++ b/apps/m3u/serializers.py
@@ -1,14 +1,17 @@
-from core.utils import validate_flexible_url
+import json
+import logging
+
 from rest_framework import serializers, status
 from rest_framework.response import Response
-from .models import M3UAccount, M3UFilter, ServerGroup, M3UAccountProfile
-from core.models import UserAgent
+
 from apps.channels.models import ChannelGroup, ChannelGroupM3UAccount
 from apps.channels.serializers import (
     ChannelGroupM3UAccountSerializer,
 )
-import logging
-import json
+from core.models import UserAgent
+from core.utils import validate_flexible_url
+
+from .models import M3UAccount, M3UAccountProfile, M3UFilter, ServerGroup
 
 logger = logging.getLogger(__name__)
 
@@ -34,10 +37,10 @@ class M3UAccountProfileSerializer(serializers.ModelSerializer):
     def get_account(self, obj):
         """Include basic account information for frontend use"""
         return {
-            'id': obj.m3u_account.id,
-            'name': obj.m3u_account.name,
-            'account_type': obj.m3u_account.account_type,
-            'is_xtream_codes': obj.m3u_account.account_type == 'XC'
+            "id": obj.m3u_account.id,
+            "name": obj.m3u_account.name,
+            "account_type": obj.m3u_account.account_type,
+            "is_xtream_codes": obj.m3u_account.account_type == "XC",
         }
 
     class Meta:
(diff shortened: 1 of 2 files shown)


SergeantPanda:
I don't think this is a good idea unfortunately. It is best to just update the credentials for the main account.

Refreshes use the stored credentials so if the URL, username or password are incorrect refreshes wont work anyway.

It's also a bit difficult to see the actual changes because formatting was applied to the code.

dvdblg:
I'm sorry for including the formatting, which made more difficult to read the changes made.

Do you think the other use case for this (mainly changing the stream output format of the provider from TS to HLS) is feasible in other ways?

SergeantPanda:
Sorry I didn't realize it also added functionality for converting to HLS.

Currently it doesn't really make sense to use HLS because we only output TS anyway.

LITUATUI:
> Sorry I didn't realize it also added functionality for converting to HLS.
> 
> Currently it doesn't really make sense to use HLS because we only output TS anyway.

HLS would be great for the AV1 codec, since it's not supported by mpegts.

SergeantPanda:
I think there's a bit of a misunderstanding there. HLS is just the transport,  it doesn't inherently 'support' AV1. What you’re actually looking for is fMP4 (CMAF) support. Even with HLS, if I keep using the standard MPEG-TS segments we use now, AV1 still wouldn't work. Moving to fMP4 is a much bigger architectural shift on the backend than just 'turning on' HLS.

LITUATUI:
> I think there's a bit of a misunderstanding there. HLS is just the transport, it doesn't inherently 'support' AV1. What you’re actually looking for is fMP4 (CMAF) support. Even with HLS, if I keep using the standard MPEG-TS segments we use now, AV1 still wouldn't work. Moving to fMP4 is a much bigger architectural shift on the backend than just 'turning on' HLS.

Curiously, AV1 works partly on Dispatcharr, however, it only allows one connection, it doesn't share the stream with multiple users. The first user can open the stream, the second can't.

AV1_VAAPI example:

`-user_agent {userAgent} -hwaccel vaapi -hwaccel_output_format vaapi -i {streamUrl} -map 0:v -map 0:a -vf "deinterlace_vaapi=rate=field:auto=1,scale_vaapi=format=p010" -c:v av1_vaapi -quality speed -b:v 5M -maxrate 10M -bufsize 20M -hls_segment_type fmp4 -c:a libopus -f webm pipe:1`

SVT-AV1 example:

`-user_agent {userAgent} -hwaccel vaapi -i {streamUrl} -map 0:v -map 0:a -vf "bwdif=1:-1:1,format=p010" -c:v libsvtav1 -preset 8 -crf 40 -hls_segment_type fmp4 -c:a libopus -f webm pipe:1`

It would be awesome to get support for the AV1-Opus combo, since they are the best non-proprietary codecs. Moreover, the AVC and HEVC hardware encoders of AMD suck, but AV1 is OK.

Screenshot_20260324_225126

SergeantPanda:
The reason it only works for the first connection is almost certainly because only the first client would get the proper headers.
Do you have a provider that outputs av1?

LITUATUI:
> The reason it only works for the first connection is almost certainly because only the first client would get the proper headers. Do you have a provider that outputs av1?

No, I get my channels locally from TVHeadend (DVB-C tuners), but I would like to transcode them from AVC-AAC to AV1-Opus, so I can watch TV on slow Internet connection when I'm not home. I got a AMD Ryzen 5 7640HS mini PC recently just because it can transcode to AV1.

PR #894, opened 2026-01-26
Feature/771 auto matching improvements

CodeBormen:
### User-Configured EPG Settings 

Adds new settings section 'EPG Settings', containing:

**Ignore Prefixes**: Removed from START only (e.g., Prime: ESPN → ESPN)
**Ignore Suffixes**: Removed from END only (e.g., ESPN HD → ESPN)
**Ignore Custom Strings**: Removed from ANYWHERE (e.g., ABC 24/7 News → ABC News)
Uses **exact string matching** 

Function runs before existing normalization, then funnels into the normalization process for:

- Lowercasing the string
- Removing bracketed/parenthesized text
- Removing punctuation
- Filtering out COMMON_EXTRANEOUS_WORDS (tv, channel, hd, uhd, east, west, etc.)
- Collapsing extra spaces

Note that since this feature introduces EPG matching control to the user, the current extraneous word filter is not strictly needed and could be removed to give granularity to the end user in the case they need to maintain a word for accurate matching (HD or region [east, west] may be needed in the channel name to accurately match EPG data).

Extensive testing was performed for this change to verify proper backend storage and retrieval (including edge cases like special characters) as well as frontend UI testing. This function was specifically written to work ahead of the existing normalization feature to not cause unintended behavior or breakage.

epg_ignore_feature

### DIFF
--- a/frontend/src/components/modals/__tests__/EPGMatchModal.test.jsx
+++ b/frontend/src/components/modals/__tests__/EPGMatchModal.test.jsx
@@ -40,6 +40,98 @@ vi.mock('../../../store/settings', () => ({
   }),
 }));
 
+vi.mock('@mantine/core', () => {
+  const React = require('react');
+
+  const RadioComponent = ({ label, value, checked, description, groupValue, groupOnChange }) => {
+    const isChecked = checked !== undefined ? checked : groupValue === value;
+    const handleChange = groupOnChange || (() => {});
+
+    return (
+      <label>
+        <input
+          type="radio"
+          value={value}
+          checked={isChecked}
+          onChange={() => handleChange(value)}
+          aria-label={label}
+        />
+        {label}
+        {description && <span>{description}</span>}
+      </label>
+    );
+  };
+
+  RadioComponent.Group = ({ children, value, onChange, label }) => {
+    // Clone children and inject group props
+    const enhancedChildren = React.Children.map(children, (child) => {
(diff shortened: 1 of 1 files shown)


CodeBormen:
My initial thought actually was a pop-up modal but I wanted to respect the existing interfaces. If you think a pop-up modal is a good way to go, I can make those changes and test it, then you can re-review and see if you like it.

CodeBormen:
Alright, let's see if you like it as a modal! I do think it is a much better fit. User-configured input is saved into the modal for users who do repeat matching.

If you do end up liking it, I think it would be a good idea to move some of the functionality from the existing normalization function into this new modal to allow for more control over matching. Some of the current normalization could be more strongly influenced by the end user (matching aggression slider, COMMON_EXTRANEOUS_WORDS could be totally removed with the addition of this modal, even the removal of bracketed/parenthesized text could be removed). Can potentially address this separately.

modal

CodeBormen:
For sure, I can do something that will make it clearer that the user can proceed without any additional config

CodeBormen:
Thanks for the suggestion, think this works much better


epg_radial


epg_radial_advanced

SergeantPanda:
I like this a lot more. Thank you for the time spent on this and being willing to change!

CodeBormen:
Of course, thanks for letting me work on something cool!

PR #895, opened 2026-01-27
Fix User-Agent not displaying in Stream Profile dialog

### LINKED ISSUE #650: [Bug]: Chosen User-Agent not showing in Stream Profile dialog
### Describe the bug

When creating a new Stream Profile, or editing an existing one, the dropdown for User-Agent does not show the selected option.

I have confirmed that the chosen option does indeed change in parameters when a stream is running, it just doesn't visually show which User-Agent is selected.

https://github.com/user-attachments/assets/c30be526-7873-49af-813a-253f12037ddd

1. Go to Settings -> Stream Profiles
2. `Add Stream Profile` or edit an existing profile
3. If editing existing profile, you will see that the chosen User-Agent dropdown is empty. Try selecting any User-Agent
4. If creating, select any User-Agent

### Dispatcharr Version

0.11.2

### Docker Version

Docker version 28.5.2, build ecc6942

Not using docker compose.

### Client Information

Tested on:
- Brave
- Microsoft Edge

CodeBormen:
The Select component was using register() which does not work with mantine Select. Changed to use setValue with onChange handler, and convert user_agent to string in defaultValues to match option values.

Fixes #650

### DIFF
--- a/frontend/src/components/forms/StreamProfile.jsx
+++ b/frontend/src/components/forms/StreamProfile.jsx
@@ -22,7 +22,7 @@ const StreamProfile = ({ profile = null, isOpen, onClose }) => {
       command: profile?.command || '',
       parameters: profile?.parameters || '',
       is_active: profile?.is_active ?? true,
-      user_agent: profile?.user_agent || '',
+      user_agent: profile?.user_agent ? `${profile.user_agent}` : '',
     }),
     [profile]
   );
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you for this. It hasn't been fixed because it is probably going to be removed since there is no code to support it. I don't think there is really a need for a default user agent on a stream profile.

SergeantPanda:
I'm going to close this. I reworked the form to add a drop-down for command selection and fixed the user-agent loading at the same time. 

I still might remove it altogether since we don't use it anywhere. I just wanted to let you know.

Issue #866, opened 2026-01-17
[Feature]: Search/Filter/Sort streams by TVG-ID


bjjnbbq-ninja:
### Is your feature request related to a problem?

In the Channels view, the left Channels panel shows an EPG column which identifies the TVG-ID information. When adding additional streams to a channel there is no way currently to identify the TVG-ID of the Stream in the right Streams panel. It is currently necessary for me to create a new temp channel to see what the TVG-ID is so I know which existing channel to add streams to.

For example, I have the following Channels created as they all have different EPG information:
- HGTV (us): `TVG-ID = hgtv.us`
- HGTV (us-west): `TVG-ID =  hgtvwest.us`
- HGTV (ca): `TVG-ID = Home.and.Garden.Television.ca2`
- HGTV (uk): `TVG-ID = hgtv.uk`
- HGTV +1 (uk): `TVG-ID = hgtv.uk.plus1`
- HGTV (au): `TVG-ID = hgtv.au`

I set them up under a single M3U source. I then set up additional M3U sources for redundancy and would like to add them to the respective channels already existing but there is no way to easy identify the TVG-ID of the stream.

### Describe the solution you'd like

Describe here.
Under the Channels screen, the right side Streams panel has column headers for Name, Group, and M3U.

This request is to add TVG-ID as a column so users can Search/Filter/Sort similar to the other columns mentioned. Being able to quickly identify the TVG-ID information in the Stream panel would make it very fast and efficient when adding new Streams to existing Channels.

### Additional context

Nothing Extra

CodeBormen:
FYI I am going to work on implementing this. I initially thought it would not be possible since EPG matching does not happen until streams are converted to channels, but after reviewing code it turns out the TVG-ID data is accessible in the API on M3U import.

SergeantPanda:
This is available in dev.

PR #904, opened 2026-01-28
Feature/866 Add TVG-ID column with filter and sort to Streams table

CodeBormen:
Changes:

- frontend/src/components/tables/StreamsTable.jsx

Added TVG-ID column to streams table
Added search filter input in column header
Added sort button for ascending/descending order

- apps/channels/api_views.py

Added TVG-ID to filterable fields
Added TVG-ID to sortable fields
Enables backend support for search and sorting

Users can now view, search, and sort streams by their TVG-ID directly in the Streams table. Fulfills #866

### DIFF
--- a/frontend/src/components/tables/StreamsTable.jsx
+++ b/frontend/src/components/tables/StreamsTable.jsx
@@ -248,7 +248,7 @@ const StreamsTable = ({ onReady }) => {
 
   // Column visibility - persisted to localStorage
   // Default visible: name, group, m3u
-  // Default hidden: tvg_id, source_info
+  // Default hidden: tvg_id, stats
   const DEFAULT_COLUMN_VISIBILITY = {
     actions: true,
     select: true,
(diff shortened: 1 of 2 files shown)


CodeBormen:
I thought the same thing as I was making this, that it would be much better if there was a column selector. If that's something you want me to work on let me know, otherwise if you already have plans on that cool, feel free to use this PR to add TVG-ID as a selectable column along with the rest

SergeantPanda:
I have only thought about wanting the ability but haven't really put a lot of thought into how exactly I wanted to implement it. 

I was thinking adding a 3 dot menu to the far right side of the stream table and an option in the menu to add/remove columns.

If you are bored and want to take that on, that would be OK.

CodeBormen:
Added the three-dot menu as discussed, introducing visibility filters for the columns. The original columns are set as default (name, group, m3u) with tvg-id as an optional (default hidden) column. I also took this opportunity to add some stream stats (Source Info) as an additional optional column to expose some of the source info to the end user. The tooltip for Stream Info provides useful info about the source and will be good context to users about how they may need to set up stream profiles depending on their downstream client capabilities.

Tested this from a fresh install as well as version upgrade without issue.

Hope you like it, put my heart into this one! <3

CodeBormen:
visibility_toggle

CodeBormen:
All good, just happy to contribute!

CodeBormen:
Changes look good to me!

PR #905, opened 2026-01-28
Tests/frontend unit tests

nick4810:
- hooks/
- store/

Included some slight refactoring for readability.

Also removed logic in a few places that wasn't fully implemented.


SergeantPanda (approved):
Looks good. Thanks again Nick!

PR #919, opened 2026-01-31
CONTRIBUTING.md  / Test Execution Guide

jcasimir:
In order to get more folks contributing to the project, I thought it'd be valuable to put together this CONTRIBUTING.md. In particular, it took me a bit to figure out how to get the tests running as expected with the ENV variable and everything. I hope this is helpful!


SergeantPanda:
Thank you for this. 

Can you change the part about installing python packages with requirements.txt as that will not be the case with the next version (we switched to UV). If you want to remove confusion until the next version, you could send this to the dev branch.

jcasimir:
Replacing with a cleaner PR that contains only documentation changes (no config file modifications)

PR #920, opened 2026-01-31
Feature/xc auth delay

jrberenguer:
# Implement per-account Auth Delay for Xtream Codes

## 📝 Description

This PR refactors the Xtream Codes rate limiting protection to be configurable on a per-account basis. It introduces a custom delay that occurs immediately after authentication and before the first resource request.

## ❌ The Problem

Some Xtream Codes providers implement aggressive rate limiting. When Dispatcharr authenticates and immediately requests the categories list, these servers return an **Error 844 (Rate Limit Exceeded)**. A global environment variable was previously considered, but per-account configuration is more flexible as different providers have different thresholds.

## ✅ The Solution

Instead of a global setting, we now have a per-account `xc_auth_delay` field. This allows users to fine-tune the delay for specific providers that are more restrictive, without affecting others.

### Backend
- **apps/m3u/models.py**: Added `xc_auth_delay` field (FloatField, default 0) to `M3UAccount`.
- **apps/m3u/serializers.py**: Exposed the new field in the `M3UAccountSerializer`.
- **apps/m3u/tasks.py**: Modified the `refresh_m3u_groups` task to respect the per-account delay.
- **apps/m3u/migrations/0019_m3uaccount_xc_auth_delay.py**: Added database migration for the new field.

### Frontend
- **frontend/src/components/forms/M3U.jsx**: Added a new "Auth Delay (seconds)" input field in the Xtream Codes section of the M3U account form.

### Documentation
- **CHANGELOG.md**: Added a description of the per-account rate limiting feature.


SergeantPanda:
This is closer. Thank you for resubmitting. Instead of a new database field, please use the custom_properties jsonb field that we already have for this sort of thing.

SergeantPanda:
Is this ready for me to review again?

jrberenguer:
> Is this ready for me to review again?

yes, please! :)

SergeantPanda:
Ok, I played around with this and looked over the code.

Overall the code didn't have any issues but I saw room for improvement on how we actually handle authentication during M3U refreshes and was able to just eliminate the extra API call that I believe was causing the rate limiting. 

This is available in the dev build now if you want to verify it does in fact fix your issue.

If it does not, there is still one more tweak I can make. Please let me know.

I appreciate the time you spent on this PR though.

jrberenguer:
Great! I tried the dev branch and it works for me too. Thank you so much!

SergeantPanda:
Awesome thank you for letting me know! Glad it works. This approach will be better in the long wrong.

PR #922, opened 2026-01-31
Add configurable navigation ordering and visibility

jcasimir:
I'd like to build plugins that have the ability to register their own pages in the UI. Before that can work, I needed the nav to be built dynamically. I thought it would be cool for users to be able to reorder and hide pages from the nav as well. Except the Setting page can't be hidden. Added a bit of persistence storage so things would remain in place across devices.

## Summary
- Allow users to reorder sidebar navigation items via drag-and-drop in Settings
- Allow users to hide navigation items (except Settings which is always visible)
- Preferences persist to the backend via the User.custom_properties JSONField

### P1 (Critical)
- Replace undefined `@swagger_auto_schema` with `@extend_schema` in api_views.py
- Add `custom_properties` validation with size limits (10KB max)
- Add merge semantics in serializer `update()` to prevent data loss on PATCH

### P2 (Important)
- Add debouncing (800ms) for drag operations to reduce API calls
- Add error handling with notifications for visibility toggle
- Use store getters in Sidebar instead of direct property access
- Cache `hiddenNav` before render loop to avoid repeated getter calls

### P3 (Nice-to-have)
- Add `canHide: false` property to settings nav item for explicit control
- Fix admin check to use `>=` instead of `==` for user level comparison
- Fix ESLint unused variable warnings

## Test plan
- [x] Backend tests passing (7/7 user preference tests)
- [x] ESLint clean on modified files
- [x] Manual testing of drag-and-drop reordering
- [x] Manual testing of hide/show navigation items
- [x] Verify settings page cannot be hidden

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Configurable sidebar navigation ordering and visibility — Thanks [@jcasimir](https://github.com/jcasimir)
+  - Sidebar nav items can be reordered via drag-and-drop in Settings → UI Settings → Navigation.
+  - Individual nav items can be hidden from the sidebar using the eye toggle. Hiding an item preserves its position in the order.
+  - A "Reset to Default" button restores the role-appropriate default order and clears all hidden items.
+  - Order and visibility are saved per-user with optimistic updates and automatic rollback on failure. Changes appear in the sidebar immediately without a page reload.
+  - Admin users see a grouped navigation: flat items (`Channels`, `VODs`, `M3U & EPG Manager`, `TV Guide`, `DVR`, `Stats`, `Plugins`) plus collapsible `Integrations` (Connections, Logs) and `System` (Users, Logo Manager, Settings) groups. The `System` group cannot be hidden.
+  - Non-admin users see `Channels`, `TV Guide`, and `Settings`, with the `Settings` item not hideable.
 - Unit tests for `NotificationCenter`, `NotificationCenterUtils`, and `M3URefreshNotification` components. — Thanks [@nick4810](https://github.com/nick4810)
 - Unit tests for DVR port resolution (`build_dvr_candidates`) and selective Redis flush behavior in modular mode. — Thanks [@CodeBormen](https://github.com/CodeBormen)
 - Floating video player now displays the channel, stream, or VOD title in the player header bar. Title is passed through from all preview entry points: channel table, stream table, stream connection card, guide, DVR, recording cards, recording details modal, VOD modal, and series modal.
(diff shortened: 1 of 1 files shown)


SergeantPanda (approved):
Hey @jcasimir — thanks for this PR, the feature is solid and we're merging it. Before doing so I did a full review pass and made several changes on top of your branch. Here's a summary of everything that was fixed:

🔴 Security: Privilege Escalation via /me/
The new PATCH /api/accounts/users/me/ endpoint used the full UserSerializer without any field restrictions. This meant any authenticated user could send {"user_level": 10, "is_staff": true, "is_superuser": true} and silently escalate themselves to admin.

Fixed by adding an ALLOWED_FIELDS allowlist to the /me/ view — only custom_properties, first_name, last_name, email, and password are accepted. Anything else returns HTTP 400. Privilege-sensitive fields remain exclusively writable via the admin-only PATCH /api/accounts/users/{id}/ endpoint. A test was added to verify this.

🟠 Two Runtime Errors
channels is not defined — getOrderedNavItems() was called with channels but the Sidebar only has channelIds (an array from the store). The badge also used Object.keys(channels).length which doesn't work on an array. Fixed in both Sidebar.jsx and navigation.js.

onLogout is not defined — The logout button referenced onLogout but the auth store exposes logout. Fixed.

Both of these were from changes since this PR were submitted, not from bugs you introduced

🟡 Double Notification on Save Failure
API.updateMe had a try/catch that caught errors, showed a notification, then re-threw — and the callers also showed a notification on catch, producing two error toasts for a single failure. Removed the inner try/catch from API.updateMe so errors propagate cleanly to the caller once.

🟡 Double-Merge Race Condition
updateUserPreferences was pre-merging custom_properties client-side before sending the full merged object to the backend, which then merged it again against the DB. In a multi-tab or multi-session scenario, session B's write could quietly overwrite keys that session A had just set — because both sessions read the same base and sent a fully-merged payload.

Fixed by sending only the delta to the backend ({ custom_properties: preferences }) and letting the backend merge authoritatively against the stored DB value. The frontend still applies an optimistic update locally and reverts on failure.

Navigation Groups (Integrations / System)
Your navigation.js had a flat structure. The dev branch sidebar had grouped navigation — Integrations (Connections, Logs) and System (Users, Logo Manager, Settings) as collapsible groups. These were restored with group support added to navigation.js, DEFAULT_ADMIN_ORDER updated accordingly, and the NAV_ITEMS flat users/logos entries removed in favor of the group definitions. The System group has canHide: false.

Housekeeping
Rebased onto current dev (the branch was 329 commits behind, causing GitHub to show ~309 files changed in the PR diff — now shows 12)
VALID_NAV_ITEM_IDS in serializers.py is now actually used in validation (it was defined but the check was never wired up)
Stale nav item IDs are scrubbed from navOrder/hiddenNav on the next preference save so the custom_properties field doesn't accumulate garbage
Backend test moved to tests folder to match project convention
Frontend tests updated to reflect the grouped nav structure (Integrations/System instead of flat Users/Logos/Settings)

SergeantPanda:
Oh, also, I moved the settings under the UI settings.

PR #925, opened 2026-02-01
Add server failover by URL

### LINKED ISSUE #504: [Feature]: Support Multiple Provider URLs with Failover for M3U Accounts

Sometimes the provider’s URL cannot be resolved due to various reasons, such as:

- Bad routing
- DNS issues
- Being blocked by the ISP
- Other network-related disruptions

When this happens, Dispatcharr cannot sync with the provider, even if the provider might still be accessible via an alternative URL.

Some IPTV providers offer multiple URLs that point to the same service as a way to mitigate accessibility issues. It would be useful if Dispatcharr supported adding multiple URLs when setting up an M3U account.

This would allow the system to:
- Attempt the primary URL first.
- If it fails (e.g., timeout, unreachable, DNS error), automatically fall back to the next URL in the list.
- Continue through the list until a working URL is found.

This failover mechanism would improve reliability when syncing with providers, reduce downtime caused by network issues, and enhance the overall user experience.

### Additional context

Nothing Extra

### LINKED ISSUE #561: [Feature]: Ability to treat M3U profiles as "fallbacks"

M3U profiles can be used within Dispatcharr to perform Regex transformations on M3U URLs, which is particularly useful when providing a second set of credentials (i.e. to bypass concurrent stream limits).

Dispatcharr uses its own source of truth for keeping track of how many streams are active for a given profile, meaning that if a stream is being watched from outside of Dispatcharr, it is unaware of this and will fail to load the stream - crucially, without attempting to use an alternative M3U profile as a fallback.

I think that it would be useful to provide a toggle somewhere (globally? per-profile basis?) that allows for M3U profiles to be used as fallbacks, so that if a stream fails to initialise or dies at some point, the next configured profile is used instead. This would allow for much more flexibility in the use of this feature (e.g. specifying alternative provider endpoints) while minimising the need to, for instance, create a separate M3U provider entry completely and reconfigure channels.

### Additional context

I imagine that this feature could be hooked into the existing fallback logic for selecting between channel stream entries with relative ease (though I don't have any knowledge of the codebase - so may be completely wrong there!)

Might also need to consider how priorities would work when choosing the fallback to use, e.g. should the next profile be tried ahead of the next stream in the channel, or vice-versa? Perhaps this could also be a configuration option!

jcasimir:
## Summary
- Add failover URL support to M3U accounts - users can enter pipe-separated URLs (e.g., `http://primary:8080|http://backup:8081`)
- Backend stores URLs as JSON array and cycles through them during fetch attempts
- Add URL validation button to test all URLs before saving
- Improve failover UX with better timeouts, status messages, and progress indicators

Closes #504
Closes #561

### Backend
- Change `server_url` from URLField to JSONField for array storage
- Add utility functions for parsing/formatting pipe-separated URLs
- Add `validate-urls` API endpoint to test URLs (max 10 URLs per request)
- Implement failover logic in XCClient with 10s timeout
- Remember successful URL for subsequent API calls
- Send WebSocket status updates during connection/transfer

### Frontend
- Add Validate button in M3U form with parallel URL testing
- Show countdown timer during validation (10s timeout)
- Display validation results below URL field
- Add tooltip explaining failover URL format
- Show failover progress in M3U table ("Trying 1/3...", "Transferring...")
- Add spinners to status messages

## Code Review Fixes

Multi-agent code review identified and fixed the following issues:

### Migration Safety (P1)
- **Fixed operation order**: Convert data to JSON format BEFORE changing field type (prevents PostgreSQL cast errors)
- **Performance**: Use `bulk_update` instead of individual saves
- **Rollback documentation**: Note that rollback only preserves first URL

### XCClient (P1)
- **Extended failover to ALL API methods**: Previously only `authenticate()` used failover; now `get_live_categories()`, `get_vod_streams()`, `get_series()`, etc. all use failover when multiple URLs configured

### Frontend (P1)
- **Memory leak fix**: Add `useEffect` cleanup for countdown intervals on component unmount
- **State reset**: Clear validation state when modal closes

## Test plan
- [x] Single URL (existing behavior) - stores as `["url"]`, works normally
- [x] Multiple URLs - stores as array, displays as pipe-separated
- [x] Validate button - tests all URLs, shows results
- [x] Primary works - no failover needed, immediate success
- [x] Primary fails, secondary works - shows failover progress, succeeds
- [x] All URLs fail - proper error after cycling through URLs
- [x] Edit existing account - array displays as pipe-separated, saves correctly
- [x] All 52 tests passing


jcasimir:
I actually spent most of the time/effort on the UI/UX, which I think came out pretty nicely! Looking forward to feedback.

jcasimir:
Hi team. Is it fair to assume you don't want feature PRs like this?

diablo81588:
@SergeantPanda can you please review this much requested feature? Thanks for your work on it @jcasimir!

SergeantPanda:
Hey I'm sorry to leave you hanging for so long on these.

This one is a tough one for me if I'm being honest because I'm not sure on this feature yet. 

At the same time, this seems like a lot of code changes for this too and I need time to go through it. 

In regards to the other PR's you submitted as well, it seems a bunch were submitted in a short period of time and were done with AI (which I am not completely against, don't get me wrong). BUT they were submitted so quickly I am skeptical that they were thoroughly tested to not cause regressions.

That is the full transparent reason why these have sat so long. I do not have enough time to go through all of the changes Claude made to this project for a feature that I'm not sure we need.

That doesn't mean that it isn't good though. Hopefully you understand.

SergeantPanda:
I'm going to close this. It has conflicts with several other changes at this point.

I am not completely opposed to having fail-over urls but am not sure how we want to handle it.

It seems this PR does not do anything for streaming and would only help if the failure occurs during M3U account refresh (which imo is a good time to let it fail, helps users see there is a problem).

With the recent changes to how XC builds urls and doesn't require a rehash when urls change, failover there is less beneficial.

PR #929, opened 2026-02-01
Fix uWSGI segfaults by disabling threads when using gevent

jcasimir:
## Summary

- Disables threading in dev and debug uWSGI configs when gevent is enabled
- Mixing threading and gevent concurrency models causes workers to crash with segmentation faults, particularly on ARM64/Python 3.13

## Problem

The dev and debug uWSGI configs had both `threads`/`enable-threads` AND `gevent` enabled simultaneously. This combination causes uWSGI workers to crash in a loop:

```
DAMN ! worker 4 (pid: 1868) died, killed by signal 11 :( trying respawn ...
```

The production `uwsgi.ini` was already correct (gevent without threads).

## Why this fix

From [How to run uWSGI](https://blog.ionelmc.ro/2022/03/14/how-to-run-uwsgi/):

> "In general the most reliable concurrency model is processes, with no threads... I wouldn't enable the gevent plugin - you're just asking for trouble with all that monkey-patching."

## Solution

Commented out `threads` and `enable-threads` in the dev and debug configs, with references to the upstream issues.

## References

- https://blog.ionelmc.ro/2022/03/14/how-to-run-uwsgi/
- https://github.com/gevent/gevent/issues/1784
- https://github.com/unbit/uwsgi/issues/2457

## Test plan

- [x] Verified dev container starts without segfaults on ARM64 Mac
- [x] Verified HTTP 200 response from frontend
- [x] Verified no worker crash/respawn loop in logs

### DIFF
--- a/docker/uwsgi.debug.ini
+++ b/docker/uwsgi.debug.ini
@@ -28,8 +28,6 @@ static-map = /static=/app/static
 
 # Worker configuration
 workers = 1
-threads = 8
-enable-threads = true
 lazy-apps = true
 
 # HTTP server
--- a/docker/uwsgi.dev.ini
+++ b/docker/uwsgi.dev.ini
@@ -28,10 +28,8 @@ vacuum = true
 die-on-term = true
 static-map = /static=/app/static
 
-# Worker management (Optimize for I/O bound tasks)
+# Worker management
 workers = 4
-threads = 2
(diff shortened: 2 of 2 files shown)


SergeantPanda:
Thank you for your submission. I have been meaning to make this change for a long time.

PR #937, opened 2026-02-04
feat: Add Redis authentication support for modular deployment

CodeBormen:
Add support for authentication when connecting to external Redis instances in modular deployments. This enables secure Redis deployments using either password-only authentication (Redis <6) or username + password authentication (Redis 6+ ACL).

Changes:
- Add REDIS_PASSWORD and REDIS_USER environment variables
- Implement URL encoding for special characters in passwords
- Update all Redis connection points to support auth
- Add documentation and examples to docker-compose.yml
- Maintains full backward compatibility (empty defaults = no auth)

All authentication mechanisms have been fully tested including password-only authentication, Redis 6+ ACL authentication with username + password, volume-mounted configuration files, and special character handling.

Existing deployments are not effected, authentication support is entirely opt-in using docker-compose.

Also, acknowledging the fact that the docker-compose file for modular deployments has grown disorganized with the addition of auth support, the docker-compose.yml file was re-formatted for better visibility in configuration. This seemed like a better strategy than mandating a .env file.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -65,6 +65,7 @@ services:
     #  - /dev/dri:/dev/dri  # For Intel/AMD GPU acceleration (VA-API)
 
     # NVIDIA GPU Support (requires NVIDIA Container Toolkit)
+    # Uncomment the following lines for NVIDIA GPU support
     #deploy:
     #  resources:
     #    reservations:


virtualm2000:
Does it support tls connection? Thanks.

CodeBormen:
@virtualm2000 this implementation does not implement TLS connections, only authentication, so if your path to Redis has connections over insecure networks your credentials and data would still be exposed. It is possible that this could be supported one day but would require mutual certificate verification which is not a trivial amount of work.

virtualm2000:
> @virtualm2000 this implementation does not implement TLS connections, only authentication, so if your path to Redis has connections over insecure networks your credentials and data would still be exposed. It is possible that this could be supported one day but would require mutual certificate verification which is not a trivial amount of work.

I meant only TLS connection without mutual authentication.

If i understood correctly the redis lib supports ssl only without cert based authentication. 

`r = redis.Redis(
    host='localhost',
    port=6666,
    ssl=True,
    ssl_cert_reqs="none",
)`

https://redis.readthedocs.io/en/stable/examples/ssl_connection_examples.html

CodeBormen:
Yes, both Redis and Postgres support TLS without cert-based auth. Do you have a significant use-case for this? It is something that I could possibly discuss adding as an opt-in feature with the developers/maintainers but for folks running this on low-resource machines the addition of TLS overhead could potentially cause performance issues.

virtualm2000:
In my setup the redis server runs outside the cluster as it is used by different services. I've already run some tests with hard coded values and I see no degrading experience. I'll make a pr to your branch with proper modifications to check it out.

CodeBormen:
If adding TLS I would prefer if there were a feature request opened for it and then a separate PR for that feature request. If TLS is being added, it should be added for both Redis and Postgres, it would not make a lot of sense to implement for only one of the two. If you want to open a feature request, I will work on adding TLS without mutual auth for both.

virtualm2000:
> If adding TLS I would prefer if there were a feature request opened for it and then a separate PR for that feature request. If TLS is being added, it should be added for both Redis and Postgres, it would not make a lot of sense to implement for only one of the two. If you want to open a feature request, I will work on adding TLS without mutual auth for both.



> If adding TLS I would prefer if there were a feature request opened for it and then a separate PR for that feature request. If TLS is being added, it should be added for both Redis and Postgres, it would not make a lot of sense to implement for only one of the two. If you want to open a feature request, I will work on adding TLS without mutual auth for both.

Brilliant, thank you. #950

PR #938, opened 2026-02-04
EPG program preview in channel form

FiveBoroughs:
Show a live "Now Playing" program preview when selecting an EPG channel in the channel create/edit modal.

  - Extracted ProgramPreview that was added in the stats tab into a shared component from StreamConnectionCard
  - Added epg_data_ids support to CurrentProgramsAPIView for direct EPG lookups without a saved channel
  - Backend uses a byte-offset index over the XMLTV file (EPGSource.programme_index) for sub-millisecond current-program lookups without parsing the full file into the DB; per-channel interleaved_channels keeps the fast path for well-behaved channels
  - When the source has no index yet, the API dispatches an async build_programme_index_task and returns a parsing placeholder for the frontend to retry against
  - Frontend retries 3x at 10s intervals when the backend signals programs are being parsed (Unsure what the value should be)
  - Checks Redis task lock before dispatching to avoid duplicate parse tasks
  - Dummy EPG sources are skipped

## Related Issue

Closes #

## How was it tested?

- Backend: `python manage.py test apps.epg.tests.test_programme_index apps.epg.tests.test_current_programs_api`, 24/24 pass. Suite includes new tests for the perf changes: `test_per_channel_interleaved_marking`, `test_no_index_dispatches_build_and_returns_timeout`, `test_no_index_skips_dispatch_when_lock_held`.
- Frontend: `npx vitest run ProgramPreview ChannelEpgPreview Channel.test`, 74/74 pass. Covers loading, retry, fetched state, and the `useEpgPreview` hook in isolation.
- Lint: `npm run lint` shows no new errors in files modified by this PR. The refactor extracting `useEpgPreview` into `frontend/src/hooks/useEpgPreview.jsx` cleared the only react-refresh warning it introduced.
- OpenAPI: `manage.py spectacular --validate` confirms `/api/epg/current-programs/` with `CurrentProgramsRequest` (including the new `epg_data_ids` field) appears in the schema.
- Manual: tested against `https://epgshare01.online/epgshare01/epg_ripper_ALL_SOURCES1.xml.gz` (~22k channels, 1.7 GB). Lookups sub-millisecond on well-behaved channels; deferred build dispatch verified for unindexed sources.

## Checklist


SergeantPanda (commented):
I'm not really sure there is much benefit to this. What is your proposed use-case?

FiveBoroughs:
A lot of EPG sources have empty or wrong data, and currently you only find out when you check your client/player and see it's wrong or empty, go back to dispacthcarr edit the channel, pick another epg source, refresh epg in plex/jellyfin settings, and open the guide, often multiple times back and forth, and for every channel. Massive waste of time.

With this, you pick an EPG source in Dispatcharr and can immediately see what program data it has, if it's wrong pick another one, save. EPG is the only feature where you pick something and have to hope it's correct; unlike streams, logos, etc. where you can preview before saving. This brings EPG in line with the rest of Dispatcharr: pick, preview that it's correct, then save.

Additionally this is a minimal change, I was careful to reuse your existing ProgramPreview component and API functions

FiveBoroughs:
Most of the LoC count isn't new code but moving the existing program preview UI from StreamConnectionCard into a dedicated component so it can be shared across interfaces.

CodeBormen:
I definitely would be interested in some way to preview EPG data so that it could be verified before channel binding

SergeantPanda:
I did play with this a bit and I if we were to accept a feature like this it'd need to be faster.

It takes far too long for the parsing of the EPG to happen to show what the current program is. 

I new function would probably need to be created that ignores everything but that tvg-id as well as be able to filter down to only current time. I am not sure that lxml would support such a thing but if it does and it can do it efficiently I would consider this a viable feature. 

I personally still don't know that it adds enough value though.

FiveBoroughs:
Thanks for the feedback. I reworked it to fix the performance problem.

When I originally built this I was testing with smaller guides (4MB, 147MB) where parse times were 0.25s and 2s respectively, fast enough that you barely notice. The original approach fell apart on large files (~11GB), where it took 118s per channel. Completely unusable, agreed.

I replaced the full-parse approach with a byte-offset index. After EPG download, `build_programme_index` scans the raw XML in 8MB binary chunks looking for `<programme` tags and recording where each channel's data starts in the file. The index stored in Redis looks like:

```json
{
  "BBC.One.uk": [14208],
  "ITV1.uk": [982016],
  "Channel4.uk": [1843200, 5765120]
}
```

Values are byte offsets into the XML file. Most channels have one contiguous block, some are split across the file (like `Channel4.uk` above with two blocks).

At lookup time, `find_current_program_for_tvg_id` grabs the offsets for the requested `tvg_id`, seeks directly to those file positions, and parses only the matching `<programme>` elements with lxml until it finds one covering the current time. No DOM tree, no scanning the rest of the file.

The API uses a 3-tier lookup:
1. DB query (instant if programs already parsed for a mapped channel)
2. Byte-offset seek into the XML file (1–10ms with index)
3. Sequential `iterparse` fallback with 10s timeout (only on cold start when no index exists yet; triggers async index build)

While testing I ran into an issue where the entire Redis DB was being wiped every time a new process started (#980). That included the index we just built, which explained the intermittent empty previews. Fixed in this branch by removing the `flushdb()` call from `RedisClient.get_client()`.

Benchmarks (run inside the Docker container, 5 EPG files from 8MB to 11GB):

| File | Channels | Index build | Index in Redis |
|---|---|---|---|
| 8 MB | 260 | 0.03s | 8 KB |
| 147 MB | 11,656 | 0.58s | 360 KB |
| 525 MB | 15,453 | 1.83s | 341 KB |
| 1.7 GB | 22,886 | 4.01s | 812 KB |
| 11 GB | 32,161 | 18.01s | 736 KB |

Indexed lookup times (3 channels per file, sampled from start/middle/end):

| File | Lookup 1 | Lookup 2 | Lookup 3 |
|---|---|---|---|
| 8 MB | 3.9ms | 2.7ms | 1.1ms |
| 147 MB | 1.6ms | 1.2ms | 0.4ms |
| 525 MB | 1.2ms | 0.9ms | 0.5ms |
| 1.7 GB | 6.5ms | 6.0ms | 1.2ms |
| 11 GB | 0.6ms | 1.7ms | 0.8ms |

Without the index (sequential `iterparse` scan, 10s timeout):

| File | Ch near start | Ch in middle | Ch near end |
|---|---|---|---|
| 8 MB | 64ms | 67ms | 67ms |
| 147 MB | 58ms | 645ms | 1,273ms |
| 525 MB | 297ms | 2,821ms | 5,371ms |
| 1.7 GB | 713ms | 8,843ms | TIMEOUT |
| 11 GB | 4,777ms | TIMEOUT | TIMEOUT |

The index build is a one-time cost per EPG refresh (~18s for 11GB). Once built, every lookup is 0.4–6.5ms regardless of file size or channel position. The sequential fallback only runs if a user opens the channel form before the index finishes building; for channels it can't reach in 10s it returns `parsing: true` and the frontend retries with backoff.

Since the index lives in Redis (ephemeral), it's lost on container restart. On first run of `scan_and_process_files` after boot, any active EPG source missing its index gets a `build_programme_index_task` queued as a background Celery task. A Redis lock (`building_programme_index_{source_id}`, 300s TTL) prevents duplicate builds if the scan fires again before the build finishes. No-ops if the file hasn't been downloaded yet.

Other changes:
- Batch endpoint capped at 50 IDs per request
- `select_related('epg_source')` to avoid N+1 queries
- Dummy EPG sources skipped
- Frontend retry changed from 3x10s fixed to exponential backoff (3s initial, 1.5x multiplier, 15s cap, 3-minute deadline)

No new dependencies, no schema changes. The preview only activates when a user selects an EPG entry in the channel form.

SergeantPanda:
This is an interesting way to handle this. I really appreciate the time invested in thinking through this.

I wonder if instead of being indexed on every boot if this makes sense to have in the table with the tvg-id for the epg source? It would then only need to be updated when the epg refreshes.

Just off the top of my head. There may be a valid reason to not do that.

FiveBoroughs:
Yeah, that makes sense. Two ways to do it:

**JSON column on `EPGSource`**: store the full `{tvg_id: [byte_offsets]}` dict as a single JSON blob on the source row. One `UPDATE` per source on refresh, 8KB-812KB depending on source size. Direct 1:1 translation of the current Redis structure.

**JSON column on `EPGData`**: store each channel's offsets (e.g. `[14208, 5765120]`) on its own row. More normalized, but means bulk-updating up to 58K rows per source on refresh. There's also an ordering issue since the index build scans the raw XML before the parse step that creates `EPGData` rows, so the pipeline would need reordering or a two-pass approach.

I'd lean `EPGSource` if we go the DB route, but I'm also fine keeping it in Redis if you'd rather skip the migration. The boot rebuild is cheap (under 20s even for 11GB) and the logic is already there. No preference either way, let me know.

FiveBoroughs:
Also, moving to the DB sidesteps a TTL problem. The Redis keys have no TTL because the EPG refresh interval is user-configurable, so there's no safe expiry to pick. With the DB it's just overwritten on refresh.

SergeantPanda:
I am still a bit on the fence about this but am interested. I think the EPGSource would probably be the way to go as well we'd just want to make sure it isn't returned with api calls as it would cause unnecessary slowness.

I assume this just searches for the first program occurrence for the given tvg-id right?

So, if one is built in a random order it may not be very beneficial, but that isn't very common.

FiveBoroughs:
Moved the index from Redis to `EPGSource.programme_index` (JSONField), overwritten on each EPG refresh. The serializer uses an explicit `fields` list so `programme_index` is excluded from API responses.

On the ordering question: the index records every non-contiguous programme block start for a tvg-id, not just the first occurrence. If `Channel4.uk` has programmes at byte positions 100K, 500K, and 1.8M, all three are stored and checked at lookup time.

For files where programmes are fully interleaved (time-sorted rather than channel-grouped), a single channel could have hundreds of separate blocks. To prevent index explosion the scanner caps block-starts per channel at 10 (`_OFFSET_CAP`, easy to adjust); once any channel exceeds that, the source is flagged as `interleaved`. The lookup then checks all 10 stored offsets (one seek per offset), and if the current programme is past those, does a forward scan from the last known position rather than re-reading from the start.

The XMLTV spec doesn't mandate any ordering ([xmltv.dtd](https://github.com/XMLTV/xmltv/blob/master/xmltv.dtd): "It doesn't explicitly group programmes by day or by channel"), but every source I tested is channel-grouped. Checked 10 providers across 8MB to 11GB including epg.pw, epgshare01, mjh.nz, tdtchannels, iptvx.one, iptv-epg.org, and m3u4u. Max observed was 3 non-contiguous blocks per channel. The interleaved path is a safety net.

Worst-case index sizes in the DB (all observed sources are channel-grouped, so actual sizes are the left column):

| Source channels | Channel-grouped (typical) | Interleaved worst case (cap=10) |
|---|---|---|
| 260 | 8 KB | 26 KB |
| 11,706 | 360 KB | ~1.5 MB |
| 23,034 | 818 KB | ~3.0 MB |
| 32,161 | 736 KB | ~4.6 MB |

Schedules Direct sources are excluded from index builds and lookups since they have no XML file.

FiveBoroughs:
On remaining Redis usage: the index itself is fully in the DB now, but Redis is still used for two `SET NX` locks (300s TTL) to prevent duplicate index builds. That's the only Redis footprint from this feature.

On the `flushdb()` removal: #1035 independently includes the same one-line fix as part of a broader stream ownership bugfix. If that lands first, this commit becomes a no-op on rebase. I can drop it from this branch if you'd rather take it through #1035.

FiveBoroughs:
To test the time-sorted/interleaved logic I re-ordered an epgshare01 file by start time: https://gist.github.com/FiveBoroughs/6d8171f18de28cc7947331ca8ba4384f

```bash
curl -sL https://epgshare01.online/epgshare01/epg_ripper_ALL_SOURCES1.xml.gz | gunzip > /tmp/fresh.xml
python reorder_xmltv_by_time.py /tmp/fresh.xml /tmp/timesorted.xml
python test_interleaved_index.py <source_id> /tmp/timesorted.xml
```

```
Building byte-offset index...
  22960 channels indexed in 5.0s, interleaved=True

Lookups:
  20.it: Walker (795.0ms)
  24.tr: Bu Ülke (243.0ms)
  3+.ch: Border Patrol Südamerika - Chile (763.2ms)
  3+.de: Border Patrol Südamerika – Chile (84.3ms)
  33.cz: None (1847.0ms)
```

Slower than channel-grouped lookups (sub-second vs single-digit ms), but couldn't find any real-world examples of this and even if slow it's handled.

FiveBoroughs:
Rebased on upstream dev.

FiveBoroughs:
Added tests for this feature on both backend and frontend, plus a small refactor in `Channel.jsx` to extract EPG preview fetch/retry into `useEpgPreview(epgDataId)`. Runtime behavior is unchanged; `ChannelForm` still gets the same `currentProgram` / `isLoadingProgram` / `hasFetchedProgram` states, and tests now exercise the real implementation directly.

CI note: the current failures look unrelated (`docker/login-action@v2`: `Username and password required`).

FiveBoroughs:
Rebased on current main (v0.22.0). No conflicts with the `flushdb()` removal. #1035 already landed that, so that commit is a no-op here. Tests pass. The `select_related("epg")` added upstream in the guide performance work was kept over the original version.

Also checked whether this could be a plugin instead of a core change. The plugin system can only render UI within its own card on the Plugins page, with no injection point for modals or form dialogs. A plugin can't add a preview to the channel edit form. It would have to be a completely separate workflow ("open plugin, search EPG channel, see preview") which defeats the purpose. Core change is the only viable path.

SergeantPanda:
Thank you for keeping this up to date. I will look at this hopefully soon.

SergeantPanda (changes requested):
Ok, sorry for the delay on actually playing with this. I made a couple of small changes that shouldn't cause any issues. My main focus in on ensuring performance isn't going to be impacted for the 99% of use-cases that the user does not need to see this information (hopefully that makes sense).

With that in mind, the only thing I can think of that may help in some cases (epgs with interleaved channels/programmes) would be to store which channels are interleaved (if any) in case it's only one or two out of potentially thousands of channels.

We need to keep in mind some users use EPG's that are 1GB + in size and could have 10k channels in them. Some users do not have SSD's either and use SMB shares so maybe a delayed celery task would be a better approach.

The only other thing I can think of for this is our channel edit form is becoming a bit convoluted and should be reworked if this ends up getting merged. I can certainly play with it if/when that happens though but currently it pushed the form down and creates odd spacing issues. I think the form in general could use a rework though with some fields being moved to be near each other that would make more sense for the flow.

If you could also resolve the conflicts, I'd appreciate it!

SergeantPanda:
Also, if you could please accept the contributor license agreement, I'd appreciate it, this was added after you submitted.

SergeantPanda:
Just checking to see if you are still interested in getting this one merged or not.

SergeantPanda:
Is this ready for another review? If so, please fill out the PR form.

FiveBoroughs:
Yes, ready for another review. PR description updated to match the template.

Both performance asks addressed: per-channel `interleaved_channels` (fast path preserved for well-behaved channels) and immediate async build on first preview against an unindexed source (no 10s sync scan). Details in the PR description.

Rebased on dev (post-v0.25.0). I only work on this on weekends, so if conflicts surface before my next pass, ping me.

SergeantPanda (changes requested):
Thanks for the updates - the per-channel interleaved_channels tracking and the deferred programme_index in the source list response both look good. A few remaining items:

1. Index build is still synchronous inside refresh_epg_data (perf, blocking)

tasks.py line ~322:

```
# Build byte-offset index for preview lookups (runs before full program parse)
build_programme_index(source.id)
```
This stretches every EPG refresh by 0.6s-18s (your own benchmarks, potentially longer on SMB) even when nobody ever opens the channel form. The deferred dispatch path is already built - this should be build_programme_index_task.delay(source.id). The redis lock prevents duplicates, and the boot rebuild in tasks.py covers the container-restart case.

2. select_related('epg_source') fetches programme_index into every row of the EPGData batch query

api_views.py line ~780:

`epg_data_entries = EPGData.objects.select_related('epg_source').filter(id__in=epg_data_ids)`
The JOIN pulls the full EPGSource row including programme_index (up to ~3MB) once per result row. For 50 entries from one source that's the same blob 50 times across the wire. find_current_program_for_tvg_id then immediately calls source.refresh_from_db(fields=['programme_index']) and discards what the JOIN fetched.

Fix: add .defer('epg_source__programme_index') to that query - Django 6 supports it with select_related. The refresh_from_db already handles the targeted read.

3. ProgramProgress and formatProgramTime are dead code in StreamConnectionCard.jsx

ProgramProgress is defined at line 68 but never called - ProgramPreview does the job now. formatProgramTime is only used inside ProgramProgress. Both can be deleted.

4. Build lock is never explicitly released

build_programme_index_task relies entirely on the 300s TTL to clear building_programme_index_{source_id}. If the build finishes in 5 seconds the lock still holds for 295 more, blocking any retry if the result is unexpectedly missing. A try/finally that deletes the key after success or failure keeps the window tight.

5. Retry deadline is dead code in useEpgPreview.jsx

deadlineMs = 3 * 60 * 1000 (180s) is checked each iteration but with maxRetries = 5 the loop exhausts in ~39 seconds (3s + 4.5s + 6.75s + 10.1s + 15s). The deadline check never fires. Either bump maxRetries to cover slow SMB builds or remove the dead constant.

#1 and #2 are the direct follow-ons to the performance feedback from previous review rounds. #3 is a one-deletion cleanup. #4 and #5 are minor but straightforward.

FiveBoroughs:
All five done in `10d558f`. Two places I made a call worth flagging.

On the retry deadline, I bumped `maxRetries` to 20 rather than dropping the constant, so the 180s deadline is the real cutoff (backoff caps at 15s) and slow SMB builds get the full window.

On the lock, I moved the `SET NX`/`delete` into `build_programme_index_task` itself with the release in a `finally`, so it also dedups the refresh-triggered and preview-triggered builds, and the no-index lookup branch dispatches without its own lock.

Two related things I found while in there.

The first is a channel-id mismatch. The index keyed `programme@channel` from raw bytes, but `EPGData.tvg_id` comes from lxml-parsed, stripped `channel@id`. So a channel like `A&amp;E.us` or one with surrounding whitespace built a key that never matched its `tvg_id`, and the preview silently came up empty. Added a shared `_decode_channel_id` (`html.unescape` guarded by `&`, plus strip) at the build and both lookup sites. No new dependency, `html` is stdlib.

The second: `getCurrentProgramForEpg` caught `request()` errors and returned `null`, which is indistinguishable from "no current program", so a transient backend error killed the retry loop. It propagates now so `useEpgPreview` retries. That error-retry path was dead before this.

27 backend tests plus the frontend preview suites pass. Rebased on latest dev.


---

edit: a few follow-up fixes from more edge-case tests.

The startup rebuild was taking the build lock and then dispatching the task, but since the lock now lives inside the task, the task saw the held lock and returned without building, so a source with no index never got one on boot. The rebuild dispatches without its own lock now.

The channel-attribute regex only matched `channel="x"`, not `channel = "x"`, which is valid XML. Loosened it to allow whitespace around the `=`, which covers the build and both lookup helpers since they share the regex.

Named HTML entities in programme text like `Caf&eacute;` made the element parse throw, so the programme was skipped and the preview came up empty. The element parse now prepends the existing HTML-entity DOCTYPE and resolves entities, matching how the full-file parser already handles it.

Also added coverage modeled on real feeds: epgshare01 FR with apostrophe and predefined entities, the all-sources channel/start-stop attribute orderings, jesmann numeric ids, single-quoted attrs, and a programme larger than the read chunk.



Ready for another review when you get a chance.

SergeantPanda (approved):
Ok, I went through everything and made a couple of small changes. I also refactored the channel edit form because it was starting to get a bit convoluted.

Overall really good job with this. Thank you for sticking to it and optimizing it to the point where I find it viable.

Good to merge.

FiveBoroughs:
Awesome, thanks. I really like the three-column form, reads way better than what I had. Migration went smooth on my end too.

SergeantPanda:
I've been wanting to reorganize it for a while now, this was a good push to do so

Issue #800, opened 2026-01-02
[Bug]: forward slash duplicates on xmltv.php link (or M3U link doesn't strip it)


ChilledSlim:
### Describe the bug

After adding an M3U link http://myhost.com/ and choosing type 'Xtreme' and entering in the password, as well as checking to create the matching EPG, the EPG is created as http://myhost.com//xmltv.php.
That forward slash is duplicated.

### How can we recreate this bug?

1. Add a http://myhost.com/ as type Xtreme Codes
2. Have it create the EPG automatically
3. The EPG is created with //xmltv.php

If I enter the link as http://myhost.com (without the trailing forward slash) it creates the EPG link correctly.
So either the M3U link should have a trailing forward slash removed if it adds /get.php and /xmltv.php or it's just creating the get.php and xmltv.php links by adding an unnecessary forward slash.

### Dispatcharr Version

0.15.1

### Docker Version

Docker version 29.1.3, build f52814d

### What's in your Docker Compose file?

services:
  dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:latest
    container_name: dispatcharr
    volumes:
      - /srv/data/dispatcharr:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - CELERY_BROKER_URL=redis://localhost:6379/0
      - DISPATCHARR_LOG_LEVEL=info
    devices:
      - /dev/dri:/dev/dri  # For Intel/AMD GPU acceleration (VA-API)
    network_mode: service:nordlynx
    depends_on:
      nordlynx:
        restart: true
        condition: service_healthy

### Client Information

N/A

CodeBormen:
Hello, I have implemented a fix to address how XMLTV paths are created from XC URLs. I initially built the fix to handle XC URLs that include a subpath, but because of how the URL is now built it also fixes this issue. #946 hoping to get it merged to dev soon!

SergeantPanda:
Available in dev.

PR #946, opened 2026-02-06
Fix XC URL sub-path handling in profile refresh and EPG creation

### LINKED ISSUE #945: [Bug]: XC profile refresh extract wrong credentials from XC URLs with sub-paths
### Describe the bug

When an Xtream Codes account has a server URL that includes a sub-path (e.g., http://server.com/portal/a/), the background profile refresh task extracts the wrong username and password from the URL. This causes authentication failures against the provider's API, resulting in error notifications and failed profile updates.

The cause is from get_transformed_credentials() using hardcoded array indices [1] and [2] to extract the username and password from the URL path, which only works when the server URL has no sub-path.

Specifically, from tasks.py:

- complete_url = f"{clean_server_url}/live/{base_username}/{base_password}/1234.ts"
- path_parts = [part for part in parsed_url.path.split('/') if part]
- transformed_username = path_parts[1]
- transformed_password = path_parts[2]

When no sub-path is in the URL, here is the effective build:

http://server.com/live/myuser/mypass/1234.ts
path_parts = ['live', 'myuser', 'mypass', '1234.ts']
path_parts[1] = 'myuser'   
path_parts[2] = 'mypass'

When a sub-path is present in the URL:

http://server.com/portal/a/live/myuser/mypass/1234.ts
path_parts = ['portal', 'a', 'live', 'myuser', 'mypass', '1234.ts']
path_parts[1] = 'a'
path_parts[2] = 'live'

1. Add an Xtream Codes account with a server URL that includes a sub-path (e.g., http://server.com/portal/a/)
2. Trigger a refresh of the account
3. Observe the profile refresh fails with a 404 or authentication error, using wrong credentials extracted from the server sub-path

### Dispatcharr Version

v0.18.1 (dev branch)

### Docker Version

v29.2.0

Stock AIO compose file

### Client Information

Any client, issue is app-side

CodeBormen:
- Fix credential extraction in get_transformed_credentials() using negative indices anchored to the known tail structure instead of hardcoded indices that break when server URLs contain sub-paths
- Fix EPG URL construction in M3U form to normalize server URL to origin before appending xmltv.php endpoint

XC accounts with sub-paths in their server URL (e.g., http://server.com/portal/a/) caused two failures: profile refresh extracted wrong credentials from the URL path due to hardcoded array indices, and EPG auto-creation appended xmltv.php under the sub-path instead of at the domain root. Both fixes normalize away the sub-path using the same approach the backend's XCClient already uses for API calls.

Fixes #945


CodeBormen:
Also fixes #800

SergeantPanda:
This looks good to me. Thank you!

PR #949, opened 2026-02-06
Fix/(ts-proxy): Harden connection slot lifecycle — leak prevention, TOCTOU race, and release safety

### LINKED ISSUE #947: [Bug]: Connection capacity leak - failed stream initialization permanently consumes connection slot
### Describe the bug

When a channel stream fails during initialization (before a stream generator is created), the profile_connections Redis counter is incremented by Channel.get_stream but never decremented because no release_stream is called on the failure paths. This permanently consumes a connection slot for the M3U account profile, blocking all future stream attempts until the container is restarted. For M3U profiles with a max connection size of 1, the provider becomes inaccessible until container restart.

generate_stream_url calls channel.get_stream which increments the Redis counter but if the function fails (DB lookup error, URL transform exception) it returns None without calling channel.release_stream. The same issue occurs in stream_ts when ChannelService.initialize_channel returns False.

1. Add an XC or M3U account with max_connections set to 1
2. Assign streams from this account to multiple channels
3. Attempt to play a channel where the provider stream is unavailable (provider down, stream offline, etc.)
  3a. The stream initialization fails — observe the 500 or 503 error response
4. Attempt to play any other channel using the same M3U account
  4a. Error: "No profiles available with connection capacity for M3U account {id}" or "All active M3U profiles have reached maximum connection limits"
5. All subsequent channel attempts fail with the same error
6. Restarting the container clears the stuck Redis counter and restores functionality

### Dispatcharr Version

v0.18.1 (dev branch)

### Docker Version

v29.2.0

Using the stock AIO compose file

### Client Information

Happens from any client when channels fail to initialize

CodeBormen:
Fixed four leak paths in the TS proxy streaming initialization:

1. url_utils.py - generate_stream_url(): Wrapped post-get_stream() code in try/except that calls release_stream() on any failure

2. views.py - stream_ts() retry loop: Added release_stream() before returning 503 to clean up the error-checking get_stream() call at line 181 which could INCR without a matching release when the retry loop times out

3. views.py - stream_ts() init failure: Added guarded release_stream() when initialize_channel() returns False, using a connection_allocated flag to prevent incorrect decrements when joining an existing channel

4. views.py - stream_ts() outer except: Added guarded release_stream() as a safety net for unexpected exceptions, only releasing when the flag confirms the slot was allocated and the channel hasn't been initialized yet

The connection_allocated flag prevents double-decrements by tracking whether request flow actually INCR'd the counter (fresh initialization) vs returned cached results (joining an existing channel on another worker)

Fixes #947


CodeBormen:
I am going to pull this PR back to draft. The changes I have submitted so far are good to correct stream initialization failure leaks, but after working with a user in discord I also realize now that there are some disconnection leaks that fail to decrement the counter which need to be added as well. Will re-mark for review after changes and testing.

CodeBormen:
Well, I guess to be more clear, the issue which I've identified does decrement the counter but the stream shutdown is never scheduled so effectively the stream is kept alive.

SergeantPanda:
Thank you for being cautious. The proxy is an extremely important part of Dispatcharr and any change will be highly scrutinized.

SergeantPanda:
On this one. I forgot to mention there is a PR that may interfere/overlap a bit with what you're trying to fix.

CodeBormen:
#838 ? I will compare across both and see what the differences are and see if there is a clear path forward

CodeBormen:
There definitely is some overlap. I will comment on that thread and see if it would be possible for me to bring the author's changes into this branch.

CodeBormen:
@SergeantPanda I spoke to the author of #838 and they were ok with me pulling in their changes to this branch to make a comprehensive PR. This now fixes many TOCTOU issues in addition to connection leaks and safety checks. I have tested these changes comprehensively, and am reaching out to folks who I know have these types of issues to see if they can test, too.

In an attempt to keep the changes in the current style of development, I did take the original Lua script from the other PR and rebuild it as native Redis commands so it is in better alignment. 

I know this is a fairly high-risk change (though I have tried to keep risk low), but if you don't mind looking over the changes if you have time at some point, I'd appreciate it

SergeantPanda:
I merged #1035 and am looking at getting all proxy changes in this next release if possible. How is this one coming along? 

Honestly, no rush and if it isn't ready or you don't have time it's not a big deal. I would rather err on the side of caution than rush it.

CodeBormen:
It's good, I haven't had any negative reports from user testing or from my own usage, should be ready to go! I have just been waiting to see if any of the VOD proxy changes have had any complaints, so far none. I will merge in the most recent dev branch this afternoon and do some final testing, then mark it as ready if I can't see any issues!

SergeantPanda:
Sounds perfect. Thank you!

CodeBormen:
Let's Rock n Roll!

CodeBormen:
There definitely are some things that still cause ghost connections, but I figured I'll go back and address those separately. These were the more important fixes and didn't want to muddy up the PR. 

For sure! If any issues come up, I'll be watching!

PR #972, opened 2026-02-12
Fix VOD movie/series IDs changing on every M3U refresh

### LINKED ISSUE #961: VOD movie IDs change on every M3U refresh, breaking XC-compat STRM URLs
## Bug Summary

## Impact

## Root Cause

Two issues work together:

```python
# Handle relation
    # Update existing relation
    relation.category = category
    ...
```

```
```

```
```

```python
    ...
```

## Environment

firestaerter3:
## Summary

- When a relation already exists for a `stream_id`, use the relation's existing movie/series instead of doing a fresh TMDB/IMDB/name+year lookup. This prevents relations from being repointed to newly created duplicate records, preserving stable IDs for external references (STRM files, XC-compat URLs).
- Fill in missing TMDB/IMDB IDs on existing movie/series records (only when currently NULL) so future lookups for genuinely new streams work correctly.
- Applied the same fix to both `process_movie_batch` and `process_series_batch`.

## Root Cause

`process_movie_batch` unconditionally reassigned `relation.movie = movie` after looking up movies by TMDB/IMDB/name+year key. When existing movies lacked a TMDB ID (common on initial import from XC providers), the key-based lookup failed, a new duplicate movie was created with a new auto-increment ID, and the relation was repointed — breaking all external references to the original ID.

## What Changed

In `apps/vod/tasks.py`, for both movie and series batch processing:

1. **Prioritize existing relation's movie/series** — if `stream_id` already has a relation, use `existing_relations[stream_id].movie` instead of doing a fresh key-based lookup
2. **Fill in NULL external IDs** — when updating an existing relation's movie, only set `tmdb_id`/`imdb_id` if they are currently NULL (prevents overwriting correct values)
3. **Don't reassign `relation.movie`/`relation.series`** on existing relations — only update metadata (category, custom_properties, last_seen)

## Test plan

- [ ] Import VOD from an XC provider with TMDB IDs in stream data
- [ ] Trigger M3U account refresh
- [ ] Verify movie IDs remain stable (no new duplicate `vod_movie` records created)
- [ ] Verify existing STRM/XC-compat URLs still resolve correctly
- [ ] Verify TMDB IDs are now filled in on movies that previously had NULL
- [ ] Verify genuinely new streams still create new movie records correctly

Fixes #961

### DIFF
--- a/apps/vod/tasks.py
+++ b/apps/vod/tasks.py
@@ -541,15 +541,33 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
         movie_data = data['movie_data']
         logo_url = data.get('logo_url')
 
-        if movie_key in existing_movies:
-            # Update existing movie
+        # Determine which movie to use:
+        # 1. If this stream already has a relation, preserve its movie to maintain
+        #    stable IDs for external references (STRM files, XC-compat URLs, etc.)
+        # 2. Otherwise, look up by TMDB/IMDB/name+year key
+        # 3. If no match, create a new movie
+        has_existing_relation = stream_id in existing_relations
+        if has_existing_relation:
+            movie = existing_relations[stream_id].movie
+        elif movie_key in existing_movies:
             movie = existing_movies[movie_key]
+        else:
+            movie = None
+
+        if movie is not None:
+            # Update existing movie metadata
             updated = False
 
             for field, value in movie_props.items():
                 if field == 'custom_properties':
                     if value != movie.custom_properties:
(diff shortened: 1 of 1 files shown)


firestaerter3:
Closing to validate the fix locally before submitting. Will reopen once tested.

firestaerter3:
Reopening with tests included. All 8 tests pass with the fix, and the 2 core regression tests (movie + series) fail on the original code — confirming they catch the bug.

PR #974, opened 2026-02-12
fix: add support for legacy numpy in celery entrypoint

patrickjmcd:
- adds support for the `USE_LEGACY_NUMPY` environment variable in celery's entrypoint

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -117,6 +117,11 @@ services:
       # Process Priority Configuration (Optional)
       #- CELERY_NICE_LEVEL=5  # Celery/EPG/Background tasks (default: 5, low priority; Range: -20 to 19)
 
+      # Legacy CPU Support (Optional)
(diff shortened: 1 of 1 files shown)


SergeantPanda:
I do believe this is probably necessary for users that don't want to use AIO compose. Would you mind taking the note from the Web service and duplicating it into the celery service (for the compose)?

patrickjmcd:
> I do believe this is probably necessary for users that don't want to use AIO compose. Would you mind taking the note from the Web service and duplicating it into the celery service (for the compose)?

absolutely, see updated PR!

SergeantPanda:
This looks good thank you for adding that!
I trust that you have tested this with your setup and it works as expected?

patrickjmcd:
> This looks good thank you for adding that! I trust that you have tested this with your setup and it works as expected?

correct, I have it running in my kube cluster

image

SergeantPanda:
Awesome, thank you for the contribution!

PR #977, opened 2026-02-13
feat: expose stream_stats in Xtream get_live_streams API

firestaerter3:
## Summary

- Adds the primary stream's `stream_stats` JSON field to the Xtream `get_live_streams` API response
- Uses Django `Prefetch` to efficiently load each channel's primary stream (ordered by `channelstream.order`) in a single query, avoiding N+1 performance issues
- Returns `null` for channels with no stream assigned or no stats populated

## Motivation

Xtream clients (e.g. Jellyfin's native HDHR tuner) currently hardcode stream metadata as 1080p because there's no way to know the actual resolution. When the real stream is 720p, this causes unnecessary full video transcoding.

With `stream_stats` exposed (populated by external tools like StreamFlow that probe channels and write stats to the DB), clients can declare the correct resolution, codec, and bitrate — eliminating wasteful transcoding.

## Response format

Channels with stats:
```json
{
  "stream_stats": {
    "resolution": "1920x1080",
    "video_codec": "h264",
    "audio_codec": "aac",
    "source_fps": "25",
    "ffmpeg_output_bitrate": "4500k"
  }
}
```

Channels without stats:
```json
{
  "stream_stats": null
}
```

## Test plan

- [ ] Query `get_live_streams` endpoint and verify `stream_stats` appears in response
- [ ] Verify channels with populated `stream_stats` in DB return the correct JSON
- [ ] Verify channels without stats return `"stream_stats": null`
- [ ] Verify no N+1 query regression (Django debug toolbar or query logging)

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -2128,7 +2128,17 @@ def xc_get_live_categories(user):
 
 
 def xc_get_live_streams(request, user, category_id=None):
+    from django.db.models import Prefetch
+    from apps.channels.models import Stream
+
     streams = []
+
+    # Prefetch primary stream (lowest order) with stats for each channel
+    stream_prefetch = Prefetch(
+        'streams',
+        queryset=Stream.objects.order_by('channelstream__order'),
+        to_attr='ordered_streams'
+    )
 
     if user.user_level < 10:
         user_profile_count = user.channel_profiles.count()
@@ -2142,7 +2152,7 @@ def xc_get_live_streams(request, user, category_id=None):
             # Hide adult content if user preference is set
             if (user.custom_properties or {}).get('hide_adult_content', False):
                 filters["is_adult"] = False
-            channels = Channel.objects.filter(**filters).order_by("channel_number")
+            channels = Channel.objects.filter(**filters).prefetch_related(stream_prefetch).order_by("channel_number")
         else:
             # User has specific limited profiles assigned
             filters = {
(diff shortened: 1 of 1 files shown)


SergeantPanda:
The issue I see with this is that is not a proper return for XC so clients wouldn't read it. In order for this to work, clients would all need to be updated to support the changes to the get_live_streams endpoint.

firestaerter3:
Hi... agree its outside XC standard, but could be added as 'additional data'. Only for clients making a special 'Dispatcharr mode'; like I'm making for Jellyfin: https://github.com/firestaerter3/Jellyfin-Xtream-Library

SergeantPanda:
If a tool is designed specifically for Dispatcharr could it not just pull the proper api that already outputs the stats `/api/channels/streams/`

SergeantPanda:
If it isn't feasible due to authentication I understand that. 
One issue you can/will run into is a channel can have multiple streams. So if you make the assumption that the codecs are going to be whatever the first stream has and it fails to a different stream with different codecs how would that be handled?

firestaerter3:
You're right — I've updated my plugin to use the proper REST API (`/api/channels/streams/` + `/api/channels/channels/`) with JWT authentication. Works great, no changes needed on the Dispatcharr side.

On the multi-stream point: currently the plugin uses the primary stream's stats since that's what Dispatcharr serves under normal conditions. If it fails over to a backup with different codecs, the metadata would be off until the next tuner refresh. It's a decent tradeoff but not perfect.

Do you have any ideas for handling that more gracefully? For example, does Dispatcharr expose which stream is currently active for a channel, or could there be a webhook/event when a failover happens?

Closing this PR since the Xtream API approach isn't needed anymore.

SergeantPanda:
Awesome, glad you got it. You could query stream stats api endpoint to see which actual stream is currently being used. Also, we just added webhooks with events in our dev build so you could use that for a when a stream fails over.

Otherwise I am not really sure how you could handle it more gracefully because you wouldn't know it failed over until after Dispatcharr makes the connection. What you have is probably going to be fine in general since most streams use H264, maybe you could make a note saying it's recommended to trans code the audio/video to a consistent codec to avoid potential issues during failover (which is preferred for Emby,Plex,JF anyway)

Issue #971, opened 2026-02-12
[Bug]: VOD Proxy: TOCTOU race condition in connection limit enforcement allows max_streams to be exceeded


CodeBormen:
### Describe the bug

The VOD proxy's connection limit enforcement has a TOCTOU (Time-of-Check-to-Time-of-Use) race condition identical to the one fixed in the TS proxy (PR #947). _check_profile_limits() performs a Redis GET to read the current connection count, checks it against max_streams, and returns a boolean. The actual INCR happens later in a separate _increment_profile_connections() call, ~50 lines apart. Between the check and the increment, concurrent requests from other workers can pass the same check, exceeding max_streams.

### How can we recreate this bug?

1. Set an M3U profile's max_streams to 2
2. Send 4+ simultaneous VOD requests that use the same profile
3. More than 2 connections are established simultaneously

### Dispatcharr Version

v0.19.0

### Docker Version

v29.2.0

### What's in your Docker Compose file?

Standard AIO cocnfig file but occurs in all versions of deployment

### Client Information

Any client, issue is server-side

CodeBormen:
This bug was discovered as part of the same fixes I am working on for PR #949. 

Even though this is the same class of bugs I am working on fixing, because it is in a different subsystem than TS proxy, I am going to work on this separate from those fixes. 

I will apply the same INCR-first-then-check pattern used in the TS proxy in the above PR.

CodeBormen:
Fix submit in #983

SergeantPanda:
Available in dev.

SergeantPanda:
We'll keep it open until it's merged

SergeantPanda:
This should have closed automatically but did not.

Issue #533, opened 2025-10-07
[Bug]: [VOD] Multi-Account issue with same source server


gardor007:
### Describe the bug

Describe the bug
When streaming multiple VODs from an Emby Server through Dispatcharr, the second stream fails, even though multiple accounts are configured.

Expected behavior
Each configured account should be able to stream VOD content simultaneously from Emby without failure.

Actual behavior
With one VOD stream active in Emby, playback works correctly.
Starting a second VOD stream fails immediately.
Live TV streams are unaffected and function correctly with multiple simultaneous connections.

Additional context
All three provider accounts are configured identically in Dispatcharr.
This issue appears specific to VOD playback initiated from Emby; Live TV streaming works as expected.

### How can we recreate this bug?

Steps to reproduce

Configure multiple Dispatcharr accounts (e.g., 3) from the same VOD provider.
Start a VOD stream from Emby
Start a second VOD stream → fails.

2025-10-07 23:44:14,551 INFO apps.proxy.vod_proxy.views [VOD-SEEK] Seeking to byte position: 0 (~0.0 MB)
2025-10-07 23:44:14,551 INFO apps.proxy.vod_proxy.views [VOD-SEEK] Rapid request detected (1.5s) - likely seeking
2025-10-07 23:44:14,551 INFO apps.proxy.vod_proxy.views [VOD-CLIENT] Client info - IP: 192.168.222.247, User-Agent: Lavf/59.27.100...
2025-10-07 23:44:14,551 INFO apps.proxy.vod_proxy.views [VOD-SESSION] Creating new session: vod_1759880654551_2352
2025-10-07 23:44:14,551 INFO apps.proxy.vod_proxy.views [VOD-SESSION] Redirecting to path-based URL: /proxy/vod/movie/12667739-e446-478c-b3a9-38475da7fec4/vod_1759880654551_2352
2025-10-07 23:44:14,552 INFO apps.proxy.vod_proxy.views [VOD-REQUEST] Starting VOD stream request: movie/12667739-e446-478c-b3a9-38475da7fec4, session: vod_1759880654551_2352, profile: None
2025-10-07 23:44:14,553 INFO apps.proxy.vod_proxy.views [VOD-REQUEST] Full request path: /proxy/vod/movie/12667739-e446-478c-b3a9-38475da7fec4/vod_1759880654551_2352
2025-10-07 23:44:14,553 INFO apps.proxy.vod_proxy.views [VOD-REQUEST] Request method: GET
[...]

PR #983, opened 2026-02-14
Fix: VOD proxy connection counter leak on client disconnect

### LINKED ISSUE #962: VOD proxy leaks Redis connection counter on abrupt client disconnect (ffprobe)
## Bug Summary

The VOD proxy increments the Redis `profile_connections:<id>` counter when a client connects, but never decrements it when the client disconnects abruptly (e.g., `ffprobe`, `curl -r 0-1023`, or any partial download). This causes the profile to permanently appear "at capacity", rejecting all subsequent VOD requests with 503.

## Impact

With `max_streams=1`, a single ffprobe request permanently locks out the VOD profile. All subsequent requests fail:

```
[PROFILE-SELECTION] All profiles at capacity for M3U account 9, rejecting request
[VOD-ERROR] No suitable M3U profile found for movie <uuid>
Service Unavailable: /proxy/vod/movie/<uuid>/<session_id>
```

The only recovery is manually clearing the Redis key:
```bash
redis-cli DEL profile_connections:7
```

## Steps to Reproduce

1. Set `max_streams=1` on an M3U account profile (via DB or UI)
2. Make a single VOD request that disconnects before the stream completes:
   ```bash
   curl -s -o /dev/null -r 0-1023 "http://<dispatcharr>:5656/movie/<user>/<pass>/<id>.mkv"
   ```
3. Check Redis: `redis-cli GET profile_connections:7` → returns `1` (leaked)
4. All subsequent VOD requests return 503

## Root Cause

The VOD proxy in `apps/proxy/vod_proxy/views.py` increments the connection counter when a stream starts but relies on the stream completing normally to decrement it. When a client disconnects mid-stream (which is normal for tools like ffprobe that only read headers/first few bytes), the decrement never happens.

This is particularly problematic for:
- **ffprobe** — reads only stream headers to extract codec/resolution/duration, then disconnects
- **Partial range requests** — `Range: bytes=0-1023` completes the requested range but Dispatcharr may not treat it as "complete"
- **Jellyfin media probing** — uses ffprobe internally to identify media streams

## Evidence

```bash
# Before any requests
redis-cli GET profile_connections:7
→ (nil)

# After one ffprobe (which disconnects after reading headers)
redis-cli GET profile_connections:7
→ "1"

# No active streams, but counter is stuck
# All subsequent requests: 503 "All profiles at capacity"
```

## Workaround

Set `max_streams=0` (unlimited) on the M3U **account** (not just the profile — the profile value gets overwritten from the account on restart via a Django signal in `m3u/models.py`):

```sql
UPDATE m3u_m3uaccount SET max_streams = 0 WHERE id = <account_id>;
UPDATE m3u_m3uaccountprofile SET max_streams = 0 WHERE m3u_account_id = <account_id>;
```

This bypasses the connection counter check entirely.

## Suggested Fix

- Decrement the connection counter in a `finally` block or disconnect handler, not just on stream completion
- Add a TTL to the Redis counter keys so leaked counters auto-expire (e.g., after the stream timeout period)
- Consider using Redis `EXPIRE` on connection keys so they self-heal

## Related

This bug was also documented in the Jellyfin-Xtream-Library project: the "Dispatcharr VOD connection leak" is a known issue when ffprobe is used against Dispatcharr-proxied streams.

## Environment

- Dispatcharr (latest as of Feb 2026)
- Redis localhost:6379
- M3U account type: Xtream Codes
- Client: Jellyfin 10.11.0 using ffprobe for media probing

CodeBormen:
Three fixes:
- TOCTOU: Replace GET-check-INCR with atomic INCR-first-then-check pattern in both connection managers to prevent concurrent requests exceeding max_streams
- Immediate DECR: Decrement profile counter directly in stream generator exit paths (normal completion, client disconnect, error) instead of deferring to daemon thread cleanup which may never execute
- Rollback: Decrement profile counter on create_connection() failure so the reserved slot is released

Fixes #962, #971, #451, #533

### DIFF
--- a/apps/proxy/vod_proxy/multi_worker_connection_manager.py
+++ b/apps/proxy/vod_proxy/multi_worker_connection_manager.py
@@ -416,8 +416,22 @@ class RedisBackedVODConnection:
 
                 logger.info(f"[{self.session_id}] Updated connection state: length={state.content_length}, type={state.content_type}")
 
-            # Save updated state
-            self._save_connection_state(state)
+            # Save updated state under lock to avoid overwriting concurrent
+            # active_streams changes (e.g., another stream's GeneratorExit decrement)
+            if self._acquire_lock():
+                try:
+                    current = self._get_connection_state()
+                    if current:
+                        # Preserve the current active_streams value — it may have been
+                        # modified by concurrent increment/decrement operations while
+                        # waiting for the upstream HTTP response.
+                        state.active_streams = current.active_streams
+                    self._save_connection_state(state)
+                finally:
+                    self._release_lock()
+            else:
+                # Fallback: save without lock but skip active_streams to avoid overwrite
+                logger.warning(f"[{self.session_id}] Could not acquire lock for get_stream state save")
 
             self.local_response = response
             return response
@@ -466,35 +480,44 @@ class RedisBackedVODConnection:
             return range_header
 
     def increment_active_streams(self):
-        """Increment active streams count in Redis"""
+        """Increment active streams count in Redis. Returns new active_streams count, or 0 on failure."""
         if not self._acquire_lock():
-            return False
+            logger.warning(f"[{self.session_id}] INCR-AS failed: could not acquire lock")
+            return 0
 
         try:
             state = self._get_connection_state()
             if state:
+                old = state.active_streams
                 state.active_streams += 1
                 state.last_activity = time.time()
                 self._save_connection_state(state)
-                logger.debug(f"[{self.session_id}] Active streams incremented to {state.active_streams}")
-                return True
-            return False
+                logger.debug(f"[{self.session_id}] INCR-AS {old} -> {state.active_streams}")
+                return state.active_streams
+            logger.warning(f"[{self.session_id}] INCR-AS failed: no state")
+            return 0
         finally:
             self._release_lock()
 
     def decrement_active_streams(self):
         """Decrement active streams count in Redis"""
         if not self._acquire_lock():
+            logger.warning(f"[{self.session_id}] DECR-AS failed: could not acquire lock")
             return False
 
         try:
             state = self._get_connection_state()
             if state and state.active_streams > 0:
+                old = state.active_streams
                 state.active_streams -= 1
                 state.last_activity = time.time()
                 self._save_connection_state(state)
-                logger.debug(f"[{self.session_id}] Active streams decremented to {state.active_streams}")
+                logger.debug(f"[{self.session_id}] DECR-AS {old} -> {state.active_streams}")
                 return True
+            if not state:
+                logger.warning(f"[{self.session_id}] DECR-AS failed: no state")
+            else:
+                logger.warning(f"[{self.session_id}] DECR-AS failed: active_streams already {state.active_streams}")
             return False
         finally:
             self._release_lock()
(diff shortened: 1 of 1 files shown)


CodeBormen:
Keeping this in draft for now but is ready for testing. This is the VOD fix, complimentary to PR #949 which was for TS.

CodeBormen:
Have had success in testing with this change and at least one other person who reports this fixed their issue without any reported side affects. I will mark this PR for review first since it includes fewer changes than the TS fix PR (#949) does (although that PR does admittedly fix more issues, but it has tested well also). Let me know what you think!

CodeBormen:
Ready for review!

SergeantPanda:
Thank you for all the time you spent on this. I will have to test this thoroughly. Different clients behave dramatically differently when timeshifting.

For example, Televizo opens a new connection EVERY time anything changes in the app (going full screen, timeshifting) where better clients like VLC or Sparkle will reuse the same connection when doing those things.

This is what has made connection management a pain. So, I'll go through the clients I currently have access to to make sure this still works as expected and no additional connections are made.

Thanks again, sorry for rambling!

CodeBormen:
No worries, I totally understand the need to be cautious. There are only so many environments we can test in and what might work for one client may not for another. If there are clients which you would like me to work on testing as well, I'm happy to try some out. I primarily have been testing using VLC and Jellyfin. 

I have also seen good results so far for the TS changes but again, same problem with limited environment.

tcb678:
> For example, Televizo opens a new connection EVERY time anything changes in the app (going full screen, timeshifting) where better clients like VLC or Sparkle will reuse the same connection when doing those things.
> 
> This is what has made connection management a pain. So, I'll go through the clients I currently have access to to make sure this still works as expected and no additional connections are made.

@SergeantPanda 

Does this imply that it would be useful to consider a framework for per-user-agent behavior monitoring and management, like a request token bucket or something?

SergeantPanda:
> No worries, I totally understand the need to be cautious. There are only so many environments we can test in and what might work for one client may not for another. If there are clients which you would like me to work on testing as well, I'm happy to try some out. I primarily have been testing using VLC and Jellyfin.
> 
> I have also seen good results so far for the TS changes but again, same problem with limited environment.

Perfect, so my ask would be to not only check connection increments and decrements in redis but also watch what your provider shows for connections. If you only have 1 VOD connected it should only show 1 connection from your provider profile (if you aren't sure how to check, let me know). Once you only see 1 listed, try timeshifting and make sure it does not increase the connection count to your provider. That is the biggest hurdle I was able to overcome when building this out.

This actually makes Dispatcharr better than some clients going directly to the provider.

Of course, you will only notice this if you have a provider that allows for more than 1 connection.

Hopefully that makes sense.

SergeantPanda:
> > For example, Televizo opens a new connection EVERY time anything changes in the app (going full screen, timeshifting) where better clients like VLC or Sparkle will reuse the same connection when doing those things.
> > This is what has made connection management a pain. So, I'll go through the clients I currently have access to to make sure this still works as expected and no additional connections are made.
> 
> @SergeantPanda
> 
> Does this imply that it would be useful to consider a framework for per-user-agent behavior monitoring and management, like a request token bucket or something?

No, we don't need to handle different clients different, we just detect the behavior and change the workflow based on what the client does.

CodeBormen:
Hmm, you're right, there is a little funk when it comes to provider connections. I'm getting closer but timeshifting or rapid scrubbing really does mess around with the re-use function. Gonna keep working on this and see if I can get those worked out.

CodeBormen:
Alright, I'm feeling pretty confident about this now. After having tested this for a few hours now between multiple clients and simulated testing, I have gotten it to be very stable. With the most recent changes, I am not able to desync counters whatsoever, and more importantly, I cannot find any scenarios that exceed the maximum connection limit even when rapidly switching VODs or timeshifting. The biggest fix for that issue was a save state within active_streams to prevent concurrent decrements when rapidly switching or timeshifting. 

I did also simulate the use-case scenario you mentioned with Televizo. While I don't have access to the client to natively test, I was able to throw rapid connects/disconnects at the same VOD connection to simulate, which tested well.

If you have any other ideas for testing, feel free to let me know!

SergeantPanda:
Ok I have been testing this morning and I am noticing that it's too aggressive at cleaning up the old connection. Which results in multiple connections to the provider when switching to full screen mode in Televizo.

If you have an android device, Televizo is a free app and it the worst one I've come across for causing such behavior (so it's a good testing app)

SergeantPanda:
This issue may have already existed, doing more testing on current dev branch. Might just need to adjust a timer a bit.

CodeBormen:
For sure, I don't have an Android device but I can probably obtain one for testing; since I couldn't reproduce that behavior in my environment it would be helpful to have access to the worst possible usage scenario for testing purposes.

Sounds like you may have already identified the issue but if you'd like me to make any additional changes or do more testing, I am happy to! And I'll look into getting an Android device for testing so that I can put changes through more scrutiny, as well.

SergeantPanda:
I think this is OK to merge since there isn't a regression in behavior. We can fix the bug that occurs with clients like Televizo with a different PR. 

Thank you for your work on this. I'll wait for your go-ahead to actually merge in case you have code in the works.

CodeBormen:
Nothing in the works for this one, I've been regression testing and it seems to be solid. If a bug does come up, I'll get it addressed ASAP. The TS change is a little larger so I'll wait until this moves to production and has a lot of user testing before asking for that one to be merged.

And for sure, just happy to be a part!

Issue #785, opened 2025-12-27
[Bug]: VOD UUID on M3U refresh


mjcocat:
### Describe the bug

I started getting 500 errors while watching VOD (series) and continually afterwards. It appears that with each M3U refresh the UUID is changing which is breaking the proxy address and causing the client to error out (500)

### How can we recreate this bug?

1. Start a stream is a series in VOD
2. refresh the M3U
3. attempt to stream the same series

### Dispatcharr Version

V15.1

### Docker Version

N/a

### What's in your Docker Compose file?

If not using Docker Compose just put not using.

### Client Information

Same issue with sparkle tv and swiptv

ullms1:
I hope this gets fixed soon? Having to recreate the strm files after a refresh is annoying.

uwjhn:
I also have the same problem.
But it's enough to "initialize" the VOD again. You don't have to recreate the STRM files.

@ullms1 A workaround is:
a) Search for the series in VOD and open it (it's enough that the episodes are shown (again). No need to play an episode from here). Then the UUIDs will be generated again. --> the same STRM will work again (for this series)
b) Build a script to run through your STRMS and call /api/vod/series/{id}/provider-info/ for each series. This is how I do it with this script: https://github.com/uwjhn/generate_selected_strm_from_dispatcharr

```
# --- PROVIDER INFO TRIGGER ---

def trigger_provider_info(series_id):
    """Triggers the background retrieval of stream information for the series."""
    url = f"{API_BASE}/api/vod/series/{series_id}/provider-info/"
    print(f"Initializing streams via Provider-Info (ID: {series_id})...")
    resp = request_with_token("GET", url, timeout=60)
    if resp and resp.status_code == 200:
        print("Successfully initialized.")
        return True
    else:
        print(f"Warning: Provider-Info failed or timed out.")
        return False
```

znake-oil:
I think this is due to a small section in /app/apps/vod/tasks.py

```
# Clear existing episodes for this account to handle deletions
Episode.objects.filter(
    series=series,
    m3u_relations__m3u_account=account
).delete()
```

If you comment this out, then it does not delete when it refreshes. I think this was added as part of a db migration back in the days. I will do some testing.

SergeantPanda:
Available in dev.

Issue #985, opened 2026-02-14
[Bug]: Episode Relations wiped on XC Sync/Refresh - unstable UUID for links


moncheez:
### Describe the bug

When running a refresh/sync on an Xtream Codes (XC) source, existing episode relations are treated as stale and removed. Episodes are then recreated without preserving their existing vod_m3uepisoderelation entries. This causes all episodes to receive new internal UUIDs, breaking stability and forcing repeated XC API calls whenever a VOD series is viewed.

This behavior results in:
- Constantly changing .strm links
- Loss of episode-to-stream continuity
- Unnecessary and repeated XC queries
- Poor cache behavior and scalability issues for large libraries

**Actual Behavior**
- During an XC refresh/sync, previously discovered episode relations (vod_m3uepisoderelation) are considered stale.
- Episode relations are deleted during cleanup.
- When visiting a VOD series page afterward, Dispatcharr re-queries the XC API to rebuild episodes.
- Newly created episodes receive new UUIDs and relations.
- All episode links change, even when nothing has changed upstream.
- This repeats on every daily or manual refresh.

- The episode relation logic exists (vod_m3uepisoderelation) and already supports updating and reassigning relations.
- The issue appears to be that the episode refresh task is not invoked as part of the XC refresh pipeline, while stale cleanup is.
- This results in episode relations being deleted without being rebuilt in the same refresh cycle.

Expected Behavior
- On XC refresh/sync:
     - Existing episodes should be matched against current XC data.
     - If a vod_m3uepisoderelation already exists for a stream (stream_id), it should be retained or reassigned, not deleted.
     - Only new or changed episodes should result in new episode or relation creation.
     - Unchanged episodes should keep their existing episode records and UUIDs.
- Cleanup should only remove episode relations that are truly gone from XC, not ones that simply weren’t refreshed in the same pass.

Problem its Causing:
- Episode UUID churn breaks downstream consumers (vod2strm plugin, share links from VOD UI)
- Visiting Series pages triggers unnecessary XC API calls. Can be throttled or limited if trying to view multiple quickly or at the same time.
- Unable to refresh an entire library when there is a large XC library.
- defeats the purpose of having persistent episode relations in the database.

**Suggested Fix (High Level)**
- Ensure episode refresh (batch_refresh_series_episodes or equivalent logic) is executed as part of XC refresh.
- Match episodes by (series_id, season_number, episode_number) and/or stream_id.
[...]

Issue #820, opened 2026-01-06
[Bug]: VOD Series Internal Server Error -- 'NoneType' object has no attribute 'episode'


daiycman:
### Describe the bug

When trying to play a VOD Series the series doesnt play and the logs show a 500 internal server error

2026-01-06 17:46:48,337 ERROR django.request Internal Server Error: /series/test/test/12117.mkv
Traceback (most recent call last):
  File "/dispatcharrpy/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/dispatcharrpy/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/app/apps/output/views.py", line 2920, in xc_series_stream
    'content_id': episode_relation.episode.uuid
                  ^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'episode'
2026-01-06 17:46:48,000 INFO uwsgi.requests Worker ID: 4 GET 500 /series/test/test/12117.mkv 9ms

### How can we recreate this bug?

Go to Series on device
Try playing an episode
see if it fails

It eventually works if you keep trying (usually)

### Dispatcharr Version

v0.16.0-20260105030320

### Docker Version

Docker version 27.0.3, build 7d4bcd8

### What's in your Docker Compose file?

docker run
  -d
  --name='Dispatcharr'
  --net='br0'
  --ip='192.168.89.243'
  --pids-limit 2048
  -e TZ="America/Chicago"
  -e HOST_OS="Unraid"
  -e HOST_HOSTNAME="unraid"
  -e HOST_CONTAINERNAME="Dispatcharr"
  -e 'TCP_PORT_9191'='9191'
  -e 'DISPATCHARR_ENV'='aio'
  -e 'REDIS_HOST'='localhost'
  -e 'CELERY_BROKER_URL'='redis://localhost:6379/0'
  -e 'DISPATCHARR_LOG_LEVEL'='info'
  -l net.unraid.docker.managed=dockerman
  -l net.unraid.docker.webui='http://[IP]:[PORT:9191]'
  -l net.unraid.docker.icon='https://github.com/Dispatcharr/Dispatcharr/blob/main/frontend/src/images/logo.png?raw=true'
[...]

PR #1016, opened 2026-02-24
Remove clearing of existing episodes on refresh

znake-oil:
This PR should fix issue #785 #985 and maybe #820.

# Problem
Every VOD refresh causes all Episode records to be deleted and recreated with new UUIDs.
Any client (Jellyfin, Emby, Plex, etc.) that has cached episode stream paths gets permanent
500 errors after a refresh, requiring a full library rescan every 4 hours.
The root cause is in refresh_series_episodes (introduced when converting to relation tables):

```
# Clear existing episodes for this account to handle deletions
Episode.objects.filter(
    series=series,
    m3u_relations__m3u_account=account
).delete()
```
This delete runs before batch_process_episodes, which means every episode is destroyed
and recreated from scratch — generating a new UUID each time.

Additionally, cleanup_orphaned_vod_content correctly marks episode relations as stale
(since their last_seen is never updated during VOD refresh), and deletes all of them:

```
Cleaned up 0 stale movie relations, 0 stale series relations, 160 stale episode relations
```
Movies are unaffected because process_movie_batch updates last_seen during refresh.
Episodes have no equivalent — their relations are always deleted as stale.

batch_process_episodes already handles create vs update correctly. It pre-fetches
existing episodes keyed by (series_id, season_number, episode_number) and routes to
episodes_to_update for existing records and episodes_to_create for new ones:

```
# Pre-fetch existing episodes
existing_episodes = {}
for episode in Episode.objects.filter(series=series):
    key = (episode.series_id, episode.season_number, episode.episode_number)
    existing_episodes[key] = episode
```
The delete before this runs is what defeats the update path — the pre-fetch finds nothing,
so all episodes go to episodes_to_create with fresh UUIDs.

Removing the delete allows batch_process_episodes to work as designed: existing episodes
are updated in place (UUID preserved), new episodes are created, and relations are rebuilt
as part of normal processing.

## Fix
Remove the delete from refresh_series_episodes

## Evidence

Before the fix - after every VOD refresh:
```
Cleaned up 160 stale episode relations   # all relations wiped
Batch processed episodes: 190 new, 0 updated, ...   # all recreated with new UUIDs
```

After fix
```
Cleaned up 8 stale episode relations   # only on-demand relations (expected)
Batch processed episodes: 0 new, 15 updated, 15 new relations, 0 updated relations
Batch processed episodes: 0 new, 112 updated, 112 new relations, 0 updated relations
Batch processed episodes: 0 new, 46 updated, 46 new relations, 0 updated relations
```

Episodes are updated in place, UUIDs preserved, relations rebuilt automatically.

## Impact

Episode UUIDs are now stable across refreshes — Jellyfin/Emby library paths remain valid
Episode relations are rebuilt during normal refresh — no manual intervention needed

No behaviour change for new episodes — they are still created correctly
Orphaned episode cleanup still works via cleanup_orphaned_vod_content last_seen logic

## Testing done

 - Note episode UUIDs before refresh
 - Trigger VOD refresh
 - Confirm UUIDs unchanged after refresh
 - Confirm episodes still playable in Jellyfin without library rescan

Note: Some help with debugging and hunting down this bug was done with Claude Code.


SergeantPanda:
You definitely were on to something here. 

The root cause was what you said but unfortunately we can't just remove that, it would leave orphaned episodes and overtime we'd have a lot of database bloat.

So, I took what you did and added some more cleanup and also added in a relationship for episodes (to their corresponding series relationship) that will allow us to properly clean this up and not just dump all episodes on every refresh.

When I wrote this originally I wasn't thinking about having STRM files and similar needing static links.

Hopefully between our changes, this will resolve the issue for everyone.

Thank you for your submission!

PR #1017, opened 2026-02-24
Add on-demand episode relation refresh in VOD proxy

znake-oil:
This PR is a companion to #1016 and helps address #820.

# Problem

When an episode has no `M3UEpisodeRelation` — on a fresh setup, after a partial refresh
failure, or when a new series is added between scheduled refreshes — the VOD proxy
immediately returns a 500 error:

```
ERROR [VOD-ERROR] Content or relation not found: episode a56ecf44-...
GET 500 /proxy/vod/episode/a56ecf44-.../vod_...  2ms
```

This is particularly visible with Jellyfin, which sends 3 simultaneous FFmpeg probe
requests on first episode access. Without a relation all three fail instantly and the
episode appears unplayable.

# Fix

When `_get_content_and_relation` finds no relation for an episode, it now triggers a
synchronous `refresh_series_episodes` call for that series before returning.

# Evidence

Without fix — all 3 Jellyfin probes fail immediately:
```
ERROR [VOD-ERROR] Content or relation not found: episode a56ecf44-...
ERROR [VOD-EXCEPTION] Error streaming episode a56ecf44-...: Content not found
GET 500 /proxy/vod/episode/a56ecf44-...  2ms
```

With fix — relation created on first request, episode playable:
```
INFO [ON-DEMAND] No relation found, triggering refresh for series Breaking Bad
INFO [ON-DEMAND] Relation created successfully for Breaking Bad
INFO Batch processed episodes: 0 new, 15 updated, 15 new relations
```

# Impact

- Episodes with missing relations become playable on first access without manual intervention
- Redis lock prevents duplicate XC API calls from concurrent client probes
- Graceful fallback if Redis is unavailable — refresh still attempted without locking
- No change to normal path — only triggered when relation lookup returns None

# When this is still needed alongside #1016

PR #1016 keeps relations stable across scheduled refreshes. This PR covers remaining
edge cases where no relation exists yet:

- Fresh Dispatcharr setup before first full refresh completes
- New series added between scheduled refreshes (up to 4h gap)
- Partial refresh failure where a series errors mid-batch
- New M3U account added — existing series have no relations for it yet

# Testing done

- Confirmed 500 error without patch on episode with no relation
- Confirmed relation created and episode playable on first access with patch
- Confirmed concurrent requests wait and reuse the result rather than duplicate the XC API call

Note: Some help with debugging and hunting down this bug was done with Claude.

### DIFF
--- a/apps/proxy/vod_proxy/views.py
+++ b/apps/proxy/vod_proxy/views.py
@@ -463,6 +463,39 @@ class VODStreamView(View):
                 if relation:
                     logger.info(f"[PROVIDER-SELECTED] Using provider: {relation.m3u_account.name} (priority: {relation.m3u_account.priority})")
 
+                # On-demand refresh if no relation found (e.g. fresh setup, new series added
+                # between scheduled refreshes, or partial refresh failure)
+                if not relation:
+                    from core.utils import RedisClient
+                    from apps.vod.models import M3USeriesRelation
+                    from apps.vod.tasks import refresh_series_episodes
+                    lock_key = f"vod_episode_refresh:{content_obj.series_id}"
+                    redis_client = RedisClient.get_client()
+                    if redis_client:
+                        lock_acquired = redis_client.set(lock_key, 1, nx=True, ex=30)
+                        if lock_acquired:
+                            try:
+                                logger.info(f"[ON-DEMAND] No relation found, triggering refresh for series {content_obj.series.name}")
+                                series_relation = M3USeriesRelation.objects.filter(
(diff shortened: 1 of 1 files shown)


SergeantPanda:
I haven't forgotten about this one. 

After our changes in the last release, is this still necessary? The issue this was fixing shouldn't exist anymore. Do you see the problem still?

znake-oil:
I think this is fixed :) Thank you!

SergeantPanda:
Thank you for everything!

Issue #1025, opened 2026-02-26
[Feature]: Filter Channels that contain stale streams (i.e. not empty)


JCBird1012:
### Is your feature request related to a problem?

Currently, it's not clear if a channel contains stale streams. 

There's no option to filter channels by "contains streams that are stale".

### Describe the solution you'd like

I see this as a two-pronged solution.

1. Adding a filter for channels that contain stale streams.
2. Change the color of said channel to anything but red? Orange? Something to distinguish it from red which are completely empty channels.

<img width="819" height="121" alt="Image" src="https://github.com/user-attachments/assets/55e47a93-46be-4b4b-bc85-5c6ce4914df2" />

### Additional context

N/A

PR #1027, opened 2026-02-26
feat: filter channels that contain stale streams

JCBird1012:
closes: #1025

### DIFF
--- a/frontend/src/components/tables/table.css
+++ b/frontend/src/components/tables/table.css
@@ -140,14 +140,14 @@ html {
   background-color: color-mix(in srgb, rgb(239, 68, 68) 30%, #27272a) !important;
 }
 
-/* Rows that contain some stale streams (may be mixed with active) */
-.partially-stale-streams-row {
+/* Rows that contain at least 1 stale stream (may be mixed with active) */
+.has-stale-streams-row {
   background-color: rgba(239, 154, 68, 0.15) !important;
   /* orange-500 @ 12% */
   box-shadow: inset 0 0 0 1px rgba(239, 154, 68, 0.3);
 }
 
-.table-striped .tbody .tr.partially-stale-streams-row:hover {
+.table-striped .tbody .tr.has-stale-streams-row:hover {
   background-color: rgba(239, 154, 68, 0.3) !important;
 }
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/tables/table.css
```diff
@@ -92,7 +92,7 @@ html {
     opacity: 0;
   }
 
-  *:hover > .resizer {
+  *:hover>.resizer {
```
JCBird1012, line 95: This is a remnant of my CSS formatter. It's valid either way.

JCBird1012:
I just realized it might be better to make the "Only Empty Channels" and "Only Containing Stale Streams" radio buttons instead of checkboxes as an affordance to show that you can only really have one or the other active at any given time. 

I don't feel that strongly about it since it should become immediately obvious when playing with the filters, but I see both sides.

SergeantPanda (changes requested):
I have considered doing this since I added the is_stale parameter. 

You aren't going to believe this but I had JUST added to our CSS for the streams table to get stale streams to have hover affects and match more closely what our empty channels look like right before I saw this PR.

I used the same name you did for the partially stale channels. I think a different name for yours would be best to make it obvious that it is not completely stale (could have a mix) 

As far as the checkbox, I think that is probably OK for now. Not really sure how we could improve that.

One small thing, I'd say let's make the "Only Containing Stale Streams" shorter if we can or make the drop down a bit wider if we can't. Currently it's wrapping.

SergeantPanda:
Weird. I don't know when automatic reviews got turned on..?

Anyway, I made a couple small changes to this. Hope you don't mind.

Basically, needed a performance overhaul since using "some" will cause extra rendering to occur.

Also renamed the css for clarity on what it is.

JCBird1012:
I'm pretty sure the Copilot reviews are a setting I have turned on - whoops. They're handy, I guess?

Edit - yup. I just flipped that off.
Screenshot 2026-03-03 at 11 31 14

And no sweat about the changes - they all make sense.

side note - I'll start using `dev` as the base of my branches now. Should've done that from the start.

SergeantPanda:
No it's all good. I thought it was something that Github had turned on. If they help you I don't have a problem with it.

I figured it was that after I asked. Lol.

I was just a bit confused why it kept commenting on things.

PR #1034, opened 2026-02-27
Add Schedules Direct lineup management

nickgerrer:
## Summary

Adds the ability to search, add, and remove Schedules Direct lineups directly from the Dispatcharr UI — no need to visit schedulesdirect.org separately.

- **SD credential privacy:** The EPG table and delete confirmation dialog no longer display the SD password (`api_key`). SD sources show `SD: <username>` instead.
- **Backend lineup API:** Two new endpoints proxy to the Schedules Direct JSON API using the source's stored credentials:
  - `POST /api/epg/sources/<id>/sd-lineups/search/` — search headends by country + postal code
  - `GET/POST/DELETE /api/epg/sources/<id>/sd-lineups/` — list, add, and remove lineups on the SD account
- **Frontend lineup management:** The EPG form gains a "Manage Lineups" section (visible for saved SD sources) where users can:
  - View currently active lineups on their SD account
  - Search available lineups by country and postal code (14 countries supported)
  - Add or remove lineups with one click
- **First-run UX:** After creating a new SD source, the modal auto-reopens in edit mode so the user can immediately search and add lineups.
- **Better error guidance:** The "no lineups configured" error message now directs users to the Manage Lineups UI instead of schedulesdirect.org.

## Test plan

- [ ] Create a new Schedules Direct EPG source — modal should auto-reopen after save
- [ ] Verify the EPG table shows `SD: <username>` instead of the password
- [ ] Verify the delete dialog shows `Account: <username>` instead of the password
- [ ] Search lineups by postal code, confirm results appear in a scrollable table
- [ ] Add a lineup, confirm it appears in the "Active lineups" list
- [ ] Remove a lineup, confirm it disappears from the list
- [ ] Trigger an EPG refresh with no lineups — confirm the improved error message appears
- [ ] Verify the lineup search results don't overflow the modal (scrollable)

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -1448,6 +1448,67 @@ export default class API {
     }
   }
 
+  static async searchSDLineups(sourceId, country, postalCode) {
+    try {
+      const response = await request(
+        `${host}/api/epg/sources/${sourceId}/sd-lineups/search/`,
+        {
+          method: 'POST',
+          body: { country, postal_code: postalCode },
+        }
+      );
+      return response;
+    } catch (e) {
+      errorNotification('Failed to search lineups', e);
+      return [];
+    }
+  }
+
+  static async getSDLineups(sourceId) {
+    try {
+      const response = await request(
+        `${host}/api/epg/sources/${sourceId}/sd-lineups/`,
+        { method: 'GET' }
+      );
+      return response;
+    } catch (e) {
+      errorNotification('Failed to fetch lineups', e);
+      return [];
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


SergeantPanda:
Thank you for the PR.

We do have concerns with implementing Schedules Direct (I know it was half baked since the beginning).

Ultimately we put a pause on developing the integration because SD is strict on number of API calls and will ban your account. (Happened to one of our devs). 

Something like this would need to be tested extensively to ensure it follows SD guidelines, would probably be best to get Dispatcharr set as an approved app with them too as I believe that allows for more api calls (if memory serves correctly).

Issue #992, opened 2026-02-17
[Bug]: Multi-worker gunicorn: Non-owner cleanup removes active stream buffers → orphaned channels


firestaerter3:
### Describe the bug

When running gunicorn with multiple workers (default `--workers=4`), Worker B's periodic cleanup routine removes stream buffers actively owned by Worker A. This causes all connected clients to lose their streams and the channel becomes orphaned.

**Timeline from logs:**

1. **Worker 1441** starts streaming channel `213ac16e`, HTTP source connects, data starts flowing
2. **Worker 1381** receives a second client request for the same channel, sees it "already being initialized"
3. **~19 seconds later**, Worker 1381 runs "Non-owner cleanup":
   ```
   INFO ts_proxy.server Non-owner cleanup: Removed stream buffer for channel 213ac16e-...
   INFO ts_proxy.server Non-owner cleanup: Removed client manager for channel 213ac16e-...
   ```
4. Worker 1381's stream generator terminates:
   ```
   INFO ts_proxy.stream_generator [client_...] Channel buffer no longer exists, terminating stream
   ```
5. Channel becomes permanently orphaned:
   ```
   WARNING ts_proxy.server Found orphaned channel 213ac16e-... with 2 clients but no owner - may need ownership takeover
   ```

Both clients receive no data. The only recovery is restarting Dispatcharr.

**Workaround:** Setting `--workers=1` prevents the issue, but then a single dead upstream connection can block the entire server (the gevent worker becomes unresponsive when an upstream provider resets the TCP connection and the buffer never fills).

### How can we recreate this bug?

1. Configure gunicorn with `--workers=4 --worker-class=gevent` (the default)
2. Open any channel via `/proxy/ts/stream/<uuid>` or `/live/<user>/<pass>/<stream_id>.ts`
3. Open the same channel from a second client (or have Emby request media sources then start playback)
4. Within ~20 seconds, the non-owner worker runs cleanup and kills the stream buffer
5. Both clients stop receiving data
6. Logs show "Non-owner cleanup: Removed stream buffer" followed by "Found orphaned channel with N clients but no owner"

### Dispatcharr Version

Git HEAD `2b9106f8` (2026-02-15), installed from source (not Docker)

### Docker Version

Not using Docker — running directly via systemd + gunicorn

### What's in your Docker Compose file?

Not using Docker Compose. Running via systemd:

```bash
#!/usr/bin/env bash
cd /opt/dispatcharr
set -a
source .env
[...]

Issue #980, opened 2026-02-13
RedisClient.get_client() calls flushdb() on every process init, nuking cross-process data


FiveBoroughs:
`core/utils.py:84` calls `client.flushdb()` every time a new process initializes its Redis connection. The problem is that multiple processes start up at different times (uWSGI workers, Celery workers, Celery beat, Daphne), and each one flushes Redis independently, wiping whatever the previous processes already wrote.

This means late-starting processes delete data from earlier ones, including:
- Celery broker messages (pending tasks)
- File mtime tracking (`processed_file:*` keys)
- Task locks (`task_lock_*` keys)

This is already handled correctly by `scripts/wait_for_redis.py`, which flushes Redis once at boot before any app process starts. The `flushdb()` in `get_client()` is redundant.

Everything stored in Redis already has TTLs (locks 300s, mtime tracking 3 days, buffers ~30s), so stale data cleans itself up.

**Fix:** Remove line 84 in `core/utils.py` (`client.flushdb()`).

FiveBoroughs:
Confirmed with logging. Added a `print()` before `flushdb()` to capture PID and key count on a fresh container start:

```python
client.ping()
key_count = client.dbsize()
sample = client.keys('*')[:10]
print(
    f"[FLUSHDB][PID {os.getpid()}] about to wipe "
    f"{key_count} keys: {sample}",
    file=sys.stderr, flush=True
)
client.flushdb()
```

```
[FLUSHDB][PID 283] about to wipe 0 keys: []
[FLUSHDB][PID 284] about to wipe 1 keys: [b'ts_proxy:worker:3a284890bc1d:283:heartbeat']
[FLUSHDB][PID 285] about to wipe 1 keys: [b'ts_proxy:worker:3a284890bc1d:284:heartbeat']
[FLUSHDB][PID 295] about to wipe 1 keys: [b'ts_proxy:worker:3a284890bc1d:285:heartbeat']
[FLUSHDB][PID 293] about to wipe 1 keys: [b'ts_proxy:worker:3a284890bc1d:295:heartbeat']
[FLUSHDB][PID 297] about to wipe 1 keys: [b'ts_proxy:worker:3a284890bc1d:293:heartbeat']
[FLUSHDB][PID 286] about to wipe 1 keys: [b'ts_proxy:worker:3a284890bc1d:297:heartbeat']
```

[...]

PR #1035, opened 2026-02-27
Fix stream ownership bugs causing streams to die after 30-200s

PFalko:
## Disclosure

This fix was developed with AI assistance (Claude). I completely understand if that's a concern — please feel free to close this PR if you'd prefer a human-only contribution. No hard feelings at all! I'm sharing it in case it's helpful, since I spent quite a bit of time debugging the root cause and others seem to be hitting the same issue.

**This PR contains no credentials, API keys, or sensitive data.** The diff is small (58 insertions, 13 deletions across 3 files) and straightforward to review.

## Problem

TS proxy streams die after 30–200 seconds in multi-worker uWSGI deployments. Three interrelated bugs:

### 1. Double ProxyServer instantiation (`apps/proxy/apps.py`)

`ProxyConfig.ready()` calls `TSProxyServer()` directly, while `TSProxyConfig.ready()` calls `TSProxyServer.get_instance()`. Both Django apps run during startup, creating **two ProxyServer instances per worker** — each with its own cleanup thread. The orphaned instance's cleanup thread can't extend ownership because it has no channels in its `stream_managers` dict.

### 2. `flushdb()` on every Redis client init (`core/utils.py`)

`RedisClient.get_client()` calls `client.flushdb()` whenever `_client is None`. With Celery's `--autoscale=6,1`, new worker processes spawning mid-stream call `get_client()` → `flushdb()`, which **nukes all Redis keys** including active ownership keys, client records, and channel metadata.

### 3. No recovery from expired ownership (`ts_proxy/server.py`)

- `get_channel_owner()` calls `redis.get()` twice in a lambda (TOCTOU race — key can expire between the two calls)
- `extend_ownership()` silently returns `False` when the key expires, with no re-acquisition logic
- Non-owner cleanup unconditionally kills streams via `_cleanup_local_resources()`, even when the worker holds the `stream_manager`

## Changes

| File | Change |
|------|--------|
| `apps/proxy/apps.py` | Use `TSProxyServer.get_instance()` instead of `TSProxyServer()` |
| `core/utils.py` | Remove `flushdb()` from Redis client initialization |
| `apps/proxy/ts_proxy/server.py` | Gevent-safe singleton using sentinel pattern (`threading.Lock` doesn't work with gevent greenlets) |
| `apps/proxy/ts_proxy/server.py` | Single `GET` in `get_channel_owner()` to fix TOCTOU race |
| `apps/proxy/ts_proxy/server.py` | `extend_ownership()` re-acquires expired keys via `SETNX` if worker has `stream_manager` |
| `apps/proxy/ts_proxy/server.py` | Non-owner cleanup attempts re-acquisition before killing streams |

## How I found it

1. Observed streams dying after ~30s with "Non-owner cleanup" in logs
2. Redis monitoring showed ownership key TTL going from 29 → -2 (expired, not deleted)
3. Added debug logging to `__init__` — found `Worker ID` logged **twice per worker** (double instantiation)
4. Stack trace showed first instance from `apps/proxy/apps.py:17` (`TSProxyServer()`), second from `ts_proxy/apps.py:13` (`get_instance()`)
5. Disabling `flushdb()` fixed the remaining ~3 minute expiry caused by Celery autoscale

## Testing

- Before fix: streams die after 27–200 seconds (depending on Celery autoscale timing)
- After fix: streams survive 10+ minutes with zero ownership gaps, zero re-acquisition events
- Verified single `Worker ID` log per worker (was 2 before)
- Patch is idempotent (safe to apply multiple times)

## Relates to

- #992 (Multi-worker gunicorn: Non-owner cleanup removes active stream buffers)
- #980 (RedisClient.get_client() calls flushdb() on every process init)

Thank you for building Dispatcharr — it's a great project! Happy to discuss any of these changes or adjust the approach if you'd prefer a different solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Channel table filter for channels that have stale streams: A new "Has Stale Streams" filter option in the channel table header menu highlights and filters channels containing at least one stale stream. Channels with stale streams are visually distinguished with an orange tint. The filter is mutually exclusive with "Only Empty Channels". - Thanks [@JCBird1012](https://github.com/JCBird1012)
 
+### Fixed
+
+- TS proxy streams dying after 30–200 seconds in multi-worker uWSGI/Celery deployments, caused by three interrelated bugs. (Fixes #992, #980) - Thanks [@PFalko](https://github.com/PFalko)
+  - **Double ProxyServer instantiation**: `ProxyConfig.ready()` called `TSProxyServer()` directly while `TSProxyConfig.ready()` also called `TSProxyServer.get_instance()`, creating two instances per worker — each with its own cleanup thread. The orphaned thread could not extend ownership because it had no entries in `stream_managers`. Fixed by using `TSProxyServer.get_instance()` in `ProxyConfig.ready()`.
+  - **`flushdb()` on every Redis client init**: `RedisClient.get_client()` called `client.flushdb()` whenever `_client` was `None`. Celery autoscale (`--autoscale=6,1`) spawning new workers mid-stream triggered this path, nuking all Redis keys including active ownership keys, client records, and channel metadata. Removed the `flushdb()` call entirely.
+  - **No recovery from expired ownership**: `get_channel_owner()` called `redis.get()` twice inside a lambda (TOCTOU race — key could expire between calls); `extend_ownership()` silently returned `False` on expiry with no re-acquisition; and the non-owner cleanup path unconditionally killed streams even when the worker held the `stream_manager`. Fixed with a single `GET` in `get_channel_owner()`, re-acquisition via atomic `SET NX EX` in `extend_ownership()`, and a re-acquisition attempt with client-aware cleanup deferral in the cleanup thread.
+- `get_instance()` deadlock: if `ProxyServer()` raised an exception during singleton construction, `_instance` was left permanently as the `_INITIALIZING` sentinel, causing all subsequent `get_instance()` callers to spin in an infinite `gevent.sleep()` loop. Construction is now wrapped in `try/except`; on failure `_instance` resets to `None` so the next call can retry.
+- Non-atomic ownership acquisition in `try_acquire_ownership()`: replaced the separate `setnx()` + `expire()` calls with a single atomic `SET NX EX`, eliminating the race window where a process crash between the two calls could leave an ownership key with no TTL (permanent ownership lock).
+
 ## [0.20.2] - 2026-03-03
 
 ### Security


firestaerter3:
This solved all my issues :-)

SergeantPanda:
So, this does similar things to #949 but they are definitely a little different.

I do like how minimal the changes are in this and I do like the find on the flushing of Redis, that is a good catch and probably is the root of some hard to track down issues.

Overall I feel pretty comfortable with this one but I want @CodeBormen to take a look and see how it plays with his PR that is currently in a draft state. I know he has put a lot of work into that one.

CodeBormen:
All good, #949 is focused pretty narrowly on connection counter hardening and TOCTOU, there is no conflict with that branch so all good! I was actually going to go back and work on client.flushdb() issues after 949 gets merged, so I'm happy to see this!

CodeBormen:
I have been running into this issue a fairly large amount actually while doing testing for DVR changes I am working on. After looking at this branch it is pretty good, but if you don't mind I have a couple recommended changes:

**SETNX + EXPIRE should be atomic in extend_ownership()**

Your current code:

```
acquired = self.redis_client.setnx(lock_key, self.worker_id)
if acquired:
    self.redis_client.expire(lock_key, ttl)
```

I recommend:
`acquired = self.redis_client.set(lock_key, self.worker_id, nx=True, ex=ttl)`

**time.sleep() in get_instance() spin-wait blocks gevent**

Should be gevent.sleep(0.05) instead of time.sleep(0.05)

**When another worker takes ownership, _cleanup_local_resources() kills the stream that is still being used.**

For this one maybe consider a continue or deferment so that the new process owner can take over

PFalko:
@CodeBormen Thank you so much for taking the time to review this so thoroughly — really appreciate it, especially since I know you've been deep in the proxy code with #949!

All three suggestions are spot on. I've pushed a commit addressing each one:

1. **Atomic `SET NX EX`** — Replaced the separate `setnx()` + `expire()` with a single `redis.set(lock_key, self.worker_id, nx=True, ex=ttl)` to eliminate the race window.

2. **`gevent.sleep()` in `get_instance()`** — Swapped `time.sleep(0.05)` for `gevent.sleep(0.05)` so the spin-wait properly yields to the greenlet hub.

3. **Deferred cleanup for active clients** — When ownership re-acquisition fails but the channel still has connected clients, we now defer to the next cleanup cycle instead of immediately tearing down. Channels with zero clients still get cleaned up right away.

**Transparency note:** As mentioned in the PR description, these changes were also made with AI assistance (Claude). Happy to adjust anything further or discuss alternative approaches!

Thanks again for the great feedback — and for all the work you're putting into the proxy improvements. 🙏

PR #1037, opened 2026-02-27
fix: debian_install.sh regression after uv migration (pip-less venv + missing gunicorn)

marcinolek:
This PR fixes debian_install.sh on clean/minimal Debian 13 (Proxmox LXC) installs.

After the switch to uv, the installer can fail because:

1) The venv created in minimal environments may not have pip available,
   which prevents dependencies from being installed and results in
   `ModuleNotFoundError: django` during migrations.

2) The systemd unit expects `/opt/dispatcharr/env/bin/gunicorn`, but gunicorn
   is not guaranteed to be present after `uv sync` (not installed as a runtime
   dependency), causing the dispatcharr service to fail at ExecStart.

3) Django requires DJANGO_SECRET_KEY; the installer did not persist/load it
   consistently for both `manage.py` steps and systemd services.

Fixes:
- Create the venv via `python -m venv` and run `ensurepip` to guarantee pip exists.
- Keep uv for dependency installation (`uv sync --no-dev`) into the target env.
- Ensure gunicorn is installed so the systemd unit can start.
- Persist DJANGO_SECRET_KEY in /opt/dispatcharr/.env, source it for migrations,
  and load it for all systemd services via EnvironmentFile.

Only debian_install.sh and the generated service units are affected; Docker configs
are unchanged.

### DIFF
--- a/debian_install.sh
+++ b/debian_install.sh
@@ -168,23 +168,50 @@ install_uv() {
 
 setup_python_env() {
   echo ">>> Setting up Python virtual environment with UV..."
-  # Install UV globally first
-  install_uv
 
   su - "$DISPATCH_USER" <<EOSU
-cd "$APP_DIR"
-export PATH="\$HOME/.local/bin:\$PATH"
-# Install UV for the dispatch user if not already available
-if ! command -v uv &> /dev/null; then
-  curl -LsSf https://astral.sh/uv/install.sh | sh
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you!

digason:
This doesn't resolve the problem.  The ensurepip command should be run after uv sync.

PR #1049, opened 2026-03-03
Tests/frontend unit tests

nick4810:

components/

Included some slight refactoring for readability.
Also extracted YouTubeTrailerModal into its own component.

### DIFF
--- a/frontend/src/components/M3URefreshNotification.jsx
+++ b/frontend/src/components/M3URefreshNotification.jsx
@@ -11,7 +11,7 @@ import { useNavigate } from 'react-router-dom';
 import { CircleCheck } from 'lucide-react';
 import { showNotification } from '../utils/notificationUtils.js';
 
-const M3uSetupSuccess = (data) => {
+const M3uSetupSuccess = ({ data }) => {
   const navigate = useNavigate();
 
   const onClickRefresh = () => {


SergeantPanda:
Hey Nick,

This one seems like a lot of just formatting changes. I don't mind good refactors and do like the test additions but we currently have a big rewrite of the frontend happening and making unnecessary formatting changes is going to make merge conflicts quite problematic.

Mind getting rid of the formatting changes and just sticking with the refactor and tests?

nick4810:
@SergeantPanda I reverted some of the styling changes. Lmk if you want anything else updated.

frontend/src/components/M3URefreshNotification.jsx
```diff
@@ -10,21 +9,54 @@ import { Stack, Button, Group } from '@mantine/core';
...
+  );
+}
 
 export default function M3URefreshNotification() {
   const playlists = usePlaylistsStore((s) => s.playlists);
   const refreshProgress = usePlaylistsStore((s) => s.refreshProgress);
   const fetchStreams = useStreamsStore((s) => s.fetchStreams);
   const fetchChannelGroups = useChannelsStore((s) => s.fetchChannelGroups);
-  const fetchChannelIds = useChannelsStore((s) => s.fetchChannelIds);
+  const fetchChannels = useChannelsStore((s) => s.fetchChannels);
```
SergeantPanda: Is there a bug you found for this change? We did refactor the channel store with the last version so we only hold all ID's now instead of all channel objects. Channels are only fetched on demand with pagination now. Sorry, I didn't dig through all of the changes yet to see if there was a good reason for it.

SergeantPanda (approved):

frontend/src/components/M3URefreshNotification.jsx
nick4810: 👍 This wasn't a bug fix or anything. I believe I just missed the change from the channel store refactor.

PR #1052, opened 2026-03-03
Fix modular mode deployment issues (#1045)

### LINKED ISSUE #1045: [Bug]: Modular Mode Postgres Checl
### Describe the bug

Environment:
- Dispatcharr version: 0.20.1
- Host PostgreSQL version: 18
- Container PostgreSQL version: 17
- Mode: modular

Issue:
- Modular mode fails at version check
- AIO mode works with same PostgreSQL 18 database
- Error: "Unable to determine external PostgreSQL version"

Expected:
- PostgreSQL 18 should be supported (backward compatible)
- Version check should succeed in modular mode

Workaround:
- Using AIO mode with external PostgreSQL works

Run container in modular mode and attempt to connect to PostgreSQL 18 

### Dispatcharr Version

0.20.1

### Docker Version

podman version 5.7.1

### Client Information

CodeBormen:
- Fix Postgres version check failing with restricted DB users (use $POSTGRES_DB instead of hardcoded 'postgres')
- Fix DVR recording broken in modular mode (respect DISPATCHARR_PORT instead of hardcoding 9191)
- Remove flushdb() from wait_for_redis.py to prevent Redis data loss on container restart
- Add DISPATCHARR_PORT to celery environment in docker-compose.yml
- Add depends_on health conditions for proper service startup ordering
- Add extra_hosts for host.docker.internal resolution on Linux
- Harden celery entrypoint with timeouts for JWT wait (120s) and migration wait (300s)
- Replace fragile showmigrations grep with migrate --check
- Add unit tests for DVR port resolution and flushdb removal regression

Fixes #1045 and also corrects other known issues reported from users. Prevents external Redis flushdb() which may be in use for other external applications. Adds some additional hardening and testing for services to guard future changes (migrations, port changes).


SergeantPanda:
I'm ok with all of the changes here I'd say except for the removal of the flush db for Redis.

I think Dispatcharr starting with a clean Redis database is important. Otherwise we could hang on to streamlocks that persist through restarts. This is why we had it flush on startup.

If it is still being done somewhere else that is OK though.

CodeBormen:
I get what you mean. I had taken out the flushdb() because it executes on every web service restart (including crashes), in modular mode since the database may not be running in the same container a flushdb() wipes all locks and causes it to lose state. In AIO it's not a problem since the stack starts together.

I can think of a couple other approaches that solve the issue: 1) we could try to do targeted removal of stale locks, or 2) we could move the flushdb() to the entrypoint where it's guaranteed to only run before services start which would prevent any current services from losing state, and ensure that it gets a clean db.

Second approach sounds much stronger to me - thoughts?

SergeantPanda:
But shouldn't Redis be cleared every time the web service is restarted? IIRC (it's been a while since I set this up) I had the flush in the wait for redis script to ensure that Redis was available before sending the flush and since uwsgi is calling the wait for redis script it also doesn't start until after the Redis DB has been flushed.

CodeBormen:
Hmmm, I get where you're coming from yeah. 

Personally I think there are pros and cons to each. In AIO mode the whole stack manages itself so it makes a lot of sense. In modular mode, if the web container crashes, celery will lose its states with flushdb(), but that is a fairly low risk. Might cause some issues where celery duplicates tasks if the web container crashes and redis is flushed. The risk of not doing a flushdb() is that any locks without a TTL will persist. But I do see the benefit of a flush on every start. 

In reality, I think both scenarios are fairly low risk. Totally up to you, I could also write the flushdb() to exclude only modular mode:

if os.environ.get('DISPATCHARR_ENV') != 'modular':
    redis_client.flushdb()

or I could just put it back in for both, again I think it is splitting hairs either way

SergeantPanda:
I think there is a bigger risk to not flushing the database personally. I could see being selecting in modular mode though if you have celery concerns (which are valid).

For sure we need to clear out Redis keys for proxy and stream locks. Really anything BUT Celery.

CodeBormen:
If you'd like I could keep it as it was for AIO mode (works fine in AIO), and in Modular mode delete all keys except celery*, _kombu*, unacked* which would mean that all non-Celery keys get removed.

Again, fairly low risk I think to just keeping it how it was, so if you prefer that for simplicity, I'm sure there wouldn't be a lot of issues caused by it. Maybe every once in a while for people crashing the app with weird behavior or having really low-resourced machines.

SergeantPanda:
Yeah I'd be OK with the selective cleanup method. Seems a good way to go about it to me.

CodeBormen:
I'm glad you called this out because now that I looked further into it, flushdb() does not currently run in modular mode at all because DISPATCHARR_ENV is not set inside of the "su -" env that is used to call wait_for_redis.py so it always runs on localhost

I'm writing in the selective cleanup method along with ensuring that it will work with DISPATCHARR_ENV and testing!

CodeBormen:
Ok, this is ready to go now. I changed the flushdb() behavior back to what it was for AIO which is fine since the celery process  isn't separate. For modular mode, flushdb() targets non-Celery keys and prevents state loss when only the web container restarts. I did confirm that this works as intended, with Celery keys persisting flushdb(). I did have to move the wait_for_redis call from uWSGI to the entrypoint for modular mode so that the DISPATCHARR_ENV=modular is available to the environment. Neither change affects AIO.

While I was looking I also found that dispatcharr.sh re-uses environment variables on restart instead of re-writing them. This isn't really a problem for AIO since it will always default to localhost, but for modular mode it is a problem because the end user may change variables that the web container relies on which would result in a breakage upon restart. Instead, it re-writes variables on every restart which will effectively fix that issue, and does also provide value to AIO in the case a user switches a variable and restarts the container for it to take effect (especially if pipeline updates are being used).

Also a small change to quoting so that postgres passwords containing certain special characters will be exported correctly.

SergeantPanda:
Yes, I like this much better now. Thank you!

PR #1057, opened 2026-03-04
fix(gevent): patch sockets for Celery Redis without corrupting import locks

endoze:
Celery's Redis transport needs gevent-patched sockets to work under uWSGI's cooperative scheduling — unpatched sockets in the connection pool silently fail. However, a full monkey.patch_all() replaces _thread.get_ident(), changing the current thread's identity from a native ID to a greenlet ID. In debug mode, uWSGI loads scripts.debug_wrapper via import, holding a per-module lock. When wsgi.py calls patch_all() during that nested import, the identity changes mid-flight — so when the outer import of debug_wrapper completes, _ModuleLock.release() sees a different owner than who acquired it and raises RuntimeError.

Using patch_all(thread=False) patches sockets, SSL, and DNS — everything Celery's Redis transport needs — while preserving _thread.get_ident() so import lock ownership stays consistent. uWSGI's gevent loop engine handles green-thread scheduling at the C level independent of Python's threading module.

The patching logic is extracted to a shared shim (gevent_patch.py) that gates on uwsgi.opt having a gevent key and skips if already patched, so both entry points call it and the second call is a no-op.


endoze:
@SergeantPanda I'm working through this on my linux desktop to ensure I recreate it all correctly, just wanted to give you a heads up I haven't given up yet 😄. Thanks for the information to debug this properly.

SergeantPanda:
Thank you for the update!

endoze:
@SergeantPanda it seems like some of the recent changes around gevent fixed all the issues I was having. Going to close this as it's no longer needed. Thanks for a great project!

SergeantPanda:
Happy to hear that!

Issue #1061, opened 2026-03-05
ValueError: invalid literal for int() with base 10: 'None' in apps.m3u.tasks due to is_adult attribute


markkoza:
### Describe the bug

When processing an M3U playlist or Xtream API stream, the apps.m3u.tasks module crashes if a stream's is_adult attribute is passed as the literal string "None" instead of a standard integer.

### How can we recreate this bug?

1. Connect an IPTV provider source that supplies stream attributes where is_adult is set to the string "None".
2. Trigger the playlist synchronization or stream parsing task (apps.m3u.tasks).
3. Check the application logs.

### Dispatcharr Version

20.02

### Docker Version

Docker version 29.2.1, build a5c7197

### What's in your Docker Compose file?

---
services:
  dispatcharr:
    # build:
    #   context: .
    #   dockerfile: Dockerfile
    image: ghcr.io/dispatcharr/dispatcharr:latest
    container_name: dispatcharr
    restart: unless-stopped
    network_mode: "container:gluetun"
    #ports:
    #  - 9191:9191
    volumes:
      - dispatcharr_data:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - CELERY_BROKER_URL=redis://localhost:6379/0
      - DISPATCHARR_LOG_LEVEL=info
      # Process Priority Configuration (Optional)
      # Lower values = higher priority. Range: -20 (highest) to 19 (lowest)
      # Negative values require cap_add: SYS_NICE (uncomment below)
      #- UWSGI_NICE_LEVEL=-5   # uWSGI/FFmpeg/Streaming (default: 0, recommended: -5 for high priority)
      #- CELERY_NICE_LEVEL=5   # Celery/EPG/Background tasks (default: 5, low priority)
    #
    # Uncomment to enable high priority for streaming (required if UWSGI_NICE_LEVEL < 0)
    #cap_add:
    #  - SYS_NICE
    # Optional for hardware acceleration
    #devices:
    #  - /dev/dri:/dev/dri  # For Intel/AMD GPU acceleration (VA-API)
    # Uncomment the following lines for NVIDIA GPU support
    # NVidia GPU support (requires NVIDIA Container Toolkit)
    #deploy:
    #  resources:
    #      reservations:
    #          devices:
    #              - driver: nvidia
    #                count: all
    #                capabilities: [gpu]

#volumes:
#  dispatcharr_data:

volumes:
  dispatcharr_data:
[...]

PR #1062, opened 2026-03-05
fix: guard is_adult parsing against invalid values

JCBird1012:
or whatever other fun things providers might shove down the wire... Long term, it might be worth making a more generic parser to guard against any BS/invalid values that Dispatcharr ingests, for other fields (`channel_group_id` maybe? since that also gets cast and a None string would break it?)

fixes: #1061


SergeantPanda:
Checking with the user to see what an adult channel has as a string, maybe we could parse either or for wider support.

JCBird1012:
The way I interpreted the issue is that the upstream provider was doing something incorrect by even passing a string in the first place. The `None` type that was causing the original error was literally the string "None" that couldn't be cast to an `int` by Python cleanly - definitely seemed like a bug on the provider side - the provider should have caught that and just omitted it or set the value to `0` instead. If providers were sending string "1" and string "0" to Dispatcharr, it would have no issue with the casting to an integer (so we already had support there, though this PR would break that technically - EDIT - fixed!).

I think it's just something to guard against, not try to support - but of course, what really is the XC specification besides what's actually happening in the wild? Providers could, in theory, pass "True" or "False" - so we could add support for that, but eh.

JCBird1012:
My logic had a bug before, last commit fixes that. 

Anything but a "1" or a "0" will just get treated like a "0" for `is_adult` now - I think that's sane.

JCBird1012:
I didn't bother reverting/squashing - just hard reset and force pushed back to my first commit. Not the "proper" way but oh well.

Let me know how that looks now.

Issue #1012, opened 2026-02-24
[Bug]: Celery worker memory grows with each M3U refresh and is never reclaimed


CodeBormen:
### Describe the bug

Each time an M3U account refreshes, the Celery worker's memory usage increases by roughly 80MB (for a ~20K stream account) and does not fully return to its previous level afterward. Over multiple refresh cycles, memory continues to creep upward. Since there is no worker_max_memory_per_child or worker_max_tasks_per_child configured, the worker process is never recycled, so memory will eventually be exhausted with scheduled refreshes.

The root cause appears to be twofold:

The garbage collection cleanup code at the end of refresh_single_m3u_account (the del statements and cleanup_memory() call) is placed after the finally block, making it unreachable whenever an exception is raised — the large data structures (extinf_data, groups, batches, and the parsed JSON data dict) are never freed in those cases.

The standard M3U batch processor (process_m3u_batch_direct) has its garbage collection lines commented out, so temporary structures like streams_to_create, streams_to_update, and existing_streams accumulate across all 14+ thread batches without cleanup.

### How can we recreate this bug?

1. Add an M3U account with a reasonably large stream list (10K+ streams)
2. Set the refresh interval to something short
3. Monitor the Celery worker's memory usage over several refresh cycles — check the Docker logs for the "Memory usage after" lines that are logged after each task
4. See that memory increases with each cycle and never returns to baseline

### Dispatcharr Version

v0.19.0 (dev branch)

### Docker Version

v29.2.0

### What's in your Docker Compose file?

Standard AIO file

### Client Information

Any, issue is server-side

CodeBormen:
Just reporting this for visibility and happy to work on a fix for this myself, but would like to get some feedback on whether gc.collect() was commented out intentionally for any purpose?

kmaid:
I'm experiencing the same issue with similar scale:

- **Channels**: ~40,000 in playlist, only ~146 selected for guide
- **Memory usage**: ~1GB RSS after fresh container boot
- **Dispatcharr version**: ghcr.io/dispatcharr/dispatcharr:dev (v0.19.0)
- **Config**: AIO mode with Redis on localhost

docker-compose config:
```yaml
dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:dev
    platform: linux/amd64
    restart: always
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
[...]

Issue #1053, opened 2026-03-03
[Bug]: High memory usage with large M3U playlists (~40K channels)


kmaid:
## Describe the bug

High memory usage (1GB+) with large M3U playlists even when only a small fraction of channels are selected for the guide.

## Expected behavior

Memory usage should be proportional to the number of selected channels, not the total playlist size.

## Actual behavior

- ~40,000 channels in playlist
- Only ~146 channels selected for guide
- Memory usage: ~1GB RSS after fresh container boot
- Memory usage seems disproportionate to the number of selected channels (146)

## Dispatcharr Version

ghcr.io/dispatcharr/dispatcharr:dev (v0.19.0)

## Docker Configuration

```yaml
dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:dev
    platform: linux/amd64
    restart: always
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - CELERY_BROKER_URL=redis://localhost:6379/0
      - DISPATCHARR_LOG_LEVEL=info
    volumes:
      - /var/media-server/jelly/dispatcharr:/data
    devices:
      - /dev/dri:/dev/dri
```

## Additional context

This may be related to issue #1012 (Celery worker memory grows with each M3U refresh), but this is about memory usage at startup with a large playlist, not memory growth over time.

kmaid:
Additional detail: The memory is being consumed by the Celery worker process:

```
PID   USER     PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
4  1000     20   0  2.9G  1.44G 435M S  0.0 18.1   0:09:56 celery
```

Process: `celery -A dispatcharr worker --autoscale=6,1`

This is after only a few minutes of runtime. The Celery worker is consuming ~1.44GB RSS.

CodeBormen:
It is possible that this is celery (need to look into it more), but I may have already addressed this in PR #1014. That PR included streaming M3U to disk instead of memory, and also included fixes to prevent multiple copies of large M3Us multiplying in memory. I see you're on v0.19.0, those changes would be available on v0.20.0+

If you don't mind upgrading and letting me know, otherwise when I get to working on a fix for #1012  I will make sure I analyze these symptoms along with it

kmaid:
Dispatcharr version: 0.20.2-20260304002340

[...]

PR #1063, opened 2026-03-05
fix(m3u): fix Celery worker memory leak during M3U/XC refresh (#1012, #1053)

CodeBormen:
- Restructure `refresh_single_m3u_account()` with an outer `try/finally` that guarantees lock release and memory cleanup on all exit paths (success, exception, early return)
- Move `del` statements into a `finally` block so they execute **before** `task_postrun`'s `gc.collect()`, enabling cyclic garbage collection
- Uncomment and fix batch cleanup in `process_m3u_batch_direct()`
- Add `CELERY_WORKER_MAX_MEMORY_PER_CHILD = 512MB` as a safety net against pymalloc arena fragmentation

**Problem**

Celery worker memory grows ~20-80MB per M3U refresh cycle and never returns to baseline. For large playlists (40K-89K+ channels), RSS reaches 500MB-1GB+ and stays there indefinitely.

Root cause: In refresh_single_m3u_account(), the del statements that free large data structures (stream lists, group dicts, batch arrays) were placed after the finally block, making them unreachable on exception or re-raise paths. More critically, even on the success path, they ran after the task_postrun signal's gc.collect() call. Since del breaks the reference cycles and gc.collect() reclaims them, executing del after gc.collect() means cyclic garbage is never collected. The objects accumulate across refresh cycles with no mechanism to reclaim them.

**Fix**

1. Guaranteed cleanup via try/finally restructure — Extract the function body into _refresh_single_m3u_account_impl() wrapped in an outer try/finally that handles lock release. The inner finally block runs del on 8 large variables (existing_groups, extinf_data, groups, batches, all_xc_streams, data, filtered_groups, channel_group_relationships). This ensures del runs before task_postrun's gc.collect(), enabling cyclic garbage collection on every exit path.

2. Batch processor cleanup — Re-enable del + connections.close_all() in process_m3u_batch_direct() (was commented out). Each batch holds ~1,500 stream objects; 60+ batches for large playlists.

3. Early del data — Free the top-level JSON dict immediately after extracting extinf_data and groups references, rather than holding it for the full processing duration.

4. Free XC stream list after batching — del all_xc_streams after slicing into batches, since list slicing creates independent copies.

5. worker_max_memory_per_child = 512MB — Safety net for pymalloc arena fragmentation. CPython's pymalloc allocator acquires 256KB arenas from the OS but only releases them when every allocation within the arena is freed. Even with perfect del/gc.collect(), memory fragmentation means RSS may not fully return to baseline. This setting recycles prefork workers after task completion if RSS exceeds 512MB, providing a hard ceiling. Workers finish their current task before recycling — no mid-task kills.

Test Results (Real provider data, 89K streams)
| Metric | Baseline (before fix) | Fixed (after fix) |
|--------|----------------------|-------------------|
| Memory freed by `gc.collect()` per refresh | **0.00 MB** (every cycle) | **1-20 MB** per cycle |
| Peak RSS during refresh | 512 MB | 329 MB |
| Post-cleanup RSS | 266-346 MB (fluctuating) | 259 MB (stable) |
| Unbounded growth protection | None | 512 MB ceiling via worker recycling |

The baseline never freed memory from refresh cleanup across any test cycle — gc.collect() ran before del broke the reference cycles, so cyclic garbage was unreachable. The fix enables consistent memory reclamation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Celery worker memory leak during M3U/XC refresh causing 20–80 MB growth per cycle with no reclamation (Fixes #1012, #1053) - Thanks [@CodeBormen](https://github.com/CodeBormen)
+  - Restructured `refresh_single_m3u_account()` with a `try/finally` that guarantees `del` of large data structures runs before Celery's `gc.collect()`, and lock release on all exit paths (success, exception, early return)
+  - Re-enabled batch data cleanup in `process_m3u_batch_direct()` (was commented out)
+  - Added `CELERY_WORKER_MAX_MEMORY_PER_CHILD = 512 MB` as a safety net against pymalloc arena fragmentation
 - EPG output was filtering programs using `start_time__gte=now` when the `days` parameter was specified, which caused currently-airing programs (started before the request time but not yet ended) to be omitted from the XML output. This produced a gap in clients' guides immediately after an EPG refresh, lasting until the next program started. Fixed by changing the filter to `end_time__gte=now` so any program that has not yet finished is included.
 - TS proxy connection slot leaks and TOCTOU races in stream initialization (Fixes #947) - Thanks [@CodeBormen](https://github.com/CodeBormen)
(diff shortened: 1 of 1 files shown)


CodeBormen:
Fixes #1012 , #1053

SergeantPanda (approved):
This looks good. Thank you again for the work on this!

PR #1067, opened 2026-03-05
fix(install): harden debian_install.sh for non-UTF8 environments

marcin-spotio:
## Summary
This PR hardens the `debian_install.sh` script to handle environments where UTF-8 locales are not pre-configured (common in minimal LXC containers) and ensures that background services have access to necessary system tools.

## Changes
- **Locales**: Added `setup_locales` to ensure `en_US.UTF-8` is generated and set as the system default. This prevents PostgreSQL from defaulting to `SQL_ASCII` encoding.
- **PostgreSQL**: Explicitly use `-E UTF8` during database creation.
- **Systemd Services**: Updated `PATH` in Celery and Daphne service files to include standard system directories (`/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin`). This fixes issues where background tasks couldn't find `ffmpeg` or `ffprobe`.

## Test plan
1. Run the installer in a fresh, minimal Debian LXC container (where `locale` returns `C`).
2. Verify that the `dispatcharr` database is created with `UTF8` encoding.
3. Verify that M3U imports containing Unicode characters succeed.
4. Verify that background tasks correctly find and use system tools like `ffmpeg`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/debian_install.sh
+++ b/debian_install.sh
@@ -12,8 +12,19 @@ fi
 trap 'echo -e "\n[ERROR] Line $LINENO failed. Exiting." >&2; exit 1' ERR
 
 ##############################################################################
-# 0) Warning / Disclaimer
+# 0) Locales & Warning / Disclaimer
 ##############################################################################
+
+setup_locales() {
+  echo ">>> Setting up locales..."
+  apt-get update
+  apt-get install -y locales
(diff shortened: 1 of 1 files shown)


marcin-spotio:
Closing to recreate from personal GitHub account. The code branches remain available on my fork.

PR #1068, opened 2026-03-05
fix(accounts): allow passwords with leading/trailing spaces

marcin-spotio:
## Summary
This PR fixes an issue where users were unable to log in if their password contained leading or trailing spaces. This occurred because Django Rest Framework's `CharField` trims whitespace by default.

## Changes
- **TokenObtainPairView**: Introduced `CustomTokenObtainPairSerializer` which explicitly sets `trim_whitespace=False` for the password field.
- **UserSerializer**: Updated the password field to set `trim_whitespace=False`.

This ensures that the password provided during login exactly matches the one stored in the database, even if it contains spaces that would normally be trimmed.

## Test plan
1. Create a user with a password that has a trailing space (e.g., `secret `).
2. Attempt to log in with that exact password.
3. Verify that login succeeds.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/apps/accounts/api_views.py
+++ b/apps/accounts/api_views.py
@@ -16,11 +16,22 @@ from dispatcharr.utils import network_access_allowed
 from .models import User
 from .serializers import UserSerializer, GroupSerializer, PermissionSerializer
 from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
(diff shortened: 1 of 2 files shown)


marcin-spotio:
Closing to recreate from personal GitHub account. The code branches remain available on my fork.

PR #1071, opened 2026-03-06
fix(accounts): allow passwords with leading/trailing spaces

marcinolek:
## Description
Users could get locked out when their password had leading or trailing spaces. Django REST Framework’s `CharField` trims whitespace by default, so the submitted password no longer matched what was stored.

**Changes:**
- **`CustomTokenObtainPairSerializer`** —  sets `trim_whitespace=False` on the login password field (JWT token view).
- **`UserSerializer`** —  sets `trim_whitespace=False` on the password field for create/update, and keeps `required=False` from `dev` after rebase.
Passwords are now handled exactly as sent/stored, including surrounding spaces.
## Related Issue
N/A
## How was it tested?
- Fresh **Debian** installs in **LXC**: one checkout on upstream `dev`, one on this branch.
- Ran **`debian_install.sh`** on both with the same flow.
- Created a user / used a password with **trailing spaces** and attempted login.
- **Before:** login failed; **after (this PR):** login succeeded.
## Checklist
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) (N/A — no frontend changes)
- [ ] Existing tests still pass (not run locally; relying on CI)

### DIFF
--- a/apps/accounts/api_views.py
+++ b/apps/accounts/api_views.py
@@ -16,11 +16,22 @@ from dispatcharr.utils import network_access_allowed
 from .models import User
 from .serializers import UserSerializer, GroupSerializer, PermissionSerializer
 from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
+from rest_framework_simplejwt.serializers import TokenObtainPairSerializer
 
 logger = logging.getLogger(__name__)
 
 
+class CustomTokenObtainPairSerializer(TokenObtainPairSerializer):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Ensure password is not trimmed to allow passwords with leading/trailing spaces
+        if 'password' in self.fields:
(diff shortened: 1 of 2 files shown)


SergeantPanda:
Thank you for the PR!

While this is being reviewed, could you please fix the conflict and fill out or PR form?

SergeantPanda (changes requested):
Sorry it took so long to get back to you on this PR. We wanted to have an internal disscussion on this to see the direction we wanted to go.

We agree with you that there is an inconsistency with how we handle whitespace during account creation and during authentication which can leave accounts in a bad spot.

The direction we'd rather go with this is to actually continue stripping out whitespace on logins and also strip it out during account creation (both management commands and in the create super user frontend form).

If you want to take that on please let me know, otherwise we can take care of it.

Thanks again!

marcinolek:
no problem

Issue #806, opened 2026-01-04
[Feature]: OpenID Connect Support


Yaya48:
### Is your feature request related to a problem?

Hi, would be really great to have OIDC support for panel authentification.

### Describe the solution you'd like

Add OIDC support for panel authentification, with auto users creation and groups matching 

### Additional context

Nothing Extra

SergeantPanda:
Could you explain the use-case for this?

Yaya48:
> Could you explain the use-case for this?

OIDC support would allow:

• Centralized authentication (no local passwords stored in Dispatcharr)
• Automatic user provisioning on first login
• Mapping IdP groups → Dispatcharr roles/permissions
• Easier user management when multiple people use the same instance

Having one account to access all services is very useful in practice. Authentication and authorization are then handled by a dedicated identity provider, which is specifically designed for that purpose.

When running an infrastructure with many self-hosted applications, managing a separate account and password for each application quickly becomes painful and error-prone. This is exactly the problem OIDC is meant to solve.

Adding OIDC support would reduce administrative overhead, improve security (MFA, password policies, access revocation), and better integrate Dispatcharr into modern self-hosted environments.

MotWakorb:
Is your goal with this for administration or for user access for watching channels?

Yaya48:
> Is your goal with this for administration or for user access for watching channels?

The administration panel, yes. I don’t believe the way IPTV applications handle authentication, nor the well-known XC API, will be much updated in the future, and I don’t see the usefulness of this for end users.

jcasimir:
Triage note: This issue appears to be mislabeled. It should have the `question` label removed as it's clearly a feature request (and already has the `Feature Request` label).

ashishraj09:
Please review change https://github.com/Dispatcharr/Dispatcharr/pull/1085

PR #1085, opened 2026-03-10
feat: add OIDC/SSO provider support (#806)

ashishraj09:
- Add OIDCProvider model with full configuration (issuer URL, client ID/secret, scopes, group mapping, button customization)
- Add oidc_provider FK on User model to track which provider authenticated each user
- Add OIDC authorization and callback views (popup + redirect modes)
- Add OidcPopupHandler component for seamless popup-based SSO login
- Add OidcCallback page as fallback for redirect-mode login
- Update LoginForm to show OIDC provider buttons and handle postMessage from popup window
- Update auth store to persist OIDC tokens and clean up all OIDC keys on logout
- Add OidcSettingsForm for managing providers in admin settings, including copyable callback URL display
- Show SSO badge with provider name in UsersTable for OIDC-linked users
- Add OIDC routes to Settings page and frontend router


SergeantPanda:
I am unable to boot this without errors. How was testing done? 

When you created the migration for this, did you use Djangos makemigrations? 

Can you explain your design choices to us please?

CodeBormen:
This was a feature I was looking at implementing down the line and I'm happy to see someone else trying to tackle it. I don't mean to sound critical at all because contributing to the community is what it's all about, but there are some fairly significant security gaps in this implementation that would need to be addressed for this to be useable especially in a potentially exposed environment. 

If you don't mind some suggestions on gaps that I can identify right off the bat, here are some that I see without doing too much of a deep dive:

- OidcPopupHandler.jsx hardcodes localhost:5656 when it should use the existing API pattern from api.js
- No PKCE support which is highly recommended for confidential clients in the OAuth spec
- The client secret is stored in plaintext and should be encrypted
- State tokens have no timestamps and no TTL and have perpetual validity
- State token are never revoked or refreshed
- No testing for any of the implemented logic

Again, not trying to be overly critical just want to see solid implementation of this feature especially since I know there are folks exposing Dispatcharr. If you would like to do any collab, happy to help work on it.

ashishraj09:
> I am unable to boot this without errors. How was testing done?
> 
> When you created the migration for this, did you use Djangos makemigrations?
>
> Can you explain your design choices to us please? 

I’ve attached a design document.
I tested it locally with Authentik, and it works fine without errors.

Yes, I used Django’s makemigrations.

ashishraj09:
> This was a feature I was looking at implementing down the line and I'm happy to see someone else trying to tackle it. I don't mean to sound critical at all because contributing to the community is what it's all about, but there are some fairly significant security gaps in this implementation that would need to be addressed for this to be useable especially in a potentially exposed environment.
> 
> If you don't mind some suggestions on gaps that I can identify right off the bat, here are some that I see without doing too much of a deep dive:
> 
> * OidcPopupHandler.jsx hardcodes localhost:5656 when it should use the existing API pattern from api.js
> * No PKCE support which is highly recommended for confidential clients in the OAuth spec
> * The client secret is stored in plaintext and should be encrypted
> * State tokens have no timestamps and no TTL and have perpetual validity
> * State token are never revoked or refreshed
> * No testing for any of the implemented logic
> 
> Again, not trying to be overly critical just want to see solid implementation of this feature especially since I know there are folks exposing Dispatcharr. If you would like to do any collab, happy to help work on it.

Yes, I have incorporated the changes you suggested. Feel free to collaborate and make changes.

SergeantPanda:
Thank you for the work on this pr.

We have decided to close this and rebuild from the start.

PR #1089, opened 2026-03-11
license update

dekzter:
When Dispatcharr was first released, we licensed it under CC BY-NC-SA. While this was done with good intentions, we later realized that Creative Commons licenses are not designed for software projects.

Because of that, we’re transitioning Dispatcharr to the GNU AGPL v3, which is a license specifically created for open-source software.

This change also makes the license less restrictive in some ways while ensuring the project and its improvements remain open to the community.

Under AGPL, the key principles are:

• Use & Modify – Anyone can use, study, and modify the software.
• Share Source – If you distribute the software or run a modified version as a service, the corresponding source code must be made available.
• Keep It Open – Derivative works must remain under the same AGPL license.

This change aligns Dispatcharr with a license that is commonly used for open-source software and ensures the project remains community-driven while keeping improvements available to everyone.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,437 +1,661 @@
-Attribution-NonCommercial-ShareAlike 4.0 International
-
-=======================================================================
-
-Creative Commons Corporation ("Creative Commons") is not a law firm and
-does not provide legal services or legal advice. Distribution of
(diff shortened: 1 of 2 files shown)


SergeantPanda (approved):
Looks good to me. Thank you!

PR #1091, opened 2026-03-12
fix: Run PostgreSQL as PUID/PGID user to resolve container startup failures

CodeBormen:
## Description

Container startup fails with `chown: changing ownership of '/data/db': Operation not permitted` when `PUID`/`PGID` is set, due to ownership conflicts between the `postgres` system user (UID 102) and the configured PUID/PGID.

Root cause: init scripts hardcoded the `postgres` OS user for all PostgreSQL operations and required `/data/db` ownership by UID 102, while all other `/data` contents used PUID/PGID. This caused two independent failure paths:

1. **UID/GID collision** — If PUID or PGID matched 102 (the postgres package UID), `01-user-setup.sh` renamed the `postgres` user/group, breaking all subsequent `su - postgres` and `chown postgres:postgres` calls.
2. **Filesystem restrictions** — On NFS (`root_squash`), CIFS, or rootless Docker, `chown` to UID 102 fails because the container's root cannot change ownership to an arbitrary UID on the host filesystem.

**Fix:** PostgreSQL now runs as the PUID/PGID user (`$POSTGRES_USER`, default "dispatch") in AIO mode. This eliminates all `chown`-to-UID-102 operations and unifies `/data` ownership under PUID/PGID. Specific changes:

- **`02-postgres.sh`**: All `su - postgres` → `su - $POSTGRES_USER`; all `chown postgres:postgres` → `chown $PUID:$PGID`; added sentinel-based ownership reconciliation to skip redundant recursive `chown` on subsequent startups; added `promote_app_role()` and `ensure_app_database()` for migrating existing clusters where the superuser role is `postgres`.
- **`entrypoint.sh`**: PostgreSQL runtime commands (`pg_ctl`, `pg_isready`) run as `$POSTGRES_USER`; role promotion and database provisioning called after startup.
- **`01-user-setup.sh`**: Auto-detects PUID/PGID from existing `/data/db` ownership when not explicitly set; validates PUID/PGID as positive non-zero integers; blocks UID 0 early.
- **`03-init-dispatcharr.sh`**: Removed redundant `/data/db` ownership check (delegated to `02-postgres.sh` sentinel).

No user-facing configuration changes required. Existing installations where PUID/PGID differs from the current `/data/db` owner are migrated automatically on first startup.

## Related Issue

Closes # 1078

## How was it tested?

- **Automated test suite** (see below)
- **Manual browser verification**: Container started on port 9292, web UI loads successfully, confirmed HTTP 200 with expected HTML content.
- **Docker build + run**: `docker build -t dispatcharr:latest -f docker/Dockerfile .` followed by `docker run` with named volume, bind mount, and various PUID/PGID combinations.

A note on the test suite, since this is a functional test script and not a unit test, it has special functionality:

`docker/tests/test-puid-pgid.sh` provides 20 automated scenarios (131 assertions) that serve as a regression safety net for the container's user/ownership initialization. This area is particularly prone to silent, environment-dependent bugs that only surface in production.

**What it covers:**

- **Fresh installs and upgrades** — Validates both clean initialization and migration from the old `postgres`-owned data layout, including sentinel creation, role promotion, and database provisioning.
- **Edge cases** — UID/GID collisions with system users (e.g., PUID=102), UID 0 rejection, invalid PUID/PGID values, and auto-detection from existing data ownership.
- **Filesystem variations** — Named volumes, bind mounts, and pre-seeded ownership patterns that simulate restricted filesystems (NFS `root_squash`, CIFS).
- **PostgreSQL version upgrades** — End-to-end `pg_upgrade` (PG 16→17) under the PUID/PGID user model.
- **Modular mode** — Confirms init scripts correctly skip internal PostgreSQL management when an external database is configured.
- **End-to-end startup** — Verifies the full stack (PostgreSQL → Django → nginx) responds with HTTP 200.

**Why it matters for future changes:**

- Any modification to `01-user-setup.sh`, `02-postgres.sh`, `03-init-dispatcharr.sh`, or `entrypoint.sh` can be validated against all scenarios, preventing fixes for one edge case from breaking another.
- Base image updates (Alpine/Debian version bumps) that shift system UIDs or change default configurations are caught before they reach users.
- The scenario names and assertions serve as an executable specification of the init system's guarantees, so contributors can understand expected behavior without reading all four shell scripts.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,13 +56,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - PUID/PGID auto-detected from existing data ownership when not explicitly set, avoiding cross-UID `chown` failures on restricted filesystems (NFS `root_squash`, CIFS).
   - PUID/PGID validated as positive non-zero integers before any user/group operations.
   - UID collisions with the `postgres` system user (e.g. PUID=102) are now handled gracefully.
+  - Ensured proper variable quoting in the /docker/ directory to guard from inappropriate input
+
 =======
 - Floating video player bug fixes
   - **Resize stuck after releasing mouse outside window**: The `mouseup` event is not delivered when the pointer leaves the viewport, leaving the `mousemove` listener active indefinitely. Fixed by checking `event.buttons === 0` at the top of `handleResizeMove`; when no button is held the resize session is torn down immediately.
   - **Drag stuck after releasing mouse outside window**: Same root cause as the resize bug. Fixed by detecting `event.buttons === 0` in the `onDrag` handler and dispatching a synthetic `mouseup` event so react-draggable cleanly ends the drag session.
   - **Player draggable off screen**: The player could be dragged off any edge, making the header (and drag handle) unreachable. The player is now fully bounded: left and top edges are clamped to `x ≥ 0` / `y ≥ 0` so the header is always reachable, and right/bottom edges are clamped to the viewport. Size and position are also re-clamped automatically when the browser window is resized, with proportional scale-down if the saved size exceeds the new viewport.
 >>>>>>> dev
-  - Ensured proper variable quoting in the /docker/ directory to guard from inappropriate input
 - **Security**: Any authenticated user could self-escalate privileges by sending `user_level`, `is_staff`, or `is_superuser` in a `PATCH /api/accounts/users/me/` request. The `/me/` endpoint now enforces an allowlist (`custom_properties`, `first_name`, `last_name`, `email`, `password`); any other fields return HTTP 400. Privilege-sensitive fields can only be changed via the admin-only `PATCH /api/accounts/users/{id}/` endpoint.
 - Double error notification when saving user preferences: `API.updateMe` was catching errors internally and displaying a notification before re-throwing, causing callers to display a second notification for the same failure.
 - Navigation preference saves from concurrent sessions could overwrite each other due to a double-merge race: the frontend was pre-merging `custom_properties` before sending, then the backend merged again against the DB value, causing the second session's write to silently drop keys set by the first. The frontend now sends only the delta; the backend merges authoritatively against the stored value.


CodeBormen:
This was a challenging PR to write and implement, but I have tested as many installation/migration scenarios as I am capable. I can verify that on every scenario I tested, which was extensive, it does work.

Most notably, the biggest problem is for users using restricted storage mounts like NFS & SMB/CIFS drivers where root cannot arbitrarily change permissions. Clear logging output is displayed in these cases, and users using these storage types should already be familiar with this limitation. This is not a new issue, NFS & SMB/CIFS mounts have the same problem today, the cause is now made clearer.

I confirmed that this works well across all reasonable deployment methods: new install, upgrade from recent version (users on extremely old versions may experience issues and have to step-upgrade before this fix will work), with both named and bind mounts on both AIO and modular deployments (modular deployments are mostly unaffected to to their nature with external data). Verified on both docker desktop and traditional docker. Debian users do not benefit from this PR since docker is not involved and have a completely different deployment method.

Migrations are automatic and display a progress bar in docker logs, and only need to happen for users changing PUID/PGID or upgrading users who are running as a user that does not match ownership of /data/db. Migrations are idempotent and include rollback capabilities.

docker/init/02-postgres.sh
```diff
@@ -56,6 +163,51 @@ if [[ "$DISPATCHARR_ENV" != "modular" ]]; then
             PG_INSTALLED_BY_SCRIPT=1
         fi
 
+        # Prepare the old cluster for pg_upgrade:
+        # 1. Promote $POSTGRES_USER to superuser (needed for post-upgrade ops)
+        # 2. Detect the bootstrap superuser (install user) — pg_upgrade
+        #    requires -U to match this role exactly.
+        # The old cluster's install user is "postgres" (pre-PUID images)
+        # or $POSTGRES_USER (post-PUID images, future upgrades).
+        _pg_tmp_port=5499
```
SergeantPanda: I wonder if we can just ensure that PG isn't running and use the POSTGRES_PORT? Sorry I am still trying to fully grasp everything that is happening here but I see a potential issue with hardcoding a temp port. (could potentially have conflicts inside a gluetun network for example)

CodeBormen: I do think that $POSTGRES_PORT will work here, I just need to go through and double check that there is no scenario that results in the port being already being bound, but I think since this is being called from entrypoint.sh, that should not be a problem.

CodeBormen:
Good change on the local variables as well. Let me make and test the change to $POSTGRES_PORT and report back. I also need to fix some inconsistent quoting on variables while I'm at it, I really don't think it should be a problem but who knows what people are putting in their env variables.

docker/init/02-postgres.sh
SergeantPanda: Ok that was my gut feeling too is it *should* be safe to do.

SergeantPanda:
> who knows what people are putting in their env variables.

Aint that the truth!

CodeBormen:
The $POSTGRES_PORT swap worked successfully, verified that it is used during the migration and stopped before the entrypoint re-starts postgres so there is no race condition. Only failure path is if the user already has the port bound in the stack, which would have been a failure anyway.

Other than that some minor quoting changes to ensure no bad input mangles the environment.

PR #1097, opened 2026-03-14
CoreSettings.objects.get() error / fix

obnoxiousmods:
Mar 14 16:08:13 lucy dispatcharr[244037]: File "/home/dispatcharr/dispatcharr/dispatcharr/utils.py", line 43, in network_access_allowed Mar 14 16:08:13 lucy dispatcharr[244037]: network_access = CoreSettings.objects.get(key=NETWORK_ACCESS_KEY).value Mar 14 16:08:13 lucy dispatcharr[244037]: 

django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.

EDIT: It seems having the worker class set to gevent is what causes this issue, and having set to gthread fixes it. I may still try to get gevent working tho. The first few locations of this exception occuring are fixed but maybe not all atm.


SergeantPanda:
What?

obnoxiousmods:
Installing this project on my arch linux server results in this exception being thrown in at least 2 different areas. my fork has commits to fix them. 

I am not very familiar with Django, but it complains about " network_access = CoreSettings.objects.get(key=NETWORK_ACCESS_KEY).value" from utils.py being exected from a async environment.  I experienced this issue on Python 3.10 3.13 and 3.14. 

The project is only officially supported for Debian but there is no logical reason for this code not to work on other platforms. 

Your current version of the repo simply just does not work on my server. 

If this isn't the best practice fix for the problem please feel free to implement another.

EDIT: It seems having the worker class set to gevent is what causes this issue, and having set to gthread fixes it. I may still try to get gevent working tho. The first few locations of this exception occuring are fixed but maybe not all atm.

EDIT2: So far it seems my most recent commit fixes support for gevent.

SergeantPanda:
This project only officially supports Docker. There is a way to get it to work with Debian, yes.

But, this PR was not submitted correctly and it was done via AI only it appears. 

The PR should have used our form which requires you to fill out certain information so we don't need to guess what this is for.

I'm not sure what you mean by your edits?

If you want this to be merged, please following the correct steps for a PR and explain in detail what issue you are seeing and how this fixes it.

SergeantPanda:
Please resubmit after reviewing the CONTRIBUTING.MD

PR #1098, opened 2026-03-15
fix: stuck channel states and ghost clients in TS proxy

### LINKED ISSUE #669: [Bug]:  Issue with stuck channels, or ghost channels.
### Describe the bug

Sometimes dispatcharr gets stuck with streams, even without any clients connected.
Sometimes it is necessary to restart dispatcharr for the streams to be finalized.

Image
Image
Image

In Proxy Settings, set:
Buffering Timeout: 5
Buffering Speed: 1
Buffer Chunk TTL: 0
Channel Shutdown Delay: 0
Channel Initialization Grace Period: 0
In Stream Settings -> Default Stream Profile: ffmpeg/Streamlink

I redid the process several times, and this always happens.

### Dispatcharr Version

v0.11.2-20251118160303

### Docker Version

### Client Information

VLC 3.0.20 vetinari

### LINKED ISSUE #695: [Bug]: Streaming of a channel stays in Stats screen if feed unstable
### Describe the bug

When watching a channel, and it encounters an unstable stream (buffering, skipping, etc.), it sometimes remains in the Stats screen despite disconnecting all clients from it.

Even after 10+ minutes pass, you can see the stream appearing/disappearing from the Stats screen as in the photo below. On the provider side, it shows a connection is still active.

If you stay connected to the unstable stream on the client (TiviMate), the stream will play so many seconds, stutter, then repeat the past few seconds, then skip ahead, etc.

The only way to currently remove it from "ghosting" is to either restart the Dispatcharr instance, or try to stream the channel again. If the channel becomes stable upon reconnecting to it, the issue seems to resolve on its own most of the time.

Streaming via Proxy mode in Dispatcharr.

This is occurring on the latest v0.12.0.

1. Stream a channel from a provider that is unstable -- buffering, skips, etc.
2. Stats page in Dispatcharr will show it appearing/disappearing as an "unknown m3u profile" (see photo)
3. Clicking on the red X to disconnect it has no effect.
4. Only way to clear the problem is to either restart the Docker container, or try connecting to the channel again. If the channel is stable, the error may resolve on its own.
Image

### Dispatcharr Version

0.12.0

### Docker Version

Docker version 29.0.1, build eedd969

services:
   dispatcharr:
    # build:
    #   context: .
    #   dockerfile: Dockerfile
    container_name: dispatcharr
    ports:
       - 9191:9191
    volumes:
       - dispatcharr_data:/data
    environment:
       - DISPATCHARR_ENV=aio
       - REDIS_HOST=localhost
       - DISPATCHARR_LOG_LEVEL=info
       - UWSGI_NICE_LEVEL=-5
       - CELERY_NICE_LEVEL=10
    cap_add:
       - SYS_NICE
    devices:
       - /dev/dri:/dev/dri  # For Intel/AMD GPU acceleration (VA-API)
    #deploy:
    #  resources:
    #      reservations:
    #          devices:
    #                count: all
    restart: always

volumes:
  dispatcharr_data:

### Client Information

Device: onn 4K Pro (via ethernet)
App: TiviMate v5.2.0

CodeBormen:
## Description

Fixes two sources of ghost stream issues in the TS proxy:

Stuck channel states: When a stream fails and the ownership TTL (30s) expires during retries, the finally block silently skips the ERROR transition, leaving channels permanently frozen with leaked profile slots. Fixed by adding a state guard fallback: write ERROR when ownership expired and no new owner exists. Also added INITIALIZING to the cleanup task's grace period monitoring as a backstop.

Ghost clients: When a worker dies, client SET entries outlive their metadata hashes. The orphaned channel cleanup saw clients but took no action. Fixed by validating SET entries via pipelined EXISTS, removing ghosts, and cleaning up channels when no real clients remain. Stats page also self-heals on read.

Additional: extracted remove_ghost_clients() helper, added ChannelState.PRE_ACTIVE constant, fixed 6 pre-existing NonOwnerWorkerKeepalive test failures.

## Related Issue

Closes #695
Closes #669

## How was it tested?

Unit tests (42 passing):

16 tests for state guard (owner valid, expired + each state, new owner never clobbered)
7 tests for ghost client detection (detailed/basic stats, orphaned cleanup)
6 fixed NonOwnerWorkerKeepalive tests
13 pre-existing tests

Integration testing (Docker container):

Stuck state — reproduced on unpatched code: 6 dead-link streams + forced ownership deletion. Channel stuck in connecting, profile counter leaked to 1. Patched: state → ERROR, counter → 0.
Ghost clients — reproduced on unpatched code: injected ghost client SET entry with dead owner. Persisted indefinitely, log warned but no action. Patched: ghost removed within one cleanup cycle, channel cleaned up.

## Checklist

- [x] Backend: migrations are included if any models were changed — N/A, no model changes
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema — N/A, no new endpoints
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) — N/A, no frontend changes

### DIFF
--- a/apps/channels/tests/test_ts_proxy_ghost_clients.py
+++ b/apps/channels/tests/test_ts_proxy_ghost_clients.py
@@ -1,15 +1,17 @@
 """Tests for ghost client detection and cleanup.
 
 Covers:
+  - ClientManager.remove_ghost_clients() pipelined EXISTS logic
   - channel_status detailed stats path removes ghost clients from Redis SET
   - channel_status basic stats path removes ghost clients and corrects count
   - _check_orphaned_metadata() validates client SET entries and cleans up
     channels where all clients are ghosts
 """
-from unittest.mock import MagicMock, patch, call
+from unittest.mock import MagicMock, patch, PropertyMock
 
 from django.test import TestCase
 
+from apps.proxy.ts_proxy.client_manager import ClientManager
 from apps.proxy.ts_proxy.constants import ChannelMetadataField, ChannelState
 from apps.proxy.ts_proxy.redis_keys import RedisKeys
 
@@ -17,6 +19,9 @@ from apps.proxy.ts_proxy.redis_keys import RedisKeys
 # ---------------------------------------------------------------------------
 # Helpers
 # ---------------------------------------------------------------------------
+
+CHANNEL_ID = "00000000-0000-0000-0000-000000000001"
+
 
 def _make_proxy_server(redis_client=None):
     """Create a minimal mock ProxyServer with a redis_client."""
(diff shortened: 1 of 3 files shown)


CodeBormen:
Want to be clear about two related issues that this PR does NOT fix:

1) Silent drops from clients (whether poor client behavior or network failure) - this would require TCP keepalive detection, which I could consider looking at in the future

2) Rapid connects/disconnects from clients - when already at a profile's max connection limit, streams that attempt to connect during the teardown period for another channel will continue to fail, citing max connections reached. This is an intentional defensive behavior in the application to prevent momentary periods where the provider max connections could be exceeded. This could potentially be addressed in the future within views.py by adding retry logic to detect stopping keys and wait for the slot to become available. Not in the scope of this PR, but worth noting.

SergeantPanda:
Thank you for this. I did read through the changes and it makes sense to me. I did some testing and didn't find any regressions right away. Seems it is worth throwing in dev and getting some good testing in.

I did make a few (small) changes, hopefully you don't mind!

CodeBormen:
Your changes look good, a couple last minute fixes from me on testing as well as a minor missing constant, but all tests well and is good to go!

SergeantPanda:
Everything looks good. Thank you again for your contributions!

PR #1105, opened 2026-03-15
Fix: audio loss during stream failover and ghost client accumulation

### LINKED ISSUE #1102: [Bug]: Audio loss during stream failover — missing reset_buffer_position() in stream_buffer.py
### Describe the bug

When Dispatcharr switches between stream sources (failover), audio is lost on certain clients while video continues playing. The root cause is a missing `reset_buffer_position()` method in `stream_buffer.py` that `stream_manager.py` already tries to call.

## The Bug

In `stream_manager.py` line ~1105, `update_url()` calls:
```python
if hasattr(self.buffer, 'reset_buffer_position'):
    self.buffer.reset_buffer_position()
```

But `reset_buffer_position()` was never implemented in `StreamBuffer`. The `hasattr` check silently returns `False`, so the internal `_write_buffer` and `_partial_packet` bytearrays are never cleared during failover.

When the new FFmpeg process starts writing, `add_chunk()` concatenates the new data with stale partial data from the old stream:
```python
combined_data = bytearray(self._partial_packet) + bytearray(chunk)
```

This creates corrupted TS packets at the transition boundary. In testing, up to 245KB of stale data was present in the write buffer at each failover, confirmed by adding the missing method and logging what it cleared:
```
Reset buffer position for channel <uuid>: cleared 245152 bytes from write buffer, 48 bytes from partial packet
```

## Client Impact

The corrupted TS packets affect clients differently based on how strict their parsers are:

| Client | Without Fix | With Fix |
|--------|------------|----------|
| VLC | Audio survives (VLC scans forward past corruption) | Audio survives |
| UHF | **Audio lost** (strict parser drops audio decoder on corruption) | Audio survives |

VLC's forgiving parser masked this bug — it recovers by scanning for the next valid TS sync byte. Stricter clients like UHF (and likely Plex native apps, Apple TV, Roku, etc.) drop the audio decoder when they encounter the corrupted packets and never re-initialize it.

## Proposed Fix

Add the following method to `StreamBuffer` in `apps/proxy/ts_proxy/stream_buffer.py`, between `add_chunk()` and `get_chunks()`:
```python
def reset_buffer_position(self):
    """
    Reset internal buffers for a clean stream transition (failover).

    Called by stream_manager.update_url() when switching between FFmpeg
    processes. Without this, _partial_packet from the old FFmpeg gets
    concatenated with the first bytes from the new FFmpeg, creating
    corrupted TS packets that break audio decoder sync in the client.
    """
    try:
        with self.lock:
            old_write_size = len(self._write_buffer)
            old_partial_size = len(getattr(self, '_partial_packet', b''))

            self._write_buffer = bytearray()
            if hasattr(self, '_partial_packet'):
                self._partial_packet = bytearray()

            if old_write_size > 0 or old_partial_size > 0:
                logger.info(
                    f"Reset buffer position for channel {self.channel_id}: "
                    f"cleared {old_write_size} bytes from write buffer, "
                    f"{old_partial_size} bytes from partial packet"
                )
            else:
                logger.debug(
                    f"Reset buffer position for channel {self.channel_id}: "
                    f"buffers were already clean"
                )
    except Exception as e:
        logger.error(
            f"Error resetting buffer position for channel {self.channel_id}: {e}"
        )
```

No changes are needed in `stream_manager.py` — the call site already exists and will automatically find and use this method via the `hasattr` guard.

## Additional Finding: AC3 Audio Codec

During testing, we also discovered that AC3 (Dolby Digital) and EAC3 (Dolby Digital Plus) audio codecs fail to recover after stream discontinuities even with the buffer fix applied. AAC (both stereo and 5.1) recovers correctly. This appears to be a player-side behavior and is separate from this buffer bug, but worth noting for users configuring FFmpeg transcode profiles for failover-heavy setups.

## Test Results Summary

| Scenario | VLC Audio | UHF Audio |
|----------|----------|-----------|
| No fix, AC3 | ❌ Lost | ❌ Lost |
| No fix, AAC | ✅ Works | ❌ Lost |
| **With fix, AAC** | **✅ Works** | **✅ Works** |
| With fix, AC3 | ❌ Lost | ❌ Lost |

1. Configure a channel with a primary and backup stream source using an FFmpeg transcode profile
2. Start playing the channel in a client (Plex, VLC, etc.)
3. Trigger a manual stream switch via the Dispatcharr UI (or wait for an automatic failover)
4. Observe that audio drops while video continues after the switch
5. Exiting and re-entering the channel restores audio, confirming the source stream itself is fine

To verify the buffer contamination:
1. Add the missing `reset_buffer_position()` method to `StreamBuffer` in `apps/proxy/ts_proxy/stream_buffer.py`
2. Trigger a failover and check logs for "Reset buffer position" — non-zero byte counts confirm stale data was present

### Dispatcharr Version

0.20.2

### Docker Version

0.20.2

Not using Docker Compose (Unraid Docker)

### Client Information

Tested with:
- VLC 3.0.21 (macOS/Windows) — audio survives failover both with and without fix (lenient TS parser masks the bug)
- UHF (iOS, KSPlayer-based) — audio lost on failover WITHOUT fix, audio survives WITH fix (strict TS parser exposes the bug)

FFmpeg 8.0.1 with NVIDIA hardware transcoding (h264_nvenc)
Streams tested: MPEG-TS sources with AAC 5.1 and AC3 5.1 audio output

### LINKED ISSUE #1103: [Bug]: Ghost clients accumulate on channels — heartbeat self-refresh defeats ghost detection
### Describe the bug

Clients accumulate on channels without ever disconnecting. In production, a single channel accumulated 9 phantom clients from the same IP over 15 minutes with zero disconnections. The Active Connections UI shows a growing list of stale client entries that never get cleaned up.

## Root Cause

The heartbeat thread in `client_manager.py` has a circular dependency that makes ghost detection impossible.

The heartbeat runs every ~10 seconds and does two things:

**Step 1 — Ghost check** (reads `last_active`):
```python
last_active = self.redis_client.hget(client_key, "last_active")
ghost_timeout = self.heartbeat_interval * GHOST_CLIENT_MULTIPLIER  # ~60s
if current_time - last_active_time > ghost_timeout:
    clients_to_remove.add(client_id)
```

**Step 2 — Heartbeat refresh** (writes `last_active`):
```python
pipe.hset(client_key, "last_active", str(current_time))
pipe.expire(client_key, self.client_ttl)
```

Step 2 unconditionally updates `last_active` for every client still in `self.clients` every 10 seconds. Step 1 checks if `last_active` is older than 60 seconds. **It never will be, because Step 2 keeps refreshing it.** The heartbeat is checking its own output.

When a player disconnects TCP, the stream generator may keep running as a zombie — gevent/OS buffers absorb the yielded data, so `yield chunk` doesn't immediately raise. Since the heartbeat keeps refreshing `last_active`, ghost detection never catches the zombie. The player reconnects with a new `client_id`, and the old one stays forever.

From production logs (channel `3d03b6e8`, FXM):
```
14:35:51  +1 NEW CLIENT  →  clients: 1
14:36:26  +1 NEW CLIENT  →  clients: 2
14:40:02  +1 NEW CLIENT  →  clients: 3
14:41:47  +1 NEW CLIENT  →  clients: 4
14:42:20  +1 NEW CLIENT  →  clients: 5
14:44:00  +1 NEW CLIENT  →  clients: 6
14:45:45  +1 NEW CLIENT  →  clients: 7
14:45:56  +1 NEW CLIENT  →  clients: 8
14:51:16  +1 NEW CLIENT  →  clients: 9
Zero disconnections in 15+ minutes.
```

1. Open a channel in a player (VLC, UHF, Plex, etc.)
2. Disconnect the player (close it, switch channels, or kill the network)
3. Reconnect to the same channel
4. Observe the Active Connections page — the old client entry persists alongside the new one
5. Repeat steps 2-3 several times — client count grows with each reconnect and never decreases
6. Enable debug logging and confirm: no "Disconnected" or "Removed ghost" log entries appear for the old clients

### Dispatcharr Version

0.20.2

### Docker Version

27.5.1, build 9f9e405

Not using Docker Compose (Unraid Docker)

### Client Information

Reproduced with - UHF (iOS, KSPlayer-based)

The bug is server-side in the client tracking layer — it occurs regardless of client type. Any player that disconnects and reconnects will leave a ghost entry. The ghost count correlates with how often the player reconnects, not which player is used.

cmcpherson274:
## Description

This PR fixes two related bugs in the TS proxy that cause client-facing issues during stream operations:

1. **Audio loss during stream failover** — caused by a missing `reset_buffer_position()` method in `StreamBuffer` that `stream_manager.py` already attempts to call
2. **Ghost client accumulation** — caused by the heartbeat thread refreshing `last_active` on all clients unconditionally, defeating its own ghost detection logic

Both bugs are in `apps/proxy/ts_proxy/` and the fixes are minimal: one new method (34 lines), one line removed, one line added.

## Related Issues

Closes #1102 (Audio loss during stream failover)
Closes #1103 (Ghost client accumulation)

### 1. `stream_buffer.py` — Added `reset_buffer_position()`

**The bug:** In `stream_manager.py` line ~1105, `update_url()` calls:
```python
if hasattr(self.buffer, 'reset_buffer_position'):
    self.buffer.reset_buffer_position()
```
This method was never implemented. The `hasattr` guard silently returns `False`, so `_write_buffer` and `_partial_packet` are never cleared during failover. When the new FFmpeg process starts writing, `add_chunk()` concatenates new data with up to 245KB of stale partial TS packet data from the old stream, creating corrupted packets at the transition boundary.

**The fix:** Implements the missing method. It acquires `self.lock` (same lock `add_chunk` uses to prevent race conditions), clears both internal bytearrays, and logs what was cleared. No changes needed in `stream_manager.py` — the existing call site finds it automatically.

### 2. `client_manager.py` — Stopped heartbeat from updating `last_active`

**The bug:** The heartbeat thread runs every ~10 seconds. It checks if `last_active` is older than 60 seconds (ghost detection), then unconditionally writes `last_active = now` for every client. Ghost detection can never trigger because the heartbeat keeps refreshing the value it's checking — it's reading its own output. Zombie clients accumulate indefinitely (observed 9 phantoms on a single channel in production over 15 minutes).

**The fix:** Removed the `pipe.hset(client_key, "last_active", ...)` line from the heartbeat loop. The heartbeat still refreshes Redis key TTLs (preventing expiry) but no longer masks dead clients.

### 3. `stream_generator.py` — Generator now owns `last_active`

**The bug:** With the heartbeat no longer updating `last_active`, something needs to. The stream generator is the correct owner because it's the only component that knows if data is actually reaching the client.

**The fix:** Added `"last_active": str(current_time)` to the existing stats dict that gets written to Redis after every successful `yield chunk`. This piggybacks on an existing `hset` call — no additional Redis round-trip. When a client disconnects and the generator stalls or exits, `last_active` stops updating and ghost detection catches it within ~60 seconds.

## How It Was Tested

Built a patched Docker image and tested on a live Dispatcharr instance (v0.20.2) with NVIDIA hardware transcoding over multiple sessions.

**Audio failover testing:**
- Triggered manual stream switches between primary and backup sources
- Verified buffer contamination via logs — 150-245KB of stale data cleared on each failover:
```
  Reset buffer position for channel <uuid>: cleared 245152 bytes from write buffer, 48 bytes from partial packet
```
- Tested with VLC 3.0.21 and UHF (iOS/KSPlayer)
- Without fix: UHF loses audio on every failover (strict TS parser drops audio decoder on corrupted packets)
- With fix: audio survives failover on both clients

**Ghost client testing:**
- Monitored Active Connections page during player reconnect cycles
- Without fix: client count climbs with every reconnect, never decreases
- With fix: zombie clients are detected and removed by existing ghost detection within ~60 seconds

**Additional finding (not a code change):** AC3/EAC3 audio codecs fail to recover after MPEG-TS discontinuities even with clean buffer data. AAC (stereo and 5.1) recovers correctly. This is player-side behavior, not a Dispatcharr bug, but worth noting for users configuring FFmpeg profiles for failover setups.

## PR Checklist

- [x] I have read the [CONTRIBUTING.md](./blob/dev/CONTRIBUTING.md) in full
- [x] I agree to the [Contributor License Agreement](./blob/dev/CONTRIBUTING.md#contributor-license-agreement)
- [x] Tests are included for new functionality *(manual testing only — documented above)*
- [x] Backend: migrations are included if models changed *(no model changes)*
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema *(no new endpoints)*
- [x] Frontend: ESLint and Prettier pass cleanly *(no frontend changes)*


SergeantPanda:
@CodeBormen since I know you have an active PR open that also touches the proxy, do you mind taking a peak at this to see if it interferes with yours?

Particularly the change around refreshing `last_active` that seems like it could be a culprit to ghost clients to me (without digging to much into it)

CodeBormen:
@SergeantPanda surprisingly, no conflicts - the two PRs are complimentary and should be able to stack on top of each other without issue.

Worth noting - this PR looks like it introduces a race condition between reset_buffer_position and add_chunk on external update_url calls. Could be corrected by extending self.lock in add_chunk to cover the full operation instead of just the redis write. Here would be my suggestion (have not built and tested this change btw):

```python
with self.lock:
    combined_data = bytearray(self._partial_packet) + bytearray(chunk)
    complete_packets_size = (len(combined_data) // self.TS_PACKET_SIZE) * self.TS_PACKET_SIZE
    if complete_packets_size == 0:
        self._partial_packet = combined_data
        return True
    complete_packets = combined_data[:complete_packets_size]
    self._partial_packet = combined_data[complete_packets_size:]
    self._write_buffer.extend(complete_packets)
    # Redis write continues inside same lock
```

Also a fairly significant bug that introduces regression in last_active that would detect a client in failover as a ghost client and inappropriately remove them. My suggestion for that is that when a keepalive is sent, last_active should also be updated the same way a regular client would. This could be achieved in stream_generator.py right before line 315:

```python
if self.proxy_server and self.proxy_server.redis_client:
    client_key = RedisKeys.client_metadata(self.channel_id, self.client_id)
    self.proxy_server.redis_client.hset(client_key, "last_active", str(time.time()))
```
 
  Happy to submit a PR to the branch if necessary also but both are simple enough fixes, just need testing

cmcpherson274:
@CodeBormen  Good catches on both points.

1. **Race condition**: You're right, `_partial_packet` and `_write_buffer` are read/written 
outside the lock in `add_chunk` while `reset_buffer_position` clears them under the lock. 
Extending `self.lock` to cover the full `add_chunk` operation is the correct fix.

2. **Keepalive ghost regression**: Also valid. During failover the chunk gap can be 3-5 seconds 
(confirmed in testing), but longer outages or slow source connections could push past the ghost 
timeout. Updating `last_active` on keepalive send correctly signals the client is still 
connected and waiting.

Happy to incorporate both changes and test on my live instance, or if you'd prefer to submit 
a PR to the branch directly that works too.

CodeBormen:
If you're willing to make the changes, go ahead! Don't want to step on toes, but I am am always happy to help where necessary as well

CodeBormen:
Awesome, I think our changes will work well together

CodeBormen:
I'm testing this out with my changes and seems to work great, if you don't mind I will submit a PR to your branch that will add a max duration cap to the keepalive to prevent clients from being held indefinitely

cmcpherson274:
Tested both changes on a live instance and everything is working great.

1. **Extended lock in `add_chunk`** — applied and tested with multiple back-to-back failovers. Audio and video survive cleanly on both directions (primary → backup and backup → primary). No issues with the longer lock hold time.

2. **Keepalive `last_active` update** — applied and confirmed that clients waiting during failover gaps are no longer at risk of being flagged as ghosts.

Both changes are now in the branch. Deploying to my production server tonight where multiple concurrent users will give the ghost client fix a real workout. Will report back on how it holds up under load.

Thanks for the thorough review; both catches were solid improvements.

CodeBormen:
I did also submit a PR to the branch with the max cap!

cmcpherson274:
@CodeBormen merge complete :) We should be good to go now.

SergeantPanda:
Mind resolving the conflicts on this one? It looks like the test was changed in the other PR. I don't want to make the assumption that the other changes are the correct ones but I suspect that is the case.

SergeantPanda:
Hah, nevermind, maybe I should have actually looked at the conflict, was just a comment...

Changes look good. I extended the ghost client multiplier to ensure it is longer than the timeouts to avoid race conditions.

PR #1125, opened 2026-03-20
fix(vod-proxy): eliminate profile_connections counter stuck-at-nonzero races

### LINKED ISSUE #1121: [Bug]: VOD proxy profile_connections counter can get stuck after PR #983 — three races in multi_worker_connection_manager
## Describe the bug

PR #983 fixed the original counter leak (#962) by adding per-stream `GeneratorExit` handling. However three race conditions in the implementation can still leave `profile_connections:<id>` permanently elevated with no active streams, causing all subsequent VOD requests to return 503 "All profiles at capacity".

**Bug 1 — `decrement_active_streams()` return value ignored (highest impact)**

In `stream_content_with_session()`, all three exit paths (normal, `GeneratorExit`, `Exception`) do:

```python
redis_connection.decrement_active_streams()
decremented = True  # set unconditionally regardless of success
```

`decrement_active_streams()` acquires a Redis lock. If the lock is busy it returns `False` and leaves `active_streams` unchanged — but `decremented = True` is set anyway. The subsequent `has_active_streams()` check then sees `active_streams > 0` and skips `_decrement_profile_connections()`. Counter stuck permanently. The `finally` block won't rescue this because it only acts when `decremented is False`.

**Bug 2 — `_decrement_profile_connections()` uses non-atomic GET-then-DECR**

```python
current_count = int(self.redis_client.get(profile_connections_key) or 0)
if current_count > 0:
    new_count = self.redis_client.decr(profile_connections_key)
```

Two streams finishing simultaneously both GET the same count, both pass the `> 0` guard, both DECR → counter goes to -1. On the next stream start, `_check_and_reserve_profile_slot()` INCRs to 0, which passes `<= max_streams`, so two streams can both get slots on a `max_streams=1` profile. The increment side already uses the correct INCR-first atomic pattern (added in #983) — the decrement side should match it.

**Bug 3 — `has_active_streams()` reads state without a lock**

Called immediately after `decrement_active_streams()` releases its lock. A concurrent stream can increment `active_streams` back in that window, causing the profile counter decrement to be skipped even though the stream finished.

## How can we recreate this bug?

With `max_streams=1` on a VOD profile:

1. Start a VOD stream and disconnect before completion (e.g. client cancel, ffprobe probe)
2. If the Redis lock on `active_streams` is briefly contended at disconnect time, `profile_connections:<id>` stays at 1
3. Check Redis: `redis-cli GET profile_connections:<profile_id>` → returns `1` with no active streams
4. All subsequent VOD requests return 503 "All profiles at capacity"
5. Manual recovery: `redis-cli DEL profile_connections:<profile_id>`

Bugs are in `apps/proxy/vod_proxy/multi_worker_connection_manager.py`. Follow-up to #962 / PR #983.

## Dispatcharr Version

0.21.1

## Docker Version

Running via uWSGI inside Docker container (multi-worker)

## What's in your Docker Compose file?

Standard single-container deployment, Redis on localhost:6379/0

## Client Information

Observed with VLC and Infuse; reproducible with any client that disconnects mid-stream.

firestaerter3:
## Summary

Fixes three race conditions in `multi_worker_connection_manager.py` that could leave the Redis `profile_connections:<id>` counter permanently elevated with no active streams, causing all VOD requests to 503 "All profiles at capacity".

Closes #1121.

### Bug 1 — `decrement_active_streams()` return value was ignored
All three generator exit paths (normal, `GeneratorExit`, `Exception`) called `decrement_active_streams()` and unconditionally set `decremented = True` regardless of whether the lock was acquired. On lock contention the decrement was silently skipped, `active_streams` remained > 0, the subsequent `has_active_streams()` check returned `True`, and `_decrement_profile_connections()` was never called. The counter was then stuck until manual `DEL`.

### Bug 2 — non-atomic GET-then-DECR in `_decrement_profile_connections()`
The previous implementation read the counter with `GET` then conditionally called `DECR`. Two concurrent decrements could both pass the `> 0` guard and both fire, driving the counter to -1. A subsequent `_check_and_reserve_profile_slot()` `INCR` would then produce 0 which passes the `<= max_streams` check, allowing an extra stream to bypass the limit.

### Bug 3 — `has_active_streams()` read state without holding the lock
The separate `has_active_streams()` call after `decrement_active_streams()` released its lock left a window where another concurrent stream could increment `active_streams` back to 1, causing the profile decrement to be skipped.

## Changes

- **New method `decrement_active_streams_and_check()`** — performs the decrement and the "are there remaining streams?" check atomically under a single lock, eliminating the Bug 1 and Bug 3 race windows. All three exit paths and the `finally` block now use this method and act on its `(success, has_remaining)` return values.
- **`_decrement_profile_connections()` rewritten** — replaces GET-then-DECR with a direct `DECR` (matching the `INCR`-first pattern already used by `_check_and_reserve_profile_slot`) and clamps the result to 0 if it goes negative (Bug 2).
- **Tests added** — `apps/proxy/vod_proxy/tests/test_profile_connections.py` with 8 tests covering all three bug scenarios including lock-contention behavior and lock-always-released-on-exception.

## How was it tested?

```
python manage.py test apps.proxy.vod_proxy.tests.test_profile_connections
```

All 8 tests pass. Also deployed to a running instance and confirmed the counter decrements correctly after client disconnect, and is no longer left stuck after lock contention (verified via `redis-cli GET profile_connections:<id>`).

## PR Checklist

- [x] I have read the CONTRIBUTING.md document in full
- [x] I opened a GitHub issue for this change before writing code (see #1121)
- [x] I understand — line by line — every change in this PR
- [x] Backend: no model changes, so no migrations required
- [x] Backend: no new API endpoints, so no OpenAPI schema changes required
- [x] PR targets the `dev` branch

## Contributor License Agreement

- [x] I have read the CLA in CONTRIBUTING.md and agree to its terms

### DIFF
--- a/apps/proxy/vod_proxy/multi_worker_connection_manager.py
+++ b/apps/proxy/vod_proxy/multi_worker_connection_manager.py
@@ -993,7 +993,8 @@ class MultiWorkerVODConnectionManager:
 
             # Create streaming generator
             def stream_generator():
-                decremented = False
+                stream_decremented = False
+                profile_decremented = False
                 stop_signal_detected = False
                 try:
                     logger.info(f"[{client_id}] Worker {self.worker_id} - Starting Redis-backed stream")
@@ -1046,15 +1047,16 @@ class MultiWorkerVODConnectionManager:
                         logger.info(f"[{client_id}] Worker {self.worker_id} - Stream stopped by signal: {bytes_sent} bytes sent")
                     else:
                         logger.info(f"[{client_id}] Worker {self.worker_id} - Redis-backed stream completed: {bytes_sent} bytes sent")
-                    decremented, has_remaining = redis_connection.decrement_active_streams_and_check()
+                    stream_decremented, has_remaining = redis_connection.decrement_active_streams_and_check()
 
                     # Schedule smart cleanup if no active streams after normal completion
-                    if decremented and not has_remaining:
+                    if stream_decremented and not has_remaining and not profile_decremented:
                         # Decrement profile counter immediately — don't defer to daemon thread
                         state = redis_connection._get_connection_state()
                         profile_id = state.m3u_profile_id if state else m3u_profile.id
                         if profile_id:
                             self._decrement_profile_connections(profile_id)
+                            profile_decremented = True
                             logger.info(f"[{client_id}] Profile counter decremented for profile {profile_id} on normal completion")
 
                         def delayed_cleanup():
@@ -1071,18 +1073,19 @@ class MultiWorkerVODConnectionManager:
 
                 except GeneratorExit:
                     logger.info(f"[{client_id}] Worker {self.worker_id} - Client disconnected from Redis-backed stream")
-                    if not decremented:
-                        decremented, has_remaining = redis_connection.decrement_active_streams_and_check()
+                    if not stream_decremented:
+                        stream_decremented, has_remaining = redis_connection.decrement_active_streams_and_check()
                     else:
                         has_remaining = redis_connection.has_active_streams()
 
                     # Schedule smart cleanup if no active streams
-                    if not has_remaining:
+                    if not has_remaining and not profile_decremented:
                         # Decrement profile counter immediately — don't defer to daemon thread
                         state = redis_connection._get_connection_state()
                         profile_id = state.m3u_profile_id if state else m3u_profile.id
                         if profile_id:
                             self._decrement_profile_connections(profile_id)
+                            profile_decremented = True
                             logger.info(f"[{client_id}] Profile counter decremented for profile {profile_id} on client disconnect")
 
                         def delayed_cleanup():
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Proxy changes are frozen atm for an upcoming feature. We will look at this after it has been merged in.

SergeantPanda:
Ok, the change freeze on the proxy is complete for now. If you want to rebase to dev and test your changes again, I'd appreciate it.

SergeantPanda (commented):
fix(vod-proxy): prevent double profile decrement and clean up stream counter naming

- Add profile_decremented flag to stream_generator to guard all four
  _decrement_profile_connections() call sites; prevents a double-decrement
  when an exception cascades through try → except/finally after the profile
  counter was already decremented in the try block

- Rename decremented → stream_decremented for clarity: stream_decremented
  tracks the active_streams counter for this individual stream, while
  profile_decremented tracks the shared per-connection profile slot counter

No functional change to any other path; rollback decrements before the
generator starts are unaffected.

PR #1141, opened 2026-03-25
1140: EPG Record - select channel number explicitly rather than use tvgId

### LINKED ISSUE #1140: [Bug]: Recording from TV Guide when sharing EPG across channels results in first stream being selected
### Describe the bug

When you use the same EPG across different channels, the channel selected to be recorded from the TV Guide is ignored and instead the tvgId is used to find the first channel associated with it.

Although I think this is by design, the behaviour is counter-intuitive. I have different quality streams assigned to their own channels and the first one in the channel list may not be the highest quality.

Create 3 channels all associated with the same EPG:

Channel 1 - BBC SD
Channel 2 - BBD HD
Channel 3 - BBC UHD

Each shares the same EPG from my provider.

If I want to record the UHD stream and select it from the TV Guide, it automatically chooses Channel 1 instead.

I've created my own fork with a commit to fix this - but before I raise a PR, I wanted to check whether this intended functionality is for another purpose, in which case it may be better to put my changes as a configurable setting rather than the default.

Fork: https://github.com/fezster/Dispatcharr
Commit: https://github.com/Dispatcharr/Dispatcharr/commit/fc24908e7870bc009e5eee1a266d6cbac06c6ecf

### Dispatcharr Version

v0.21.1

### Docker Version

Not using

### Client Information

Dispatcharr web ui

fezster:
## Description

When you use the same EPG across different channels, the channel selected to be recorded from the TV Guide is ignored and instead the tvgId is used to find the first channel associated with it. Instead, use the specific channel being selected to be recorded.

## Related Issue

Closes #1140

## How was it tested?

Create 3 channels all associated with the same EPG:

Channel 1 - BBC SD
Channel 2 - BBD HD
Channel 3 - BBC UHD

Now select to record channel 3 and that is the channel that will be recorded under DVR.

## Checklist

- [ x] I have read the [CONTRIBUTING.md](../blob/dev/CONTRIBUTING.md) in full
- [ x] I agree to the [Contributor License Agreement](../blob/dev/CONTRIBUTING.md#contributor-license-agreement)
- [ x] I understand — line by line — every change in this PR and can explain it if asked
- [x ] This PR targets the `dev` branch
- [ x] Tests are included for new functionality
- [ x] Existing tests still pass
- [ x] No `console.log`, `print()`, debug statements, or commented-out code is left in the diff
- [ x] I have not reformatted or refactored code outside the scope of this change


SergeantPanda:
I'm a little concerned on this one that you checked these items off the list for the PR form but didn't actually read it. 

This PR targets the `main` branch but you checked it off that it targets `dev`.

fezster:
> I'm a little concerned on this one that you checked these items off the list for the PR form but didn't actually read it.
> 
> This PR targets the `main` branch but you checked it off that it targets `dev`.

My mistake. I read it as "develop" branch - an oversight. I'll change it.

EDIT - looking at this more closely (and nothing specific I can see in the CONTRIBUTING.md on this), looks like I should be raising my PR to the dev branch.

SergeantPanda:
Thank you!

PR #1157, opened 2026-03-29
perf: use PostgreSQL Unix socket for AIO

JCBird1012:
## Description

Use the PostgreSQL Unix socket for AIO internal database connections instead of TCP to `localhost`. This keeps modular mode unchanged, reduces local connection overhead inside the container (in theory, will there actually be a noticeable performance boost? Maybe for some queries/workloads, but it won't hurt) - I try to use Unix sockets for database connections where possible and I figured Dispatcharr would be a good place to make sure we're using it where we can - the creator of Postgres chats about it [here](https://momjian.us/main/blogs/pgblog/2012.html#June_6_2012).

## How was it tested?
This is a small change - to be honest, I tested it just by relaunching Dispatcharr and ensuring the application still worked while running as `aio`. [02-postgres.sh](https://github.com/Dispatcharr/Dispatcharr/blob/main/docker/init/02-postgres.sh) does a lot of pre-configuring of the Postgres socket for us (in fact, a comment in there mentions running single-app containers like we do with AIO!), this change takes advantage of that. It doesn't impact the ability to override and use Postgres over the network in AIO (though I'm not sure why anyone would do that), and running as modular remains the same.

## Checklist

- [X]  I agree to the [Contributor License Agreement](../blob/dev/CONTRIBUTING.md#contributor-license-agreement)
- [N/A] Backend: migrations are included if any models were changed
- [N/A] Backend: new API endpoints appear correctly in the OpenAPI schema
- [N/A] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`)
- [N/A] Tests are included for new functionality


SergeantPanda:
I will look at this for sure. I will want to do some testing and maybe have someone test that is on a mac (I know their permissions are a bit weird).

There is a potential for issues with permissions but I *think* it *should* be safe.

JCBird1012:
For what it's worth, I tested on a Mac within Docker + Podman Desktop and had no issues. I think we're safe-ish there since the container itself runs inside a Linux VM atop macOS - but yeah, I agree - this should be thoroughly tested. Like I said, thankfully most permissions are preconfigured by [02-postgres.sh](https://github.com/Dispatcharr/Dispatcharr/blob/main/docker/init/02-postgres.sh) - so long as that runs beforehand, things should be ok.

SergeantPanda:
Awesome. I do agree, it is probably going to be OK.

I am going to have to wait until next week sometime to merge (at the earliest) as I will be on vacation. (just so you know I am not deliberately ignoring this!)

SergeantPanda:
Ok. I had time to look this over. It all looks good to me, thank you!

I did make a (small) change to detect AIO users that did still have localhost set (or 127.0.0.1 for that matter) for POSTGRES_HOST to also use the socket connection.

PR #1158, opened 2026-03-30
Html-entity-with-lxml

### LINKED ISSUE #1095: [Bug]: TV guide not working when there are special characters in the EPG
### Describe the bug

If you chose an EPG with special characters (in French for example), the name doesn't show up as it should and the TV guide doesn't work for this channel. It happens on all my channels with them.

Use an EPG list that contains special characters in the name for a particular channel. Choose it and go in the TV guide.

Image

And in the tv guide : 

Image

### Dispatcharr Version

0.20.2

### Docker Version

29.3.0

Not using

### Client Information

Not relevant

SergeantPanda:
HTML named entities in XMLTV EPG files are now correctly preserved during lxml parsing. Some EPG providers (particularly French and other European sources) use HTML named entities like `&eacute;`, `&icirc;`, `&uuml;` in channel names, program titles, and metadata. These are not valid XML entities — lxml 6.0.2 with `recover=True` silently drops them, causing characters to go missing (e.g., "Chaîne Télé" becomes "Chane Tl"). This is now fixed by injecting an XML `<!DOCTYPE tv [...]>` internal subset declaring all 252 HTML 4 named entities directly into the byte stream that lxml reads, using a lightweight in-memory wrapper (`_PrependStream`) with zero disk I/O. libxml2 resolves the entities during its normal C-level parse pass — no Python-level preprocessing or temporary files are involved. The DOCTYPE block (~8 KB) is built once at module load from Python's stdlib `html.entities.name2codepoint` and reused for every parse. Files that already declare their own `<!DOCTYPE>` are passed through unchanged. (Closes #1095)


SergeantPanda:
This appears to have been merged. Github got stuck for some reason.

PR #1165, opened 2026-04-01
Add select_related to Channel queries in output views

### LINKED ISSUE #1127: [Bug]: Xtream Codes get_live_streams endpoint is slow due to N+1 SQL queries (missing select_related)
### Describe the bug

The `xc_get_live_streams` function in `apps/output/views.py` (line ~2169) queries channels without `select_related()`, causing a separate SQL query for each channel's `logo` and `channel_group` foreign keys. For setups with many channels, this results in thousands of individual database queries per request.

**Example**: A user with ~1500 channels reports the `/player_api.php?action=get_live_streams` endpoint takes **~15 seconds** to respond. This is because:

- 1 query to fetch the channel list
- ~1500 queries to lazy-load `channel.logo` (ForeignKey)
- ~1500 queries to lazy-load `channel.channel_group` (ForeignKey)
- Additional `get_or_create` queries for channels without a group

At ~5ms per query, 3000+ queries = ~15 seconds. This causes downstream clients (e.g. Emby tuner plugins) to time out waiting for the response.

1. Have 500+ channels configured in Dispatcharr
2. Call the Xtream Codes API endpoint: `GET /player_api.php?username=<user>&password=<pass>&action=get_live_streams`
3. Measure the response time — it scales linearly with channel count

### Suggested fix

Add `select_related('logo', 'channel_group')` to the channel querysets in `xc_get_live_streams`. This collapses thousands of individual queries into a single SQL JOIN:

```python
# Before (N+1 queries):
channels = Channel.objects.filter(**filters).order_by("channel_number")

# After (single JOIN query):
channels = Channel.objects.filter(**filters).select_related('logo', 'channel_group').order_by("channel_number")
```

This applies to all three queryset paths in the function (unrestricted, profile-filtered, and admin).

Additionally, the `ChannelGroup.objects.get_or_create(name="Default Group")` call inside the stream-building loop could be hoisted outside the loop so it only executes once.

### Dispatcharr Version

Reported on latest (March 2026). The code pattern has been present since the function was introduced.

### Docker Version

N/A — this is a code-level issue, not environment-specific.

N/A — standard Dispatcharr deployment.

### Client Information

Emby with Xtream tuner plugin. The plugin calls `get_live_streams` on startup and every 5 minutes to refresh the channel list. With the current response times, the request times out (default 10s HttpClient timeout) and the guide fails to load.

xBOBxSAGETx:
## Description

`generate_m3u`, `generate_epg`, and `xc_get_live_streams` iterate over all channels and access `channel.logo` and/or `channel.channel_group` on every row. Without `select_related`, each access fires a separate query — producing 2N extra queries for N channels.

This adds `select_related('channel_group', 'logo')` (or just `'logo'` where `channel_group` isn't accessed) to all 14 Channel query paths in these three functions. `xc_get_epg` queries were intentionally left unchanged as they fetch single objects (`.first()` / `get_object_or_404`) where `select_related` provides no benefit.

## Related Issue

Closes #1127

## How was it tested?

**Response times with production data (10,878 channels):**
- `get_live_streams`: **1.1s** (3.8 MB, all 10,878 streams with logos and category IDs populated)
- `generate_m3u`: **1.1s** (2.6 MB)
- `generate_epg` (XMLTV): **56s** cold generation (259 MB)
- `get_live_categories`: **31ms** (183 categories)

**Edge cases tested:**
- Channel with `logo=NULL` and `channel_group=NULL` — `get_live_streams` returns the channel correctly (stream_icon=None, category_id falls back to Default Group via `get_or_create`)
- User with zero profile memberships — returns all 10,878 streams
- Anonymous M3U request (no user, no profile) — 200 OK

## Checklist

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -138,7 +138,7 @@ def generate_m3u(request, profile_name=None, user=None):
                 # Hide adult content if user preference is set
                 if (user.custom_properties or {}).get('hide_adult_content', False):
                     filters["is_adult"] = False
-                channels = Channel.objects.filter(**filters).order_by("channel_number")
+                channels = Channel.objects.filter(**filters).select_related('channel_group', 'logo').order_by("channel_number")
             else:
                 # User has specific limited profiles assigned
                 filters = {
@@ -149,9 +149,9 @@ def generate_m3u(request, profile_name=None, user=None):
                 # Hide adult content if user preference is set
                 if (user.custom_properties or {}).get('hide_adult_content', False):
                     filters["is_adult"] = False
-                channels = Channel.objects.filter(**filters).distinct().order_by("channel_number")
+                channels = Channel.objects.filter(**filters).select_related('channel_group', 'logo').distinct().order_by("channel_number")
         else:
-            channels = Channel.objects.filter(user_level__lte=user.user_level).order_by(
+            channels = Channel.objects.filter(user_level__lte=user.user_level).select_related('channel_group', 'logo').order_by(
                 "channel_number"
             )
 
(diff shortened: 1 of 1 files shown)


SergeantPanda (approved):
This looks good. Thank you!

PR #1166, opened 2026-04-01
Skip stream validation for redirect profiles

xBOBxSAGETx:
## Description

Redirect profiles (HTTP 302) hand the stream URL directly to the client — the client validates it by attempting to play. The server-side `validate_stream_url` probe adds 160-900ms of latency per channel switch with no benefit in this mode, since the server never touches the stream data.

This removes the validation call and alternate-stream failover loop from the redirect path. Stream lock release and non-HTTP protocol handling (RTSP/RTP/UDP) are preserved. Proxy and transcode profile paths are completely untouched.

## Related Issue

N/A — performance improvement for redirect mode users.

## How was it tested?

**Redirect timing:**
- Channel switch via `/live/{user}/{pass}/{id}.ts`: **31ms** to return 302 with correct Location header
- Location header correctly points to the upstream provider URL
- Non-HTTP protocol redirect path preserved (manual `HttpResponse(status=301)` for rtsp/rtp/udp)

**Edge cases tested:**
- Channel with no streams assigned — returns 503 (graceful, upstream `generate_stream_url` handles this before the redirect path)
- Stream with empty URL — upstream code errors before reaching redirect path (not a regression)

## Checklist

### DIFF
--- a/apps/proxy/ts_proxy/views.py
+++ b/apps/proxy/ts_proxy/views.py
@@ -274,93 +274,24 @@ def stream_ts(request, channel_id, user=None):
             # Generate transcode command if needed
             stream_profile = channel.get_stream_profile()
             if stream_profile.is_redirect():
-                # Validate the stream URL before redirecting
-                from .url_utils import (
-                    validate_stream_url,
-                    get_alternate_streams,
-                    get_stream_info_for_switch,
-                )
-
-                # Try initial URL
-                logger.info(f"[{client_id}] Validating redirect URL: {stream_url}")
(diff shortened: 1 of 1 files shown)


SergeantPanda:
I'm not opposed to having this configurable in the proxy or system settings (to skip stream checking on redirect) but there is most definitely a benefit to checking the stream before blindly handing it off to a downstream client.

The code exists for a reason. If the stream is down currently, we can detect that and send a different URL to the client.

In this PR's current state where it just removes the code, I can't merge this. If you'd like to rework it to be user configurable I'd take a look.

PR #1191, opened 2026-04-10
fix: graceful container shutdown - exit 137 on docker stop

Shokkstokk:
## Description

Fixes exit code 137 (SIGKILL) on `docker stop dispatcharr` in AIO mode.

Two bugs in `docker/entrypoint.sh` cause the container to never shut down gracefully:

1. **`set -e` silently aborts cleanup** — The `cleanup()` function uses `kill -0 $pid` to check if processes are alive. Under `set -e`, when a PID no longer exists `kill -0` returns non-zero, causing the entire cleanup function to abort immediately before signalling remaining processes. The `wait` never executes.

2. **Child processes never receive SIGTERM** — `pids[]` only tracks postgres, nginx, and the `su` wrapper PID for uwsgi. uwsgi workers, celery, daphne, and redis are children of that `su` wrapper and are NOT in `pids[]`. When the `su` wrapper receives SIGTERM it kills its shell session but does not forward the signal to its children. They continue running as orphaned processes until Docker force-kills the container.

Fix: add `set +e` at the top of `cleanup()`, and explicitly `pkill -TERM` the untracked processes (uwsgi, celery, daphne, redis-server) before processing the `pids[]` array. A `pkill -KILL` safety net runs after the grace sleep.

## How was it tested?

- Ran `docker stop dispatcharr` repeatedly before and after the fix on a live AIO deployment
- Confirmed exit code 137 (SIGKILL) without fix
- Confirmed exit code 0 with fix
- Verified clean postgres checkpoint shutdown in container logs
- Confirmed `✅ All processes stopped cleanly.` appears in logs on shutdown
- Diagnosed root cause using `docker exec dispatcharr ps aux` to observe orphaned processes during shutdown

## Checklist

- [x] I have read the contributing guidelines in full
- [x] This is a small, obvious bug fix (no issue required per guidelines)
- [x] I understand every line of code in this PR
- [x] No migrations, endpoints, or schema changes
- [x] No new dependencies introduced
- [x] No debug artifacts left in code
- [x] PR targets the `dev` branch

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -4,7 +4,20 @@ set -e  # Exit immediately if a command exits with a non-zero status
 
 # Function to clean up only running processes
 cleanup() {
+    set +e  # Disable exit-on-error so cleanup always runs fully
     echo "🔥 Cleanup triggered! Stopping services..."
+
+    # Explicitly stop uwsgi workers - children of 'su' wrapper, not tracked in pids[]
+    echo "⛔ Stopping uwsgi workers..."
+    pkill -TERM -f uwsgi 2>/dev/null || true
+
+    # Stop celery, daphne, redis - also not tracked in pids[]
+    echo "⛔ Stopping celery, daphne, redis..."
+    pkill -TERM -f "celery" 2>/dev/null || true
+    pkill -TERM -f "daphne" 2>/dev/null || true
+    pkill -TERM -f "redis-server" 2>/dev/null || true
+
+    # Stop tracked processes (postgres, nginx, su/uwsgi wrapper)
     for pid in "${pids[@]}"; do
         if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
             echo "⛔ Stopping process (PID: $pid)..."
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Could you please fill out the form and read through the contributing guidelines?

Shokkstokk:
Updated - let me know if anything else is needed.

SergeantPanda (approved):
Nice fix, this was a real issue. A few things were added on top while reviewing:

Polling loop instead of sleep 3 - exits as soon as all processes are gone rather than always waiting the full 3 seconds. Capped at 8 seconds to stay within Docker's default 10 second stop grace period.

PostgreSQL fallback uses pg_ctl stop -m immediate instead of SIGKILL. Killing postgres with SIGKILL can leave data in a bad state and forces crash recovery on next startup. pg_ctl -m immediate is fast but still exits cleanly.

Double-cleanup guard - the trap and the explicit cleanup call at the bottom of the script were both firing on docker stop. Added a _cleanup_done flag so the body only runs once.

Process name tracking - PIDs are now registered with names at launch so crash diagnostics show postgres, nginx, uwsgi etc. instead of unknown.

Crash diagnostic suppressed on normal stop - the "one of the processes exited unexpectedly" block no longer fires on a clean docker stop, only on genuine crashes.

Empty PID guards - pgrep-based captures for nginx, vite, and postgres now guard pids+= against empty values. This was visible in the original logs as a stray PID: line.

Issue #1196, opened 2026-04-16
[Feature]: Per-channel lock and hide flags for auto-sync'd channels


CodeBormen:
### Is your feature request related to a problem?

When using auto_channel_sync for broad channel management, users cannot selectively customize individual channels without those changes being overwritten on the next M3U refresh.

Specifically:

• Renaming an auto-sync'd channel ("USA AMC West" to "AMC") reverts to the stream-derived name on the next refresh
• Renumbering an auto-sync'd channel to place it in a preferred position reverts on the next refresh
• Deleting an unwanted auto-sync'd channel results in it being recreated on the next refresh because the underlying stream still exists

The workaround in #279 (creating a manual duplicate of the same stream at the preferred channel number) works but creates duplicate entries in the channel list and in downstream clients. For auto-synced lineups with hundreds of channels, duplicates for every customized channel is not scalable.

Currently, auto_channel_sync operates as all-or-nothing at the group level. Users who want a broad auto-managed lineup but also want to hand-customize a subset of those channels are forced to choose between automation and customization, with no middle ground. RegEx filters are another workaround but when providers bundle thousands of channels into their categories, it gets messy quickly.

### Describe the solution you'd like

Two optional boolean fields on the Channel model that let users tell auto-sync to respect their intent on a per-channel basis:

**user_hidden (default: False)**

• When True, the channel is excluded from all outputs (HDHR lineup, M3U output, XMLTV EPG output) so downstream clients never see it
• Auto-sync skips hidden channels entirely (no metadata updates, no deletion, no recreation)
• Replaces "delete" as the removal mechanism for unwanted auto-sync'd channels. The channel record stays in the database so auto-sync recognizes the stream is already handled and doesn't recreate it
• UI action: a "Hide" button alongside the existing Delete action on channels

**user_locked (default: False)**

• When True, auto-sync skips identity updates (name, channel number, channel group) but continues to update provider metadata that should stay current (EPG data, tvg_id, tvc_guide_stationid, logo)
• This means the user's chosen name and number persist, while EPG program data and logos continue flowing from the provider
• Locked channels are also preserved during the stale-channel delete pass (if their stream disappears, the channel persists)
[...]

PR #1204, opened 2026-04-22
feat: user_hidden and user_locked per-channel flags for auto-sync

CodeBormen:
## Description

Adds two new BooleanFields to the Channel model so users can selectively exempt auto-synced channels from refresh overwrites without losing their customizations.

- user_hidden: excludes a channel from HDHR, M3U, EPG, and XC output. Sync continues to refresh its provider metadata (tvg_id, logo, epg_data, tvc_guide_stationid) and never overwrites the flag itself.

- user_locked: preserves the user-chosen name, channel_number, and channel_group on auto-created channels across refreshes while provider metadata continues to flow through. Referred to as "Protected/Unprotected" in frontend to avoid naming collision with the "Unlock for editing" channels table feature.

Both default to false, existing installs see zero behavior change.

Backend changes:

- New fields + migration on Channel; user_hidden is indexed because it filters every downstream output queryset.
- ChannelSerializer exposes both fields (visible in OpenAPI schema).
- sync_auto_channels in apps/m3u/tasks.py gates renumber/rename/group-move on user_locked, and gates stale-delete / empty-group-delete / orphan-cleanup on a shared _channel_is_frozen helper (locked OR hidden).
- HDHR lineup, M3U output, EPG output, and XC (live streams + per-channel EPG + category channels) all apply .exclude(user_hidden=True).
- ChannelViewSet.get_queryset gains a visibility_filter query param (active / hidden / all) scoped to list, get_ids, and summary actions so retrieve/update/delete can still reach a hidden channel by id (needed for unhide).

Frontend changes:

- Row action menu: Hide/Unhide on every row, Protect/Unprotect only on auto_created rows.
- Shield icon on auto-created rows (grey = unprotected, yellow = protected); EyeOff badge on hidden rows, both with explanatory tooltips.
- Bulk edit modal: tri-state Selects for Hide from Clients and Protect from Auto-Sync. Protect is disabled when the selection contains no auto-created rows, and the submit path splits user_locked into a follow-up PATCH targeting only auto-created ids so manual channels never carry a misleading user_locked=true.
- Toolbar visibility filter: Active (default) / Hidden Only / Show All.
- Inline identity edits (name / channel_number / channel_group) on an auto-created channel auto-set user_locked=true in the outgoing PATCH so the customization persists across future refreshes.
- Delete confirmation warns for auto-created rows that the channel will be recreated on the next refresh and recommends Hide instead.

## Related Issue

Fulfills #1196

## How was it tested?

- Django unit tests: 14 tests covering field defaults, serializer round-trip, sync preservation of user_hidden and user_locked across refresh, hidden-channel metadata flow-through, locked channel identity preservation, locked channel survival of stream removal, provider tvg_id updates on locked channel, hidden channel number reservation. All 14 pass against a freshly migrated container.
- Vitest: 2701/2701 pass after updating mocks for the refactored ChannelUtils helpers (selectAutoCreatedInSelection, applyAutoProtect, IDENTITY_FIELDS) and the new user_hidden/user_locked default values.
- Playwright feature spec (not included in this PR): covers model + API basics (field exposure, default values, visibility_filter query param, PATCH round-trip), downstream output exclusion across HDHR / M3U / EPG / XC, sync behavior across all four flag permutations, row action menu for every transition, inline-edit auto-protect, bulk edit modal tri-state flow, bulk Protect filtering to auto-created rows only, visibility filter dropdown, shield / EyeOff indicators, delete confirmation warning grammar, edge case of every channel in a group hidden.
- Significant manual testing and verification

## Checklist

### DIFF
--- a/apps/channels/api_views.py
+++ b/apps/channels/api_views.py
@@ -586,6 +586,7 @@ class ChannelViewSet(viewsets.ModelViewSet):
         show_disabled_param = self.request.query_params.get("show_disabled", None)
         only_streamless = self.request.query_params.get("only_streamless", None)
         only_stale = self.request.query_params.get("only_stale", None)
+        visibility_filter = self.request.query_params.get("visibility_filter", "active")
 
         if channel_profile_id:
             try:
@@ -608,6 +609,15 @@ class ChannelViewSet(viewsets.ModelViewSet):
         if only_stale:
             # Filter channels that have at least one related stream marked as stale
             q_filters &= Q(streams__is_stale=True)
+
+        # Only apply visibility filtering to list-style reads. Retrieve/update/
+        # delete actions must still be able to reach a hidden channel by id so
+        # the frontend can PATCH user_hidden=False to unhide it.
+        if self.action in ("list", "get_ids", "summary"):
+            if visibility_filter == "hidden":
+                q_filters &= Q(user_hidden=True)
+            elif visibility_filter != "all":
+                q_filters &= Q(user_hidden=False)
 
         if self.request.user.user_level < 10:
             filters["user_level__lte"] = self.request.user.user_level
--- a/apps/channels/migrations/0035_channel_user_hidden_channel_user_locked.py
+++ b/apps/channels/migrations/0035_channel_user_hidden_channel_user_locked.py
@@ -1,0 +1,23 @@
+# Generated for FR #1196 (per-channel lock and hide flags for auto-sync'd channels)
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('dispatcharr_channels', '0034_remove_stream_dispatcharr_stream_id_idx_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='channel',
+            name='user_hidden',
+            field=models.BooleanField(db_index=True, default=False, help_text='When True, the channel is excluded from HDHR, M3U, and EPG output and preserved across auto-sync refreshes.'),
+        ),
+        migrations.AddField(
+            model_name='channel',
+            name='user_locked',
+            field=models.BooleanField(default=False, help_text='When True, auto-sync preserves user-set name, number, and group but continues updating EPG data, tvg_id, logo, and tvc_guide_stationid.'),
+        ),
+    ]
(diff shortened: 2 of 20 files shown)


CodeBormen:
A few screenshots of the UX:

image



image




image

CodeBormen:
Hope this aligns with expectations! Happy to take feedback and make enhancements, if needed!

SergeantPanda:
I will play around with this for sure later on to give better feedback on the actual implementation.

I do have a couple of concerns right away thought.

Is this approach an all-or-nothing from an editing standpoint? As in, can the user just lock the name field or do they lock all of those fields you listed if anything is modified? I know some users for example have mentioned they wished they could set a custom EPG and not have it reset every refresh.

Another concern of mine is this is enabling users to use auto channel sync in ways it isn't really intended for (sports and PPV events mainly). 

My thought on something that would be more widely used for non event/sports channels would be to have a completely separate workflow for automatically created channels that would work across all m3u's/groups as a more efficient means to create many channels at once.

Currently, the auto channel sync code isn't the most efficient at handling thousands of channels at once which is why I generally discourage the behavior.

Maybe to satisfy your issues with automation that would be more suitable? 

The other option may be if we wanted more granular control of channels that are auto channel synced would be to have a separate table for auto created channels to store the metadata from anything the user overrides with a priority and pulling from that table first in both the front and the back ends.

Not saying all or even any of that is the best way either. Just off the top of my head.

Like I said, I'll still play with this and see how the flow works.

Thank you for the submission!

CodeBormen:
Yes, you're correct that the way this PR is written is an all-or-nothing on the user lock: if Protected, the channel name, channel number, and channel group are all locked; the remainder of fields remain dynamic. I didn't realize there was much demand for custom EPG - if you decide that is the better route, it wouldn't be too hard to go back and change to per-field locks so that individual fields could be locked. It would mostly be a few new db fields and some frontend changes to expand the Protect across existing fields. Though, that may be moot if the intention would be to have an entirely different flow for auto-created channels.

I wrote this to be fairly minimal with a narrow scope - I think that there is definitely an argument for separate workflows and tables if we can dream up the features to justify the work. That's more of an overhaul/design discussion that I am happy to chat on in DC with feedback from the rest of the team, too.

CodeBormen:
Working on these features as part of an expanded auto-sync feature set, will submit a different PR

PR #1208, opened 2026-04-24
Refactor authentication logic to implement functions hookable by plugins

sethwv:
## Related Issue

Closes #

## Checklist

- [ ] I have read the [CONTRIBUTING.md](../blob/dev/CONTRIBUTING.md) in full
- [ ] I agree to the [Contributor License Agreement](../blob/dev/CONTRIBUTING.md#contributor-license-agreement)
- [ ] I understand — line by line — every change in this PR and can explain it if asked
- [ ] This PR targets the `dev` branch
- [ ] No `console.log`, `print()`, debug statements, or commented-out code is left in the diff
- [ ] I have not reformatted or refactored code outside the scope of this change

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -1959,12 +1959,21 @@ def xc_get_user(request, username=None, password=None):
     try:
         user = User.objects.get(username=username)
         custom_properties = user.custom_properties or {}
-        if custom_properties.get("xc_password") == password:
+        xc_password = custom_properties.get("xc_password", "")
+        # Blank xc_password means XC is disabled for this user
+        if not xc_password:
+            return None
+        if xc_password == password:
             return user
     except User.DoesNotExist:
         user = None
 
     # Plugin authentication hook fallback
+    # Skip if user exists but has blank xc_password (XC disabled)
+    if user is not None:
+        custom_properties = user.custom_properties or {}
+        if not custom_properties.get("xc_password", ""):
+            return None
     try:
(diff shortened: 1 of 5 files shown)


SergeantPanda (commented):
I made a couple minor changes. The check for blank xc_password would have never fired so I removed it (check happens earlier). Added username validation for the xc authentication (basically user returned has to match user request) this is probably not needed to be honest but seemed like it couldn't hurt.

I think it would be good to have this ability turned off by default as a system setting so users have to specifically allow it to function.

It may also be good to include an extra warning when either installing or enabling a plugin that has authentication abilities.

SergeantPanda (commented):
Also, please target the dev branch :)

SergeantPanda (changes requested):
~~Sorry it took so long to get back to you on this PR. We wanted to have an internal disscussion on this to see the direction we wanted to go.~~

~~We agree with you that there is an inconsistency with how we handle whitespace during account creation and during authentication which can leave accounts in a bad spot.~~

~~The direction we'd rather go with this is to actually continue stripping out whitespace on logins and also strip it out during account creation (both management commands and in the create super user frontend form).~~

~~If you want to take that on please let me know, otherwise we can take care of it.~~

~~Thanks again!~~

Edit: I submitted this to the wrong PR... Been a long day...

sethwv:
Will revisit this at a later time after some more thinking about implementation

PR #1214, opened 2026-05-01
restrict XC access per user by IP & CIDR allowlist

sethwv:
## Description

Extends per-user network access controls to cover all access types (M3U/EPG, Streams, XC API, UI). Admins can set IP/CIDR allowlists per user on the API & XC tab. If no ranges are set, the user inherits global network access settings. Requests from IPs outside a user's allowlist are rejected with 401.

The existing `network_access_allowed()` utility is extended with an optional `user` argument rather than adding a new check. Per-user restrictions are stored in `custom_properties['allowed_networks']` alongside existing user properties. No new model fields or migrations are required.

The global Network Access settings form is also updated to use tag-style inputs instead of a plain text field, making it easier to manage multiple CIDRs.

A management command `reset_user_network` is included to clear per-user restrictions if needed.

## How was it tested?

- Created a user with an XC password and set an IP/CIDR allowlist
- Confirmed XC API requests from an allowed IP succeed
- Confirmed requests from a non-allowed IP return 401
- Confirmed leaving the allowlist empty inherits global settings
- Confirmed CIDR and plain IP validation rejects malformed entries
- Confirmed existing global network access settings load correctly after the TagsInput conversion

## Checklist

- [x] Backend: no model changes, no migrations needed
- [ ] ~~Backend: new API endpoints appear correctly in the OpenAPI schema~~
- [ ] ~~Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`)~~
- [ ] ~~Tests are included for new functionality~~


SergeantPanda:
Looking this over and I think if we do a migration for this I'd rather it be more generic of a field. Maybe an "allowed_networks" jsonb field with an XC key under it. then it can be expanded out later if we want more granular control elsewhere.

I *think* I'm ok with the accounts being different for assigning CIDR networks, but I actually think the settings menu should be changed to match what you did as it looks nicer and is easier to read/understand for most users.

apps/output/views.py
```diff
@@ -1926,6 +1926,9 @@ def xc_get_user(request):
     if custom_properties["xc_password"] != password:
         return None
 
+    if not network_access_allowed(request, 'XC_API', user):
```
SergeantPanda, line 1929: Does it make more sense for the xc api's to include the user in the network_access_allowed so it doesn't have to run twice? 

I could see it either way (could move them all into xc_get_user for example)

apps/proxy/ts_proxy/views.py
```diff
@@ -572,6 +572,9 @@ def stream_xc(request, username, password, channel_id):
     if custom_properties["xc_password"] != password:
         return Response({"error": "Invalid credentials"}, status=401)
 
+    if not network_access_allowed(request, 'STREAMS', user):
```
SergeantPanda: Should this return a 403 for consistency? Maybe move it above the password check.

apps/proxy/vod_proxy/views.py
```diff
@@ -1031,6 +1031,9 @@ def stream_xc_movie(request, username, password, stream_id, extension):
     if custom_properties["xc_password"] != password:
         return Response({"error": "Invalid credentials"}, status=401)
 
+    if not network_access_allowed(request, 'STREAMS', user):
```
SergeantPanda: Same questions I have from ts_proxy

apps/proxy/vod_proxy/views.py
```diff
@@ -1065,6 +1068,9 @@ def stream_xc_episode(request, username, password, stream_id, extension):
...
```
SergeantPanda: And here

SergeantPanda (commented):
Overall looks pretty good, I made a few comments if you don't mind looking them over. I'm not saying I'm 100% correct (I didn't put a TON of thought into it). At face value, it seems like it should be changed but I can definitely be persuaded to keep it.

sethwv:
@SergeantPanda I think you're right here, I'm all for sticking to the paradigm. I'll get those staged.

SergeantPanda (commented):
Looks much better, thank you!

I'll play with it after the next release and merge if tests are OK

PR #1215, opened 2026-05-01
Feat/1196 auto sync overhaul

### LINKED ISSUE #1196: [Feature]: Per-channel lock and hide flags for auto-sync'd channels

CodeBormen:
## Description

Comprehensive rebuild of the M3U auto-channel-sync flow to make auto-synced channels safely customizable at scale.

The core addition is a new ChannelOverride table (one-to-one with Channel) that holds user-specified values for name, channel_number, channel_group, logo, tvg_id, tvc_guide_stationid, epg_data, and stream_profile. Sync continues to write only to Channel.* columns; the override row takes precedence at read time via a new with_effective_values() queryset helper that coalesces both sources at the SQL layer. Every output surface (HDHR, M3U, EPG, XC, channels list/detail, TV Guide summary) reads through this helper, so overrides surface consistently.

**New backend models / fields:**

- ChannelOverride (one-to-one Channel) with 8 nullable override columns
- Channel.user_hidden (boolean, indexed): excludes from output, preserved across refreshes
- Channel.channel_number is now nullable
- ChannelGroupM3UAccount.auto_sync_channel_end (float, nullable): inclusive upper bound for auto-sync range
- M3UAccount.auto_cleanup_unused_channels (boolean, default False): off-by-default toggle for the previously-always-on orphan cleanup; hidden channels are always preserved regardless

**New API endpoints:**

- GET /api/m3u/accounts/{id}/auto-created-channels-count/ - count + sample names for the delete-confirmation dialog
- POST /api/m3u/accounts/{id}/repack-group/?channel_group_id=X - manual compact-mode re-pack for a single group
- GET /api/channels/channels/numbers-in-range/?start=X&end=Y - occupants of a channel-number range, used by the live overlap warning
- GET /api/channels/streams/regex-preview/?channel_group=X&find=...&match=...&exclude=...&replace=... - server-side regex preview for the per-group config modal

**New backend modules / helpers:**

- apps/channels/compact_numbering.py (new): compact-mode pack/release/repack helpers used by sync, the post-save signal, and the manual repack endpoint.
- apps/channels/managers.py (new): SQL-level effective-value coalescing helper plus the custom ChannelManager.
- apps/m3u/tasks.py: sync_auto_channels rewritten for bulk operations, override-aware writes, range enforcement, EPG dispatch deduplication, multi-stream safety.
- apps/channels/signals.py: post-save handler triggers compact re-numbering on hide -> unhide transitions.

**Frontend additions:**

- Per-field reset-to-provider icon on every overridable field in the channel edit form. "Provider: <value>" subtext + Undo2 button on auto-synced channels with active overrides.
- Auto-created channels show "Auto-created from: <provider> / <stream>" at the top of the edit form.
- Per-group inline configuration in the M3U Live group filter: each group row has a Sync toggle, a Numbering Mode selector (Provider / Next Available / Fixed), and Start/End channel-number inputs alongside existing fields.
- Per-group "Configure" modal (GroupConfigureModal.jsx) opened from a gear icon on each group row. Holds Channel Sort Order, Compact Numbering toggle + Re-pack now button, Group Override (re-route auto-channels into a different ChannelGroup), Name Regex Find/Replace, Exclude Regex pattern, Force Dummy EPG, and a live regex preview pane with sample matches.
- Live overlap warning across group rows: detects cross-row range overlaps and unexpected occupants in real time as the user edits Start/End. Informational, non-blocking.
- Channels table visibility filter (Active Only / Hidden Only / Show All) and "Has Overrides" filter in the table header.
- Channels table inline indicators: yellow Pencil icon when overrides are active (tooltip lists the overridden field names); EyeOff icon when the channel is hidden.
- Inline cell edits route through the override row for auto-synced rows so edits survive the next sync.
- Bulk hide / unhide in the bulk edit modal, alongside the existing bulk override apply.
- Selection summary in the bulk edit modal: Selection: X auto-synced, Y manual.
- Failure modal grouped by reason (RANGE_EXHAUSTED, INTEGRITY_ERROR, OTHER) in the auto-sync completion notification, with collapsible sections and a per-group cap. The in-memory cap was raised from 50 to 1000 so realistic multi-provider failure sets are not truncated.
- Delete-Playlist preview in the confirmation dialog showing affected channel count.

**Behavior changes to pre-existing surfaces:**

- Channel.channel_number is now nullable. Output endpoints filter out null channel_number rows.
- M3U account delete cascades all auto-created channels owned by the account, regardless of user_hidden or override state; hidden status does not preserve the channel because there is no provider left to populate it on the next sync. Manual channels are unaffected; only the deleted account's streams cascade away. The destroy response now returns {"deleted_channels": N} (HTTP 200) instead of an empty 204.
- Bulk channel edit auto-routes auto-created channels through the override path so edits survive the next sync.
- Inline cell edits on the channels table similarly route through the override row for auto-synced rows.
- Channel-number duplicates are explicitly allowed at the API layer (sync's used_numbers seed avoids same-number collisions during a sync run).

**Bug fix included:**

- Bulk channel edit silently failed on API rejection. The catch block in ChannelBatch.jsx only emitted console.error, so the user saw the spinner stop with no notification and assumed success. Now surfaces a red toast with the server-provided detail and keeps the form open with the in-progress selection intact.

**Performance metrics and enhancements**

The new flow does significantly more work than the previous path; the implementation choices below keep it viable at scale. Measured against the verification fixture (1700 streams, ~6700 channels):

| Operation                                   | Result |
| ------------------------------------------- | ------ |
| Auto-sync cold, 1700 streams                | ~5.5s  |
| Auto-sync cold, 5000 streams                | ~43s   |
| HDHR lineup fetch, 6.7k channels            | ~0.3s  |
| M3U download, 6.7k channels                 | ~0.6s  |
| EPG XML download, 6.7k channels             | ~23s   |
| Channel list, page_size=1000                | ~1.4s  |
| Bulk PATCH, 100 channels                    | ~1.3s  |
| Bulk PATCH, 500 channels                    | ~3.5s  |

Implementation choices (full detail in CHANGELOG.md):

- Single-pass collision detection via a global used_numbers set seeded once per run
- Logo and EPGData lookup caching during sync
- Override-aware bulk PATCH endpoint partitions the selection once, issues two bulk PATCHes

**Migrations**

- apps/channels/migrations/0037_auto_sync_overhaul migration - merged migrations which includes AddField, CreateModel, AddField, RunPython, AlterField, and RunPython to add support for Channel.user_hidden, ChannelOverride, ChannelGroupM3UAccount.auto_sync_channel_end, Channel.channel_number (nullable), and reverse_backfill_channel_number_nulls for rollback
- apps/m3u/migrations/0020_m3uaccount_auto_cleanup_unused_channels.py - adds the off-by-default cleanup toggle

## Related Issue

Closes #1196

## How was it tested?

**Frontend unit tests** New coverage in this PR:

- frontend/src/utils/forms/__tests__/ChannelUtils.test.js: parameterized sentinel battery across all overridable fields; provider hints; override payload building; normalization.
- frontend/src/utils/forms/__tests__/ChannelBatchUtils.bulkOverrideRouting.test.js (new file): bulk override apply logic, auto vs manual partitioning.
- frontend/src/utils/forms/__tests__/GroupSyncUtils.test.js (new file): range reservation detection, cross-row overlap finding, overlap-message formatting.
- frontend/src/components/forms/__tests__/Channel.test.jsx: provider hint rendering, reset-to-provider, override reactivity, form-failure path.
- frontend/src/components/forms/__tests__/ChannelBatch.test.jsx: selection-summary reactivity, bulk override apply, form-failure path.
- frontend/src/components/tables/ChannelsTable/__tests__/EditableCell.test.jsx: inline cell edits route to the override row for auto-synced channels.
- frontend/src/components/__tests__/M3URefreshNotification.test.jsx: failure modal grouping by typed reason, modal lifecycle, summary text.

**Backend unit tests** New coverage in this PR:

- apps/m3u/tests/test_sync_compound.py (new file): compound-fixture sync test seeding multi-stream, hidden, overridden, and manual channels in one fixture, runs sync, asserts each invariant in isolation.
- apps/m3u/tests/test_sync_correctness.py (new file): comprehensive correctness audit covering range enforcement, numbering modes, exclude filter, compact mode, override persistence, auto-cleanup toggle, multi-account scenarios, EPG dispatch, and the 0038 backfill migration.
- apps/m3u/tests/test_account_destroy.py (new file): M3UAccount destroy cascades auto-created channels regardless of hidden or override state, manual channels with provider streams keep their non-provider streams, proxy session cleanup runs outside the DB transaction.
- apps/channels/tests/test_channel_api.py: extended with override CRUD, visibility filter, has-overrides filter, summary endpoint effective-value coverage, and the custom manager wrapper.

**Container deploy and migration round-trip:**

Forward -> rollback -> forward cycle tested against a real upgrade scenario. Built a baseline image from dev, populated it with channels via the bulk M3U fixture, then swapped to this branch's image against the same volume. All 5 new migrations (now merged to a single migration file) applied cleanly; channel count stable. Exercised every new surface (overrides, hidden flags, NULL channel_number, auto_sync_channel_end, auto_cleanup toggle), then rolled back to baseline image: the 0040 reverse hook safely backfilled the NULL channel_number so NOT NULL could be re-imposed; app remained HTTP 200; channel count still 609. Re-applied forward migrations: clean. Channel rows survive the full cycle; FR #1196 customization data (override rows, hidden flags, the cleanup toggle) is dropped on rollback because its columns no longer exist (expected).

**Manual UI walkthrough across all touched surfaces and a local Playwright suite (not committed) for regression coverage.**

## Checklist

- [x] I have not reformatted or refactored code outside the scope of this change - caveat, it's a big change tho


CodeBormen:
Sorry for the massive change, but after our discussion last week this is what was needed to make it viable 😂

Of course, any feedback you have please let me know!

Some screenshots of frontend changes (tried to keep it minimal and sensible):

channel_modal

delete_modal

auto-sync

auto-sync_modal

channel_table

apps/m3u/migrations/0020_m3uaccount_auto_cleanup_unused_channels.py
SergeantPanda, line 1: I think this might be a good candidate to just be in the custom_properties for the M3U account instead of it's own field. Have it only exist if enabled (not default)

This is a change of behavior from what auto channel sync was iirc. Channels used to auto delete if the stream didn't exist anymore. Is there much of a reason to keep a stale channel around?

apps/m3u/tasks.py
```diff
@@ -2323,12 +2493,171 @@ def sync_auto_channels(account_id, scan_start_time=None):
...
+                if rows_to_flip or rows_to_create:
+                    logger.debug(
+                        f"Reconciled memberships for "
+                        f"{len(existing_channel_ids)} channels "
+                        f"({len(rows_to_flip)} flipped, "
+                        f"{len(rows_to_create)} created)"
+                    )
+
+            # Delete channels whose streams have all disappeared.
+            # Hidden channels are preserved so event/PPV holds across
```
SergeantPanda, line 2647: Can you expand on the logic behind not deleting hidden channels a bit more? When I am thinking through this I'd think you'd want to.

frontend/src/components/forms/M3U.jsx
```diff
@@ -68,6 +69,7 @@ const M3U = ({
       stale_stream_days: 7,
       priority: 0,
       enable_vod: false,
+      auto_cleanup_unused_channels: false,
```
SergeantPanda: I feel like maybe this should be where you configure auto channel sync to avoid any sort of confusion.

To be fair, I am just reading through the code at this point and not playing with it yet so I may be misinterpreting it.

apps/m3u/migrations/0020_m3uaccount_auto_cleanup_unused_channels.py
CodeBormen: I don't have a strong argument for not moving it to custom_properties. I think there are maybe tiny tradeoffs to each that in both cases wouldn't really affect the overall app or UX. If you'd prefer it there instead of as a migration, happy to do that.

Regarding behavior change, you're right. The current behavior is that stale channels disappear on next refresh by default. This change flips that behavior, and channels are retained by default unless the toggle is on. There were a few reasons behind that actually, and are mostly based off of other things added by this PR (and not necessarily to keep an empty channel around). With the current behavior, there wouldn't be much reason to carry a channel between refreshes because an auto-sync channel is just a single row populated by M3U sync. With the changes in this PR, an auto-sync'd channel can have ChannelOverride for a number of fields, hidden_from_client (the user_hidden replacement), or channel profile membership set; beyond that, the channel could be tied to a DVR recording rule, or be saved by a downstream client. If the auto-sync silently deletes the channel, all of the customized data is gone along with it. This to me is more of a safety feature to prevent user-customized data from randomly disappearing if their provider has transient stream availability, it's not meant to keep old dead channels around. For user who DO want their channels removed when the streams disappear, they can use the toggle and they will keep the old default behavior.

apps/m3u/tasks.py
CodeBormen: I expanded on this in my other comment!

frontend/src/components/forms/M3U.jsx
CodeBormen: I can look into moving this somewhere else. It only affects auto-sync channels, so I agree that this is not the 100% best placement, but I also did not want the toggle to be per-group since this is a per-account function. If a user who only has manual channels configured toggles this on, they get no functional change.

apps/m3u/migrations/0020_m3uaccount_auto_cleanup_unused_channels.py
SergeantPanda: Understood on your reasoning but I'd push back a little bit on that then. Auto Channel Sync is predominately supposed to be for PPV/Sports channels and how most users should/will use it.

I'm definitely OK with it being expanded for some users who do not care about redundancy/failover streams but I want to keep the core feature for the normal users. Having the ability to keep them if they want them would be my preference. 

At the end of the day, if the stream is gone, there really isn't a reason to keep the channel around. Now, we could have an option for the user to auto cleanup when the stream is deleted vs how it is now where it does not follow stale stream retention.

Out of curiosity, if the stream is deleted from the channel, how would it get rematched to the channel after? Prior to this PR, I used the providers stream id as the source of truth for matching.

tldr: I think the default should be auto cleanup, if for no other reason than for backwards compatibility but I also don't believe users should be using this feature for all of their channels.

frontend/src/components/forms/M3U.jsx
SergeantPanda: Agreed, not per group, probably at the very top for settings that would affect all groups.

SergeantPanda:
Ok. I did my initial read through of the code. I will do a more thorough review later when I have time to play with it.

This is a lot more than I figured it'd be. I want to be very cautious with not impacting users experience (web ui load times, UI bloat) that have no interest in this feature. 

I'll have better feedback after I play with it.

Thank you for the PR!

CodeBormen:
> Ok. I did my initial read through of the code. I will do a more thorough review later when I have time to play with it.
> 
> This is a lot more than I figured it'd be. I want to be very cautious with not impacting users experience (web ui load times, UI bloat) that have no interest in this feature.
> 
> I'll have better feedback after I play with it.
> 
> Thank you for the PR!

Sounds good, I had initially tried to keep the change small with the last PR that was submitted but after discussing last week and with performance concerns being brought up, I had expanded it so that the use case could grow beyond just PPV/Sports so that users who did really want to auto-manage their lineup could do so and still have it be performant.

I was pretty cautious about not affecting any behavior that is not directly related to auto-sync, and a lot of the changes significantly improve performance with auto-sync. I took a lot of performance baselines on the current image and re-measured them after this PR and had good results - essentially they were the same for M3U refresh with no affect to manual channels, and for auto-sync on a provider with 5K streams reduced from ~1+ hour to ~45 seconds.

If you can find any performance related issues that I can look into, happy to do so. Otherwise, I will work on addressing the few other small changes we discussed above!

apps/m3u/migrations/0020_m3uaccount_auto_cleanup_unused_channels.py
CodeBormen: You're correct that a stream would not usually be re-matched. If a stream disappeared and came back with a different stream.id, it would create a new channel. The user would manually have to re-map the stream to the customized channel. It would only be re-matched if the provider brought a stream back with the same stream.id.

Based on what we discussed, would it sound good for me to flip the toggle back to the current default behavior with an option for users to toggle it on to retain customized channels then? I would also move the toggle location like we discussed.

SergeantPanda: Yes that sounds perfect. Thank you.

I assumed that must be the behavior still without reading through all of the code to find out 😂.

Should definitely be able to remove channels that do not have customizations right? Could have an extra option (if you want) to clean up all removed (default), cleanup non modified, and never remove.

SergeantPanda:
I really appreciate you being cautious with it!

Out of curiosity, did you time the initial login before and after the PR?

CodeBormen:
I did not time initial login, however I did time filtering within the channel table along with bulk operations and found no difference. When I go back to make changes, I will also baseline pre- and post-PR login times.

SergeantPanda:
Awesome. I will hold off on checking it out until you are ready. Looking forward to playing with it.

CodeBormen:
Alright, took all that feedback and made the changes.

The auto-sync toggle was moved from the M3U account page to the Groups page, and instead of a binary toggle changed it to a 3-state with the options Always Removed (default - current behavior), Preserve Customized, Never Removed.

Moved the M3UAccountSerializer.update() from migration to custom_properties.

Added a few new tests as well for the new configurations, and updated CHANGELOG.md

Regarding initial login time, I did pre- and post-login measurements and could not measure a meaningful difference. Between the current build and PR build even with 5k channels configured all UI login times were within .5 seconds of each other. The only measurable difference is that because output endpoints now use with_effective_output(), overhead is added to API calls and HDHR/M3U/EPG, but in my testing the difference is in milliseconds and not observably different.

Happy to hear what additional feedback you have on it!

SergeantPanda:
Excited to play with it. Thank you for making those changes.

Sounds promising on the performance too. I will do some tests as well.

SergeantPanda:
Still playing with the code but I am seeing a roughly 69% increase in load times with around 2k channels across 10 tests averaged out.

Test # | Old Code | New Code
-- | -- | --
1 | 4.52 | 7.59
2 | 3.72 | 7.56
3 | 5.17 | 10.35
4 | 4.74 | 7.82
5 | 4.72 | 6.88
6 | 3.84 | 7.58
7 | 6.76 | 6.56
8 | 4.07 | 7.32
9 | 4.63 | 11.62
10 | 5.44 | 6.97
Average | 4.761 Seconds | 8.025 Seconds

SergeantPanda:
I have some test code I'm playing with that brings it down to pretty much what is was before:

Test # | Old Code | New Code | Test Code
-- | -- | -- | --
1 | 4.52 | 7.59 | 6.05
2 | 3.72 | 7.56 | 3.8
3 | 5.17 | 10.35 | 4.42
4 | 4.74 | 7.82 | 3.6
5 | 4.72 | 6.88 | 4.74
6 | 3.84 | 7.58 | 6.94
7 | 6.76 | 6.56 | 4.76
8 | 4.07 | 7.32 | 3.93
9 | 4.63 | 11.62 | 7.36
10 | 5.44 | 6.97 | 3.94
Average | 4.761 Seconds | 8.025 Seconds | 4.954 Seconds

CodeBormen:
Interesting, for load times how do you mean? On the channels table on refresh or from a consumer client perspective? I did not see any regression on load times in my testing but would like to understand how you're testing so that I can repeat that in my environment as well.

By all means though if you are finding better efficiencies that's good

SergeantPanda:
I use our buddy Jesse's ridiculous database for testing, he has 112k streams, 1959 channels, 1850 groups, 1850 logos, ~10 M3U accounts and at least 5 large epgs.

I'll push what I have soon here after I do a bit more testing.

CodeBormen:
Your changes look good and all testing come back fine. I applied the same strategy you used across a couple other places to improve N+1 efficiencies. Still working on my dev env to see if I can reproduce the same level of scrutiny you are getting!

SergeantPanda (changes requested):
Ok, I've been playing with it a bit more and I think its *very* close.

A couple of things that I'd like to see changed:
1) There is no "Save" button or similar at the bottom of the group modal for auto channel sync settings, this may confuse users
2) The search for regex-preview uses "channel_group" name to determine what a match may look like. It is possible that the name may overlap with another M3U account so this should be tied to the actual M3U that is being queried against imo. It may also be worth noting to the user that this is not going to work until AFTER a successful refresh.

CodeBormen:
Both added!

The group modal now has "Done" and "Cancel" buttons at the bottom. Done saves changes to the form state and settings are persisted when the parent modal is saved, Cancel removes changes from the form state. 

The regex preview is now scoped to the M3U account being edited, so other account streams do not show. I did have a message for pre-refresh in there but it was ambiguous, so now the preview will state "No streams in this group yet. Run an M3U refresh first to populate the streams."

SergeantPanda:
I wonder if this needs a rebase now? The changelog is all messed up for the merging.

CodeBormen:
Thanks for taking care of the changelog, I did rebase before I submit the PR but not sure how much got committed in that time. Thanks again!

PR #1232, opened 2026-05-09
fix: plugin event dispatch crashes on first disabled plugin

### LINKED ISSUE #1231: Plugin event dispatch silently aborts on first disabled plugin (AttributeError in connect/utils.py)
## Summary

`apps.connect.utils.trigger_event` iterates `pm.list_plugins()` and on disabled plugins emits a debug log that accesses dict items as if they were attributes, raising `AttributeError`. The exception bubbles out of `trigger_event` and aborts the whole plugin event dispatch loop. Any plugin whose iteration order falls **after** a disabled plugin never receives events, with no error visible at the default INFO log level.

The bug only manifests when the iteration of `pm.list_plugins()` is reached with a disabled plugin — installations with all plugins enabled (or no plugins installed) work fine, which is probably why this hasn't been noticed.

## Reproducer

1. Install any plugin that subscribes to events via `events: [...]` on an action (e.g. [R3XCHRIS/telegram-alerts](https://github.com/R3XCHRIS/telegram-alerts), which subscribes to `channel_start`, `channel_stop`, `channel_reconnect`, `stream_switch`).
2. Have at least one **disabled** plugin sorted before the enabled-with-events one in `pm.list_plugins()` ordering.
3. Trigger any supported event (e.g. start a channel for `channel_start`).
4. The disabled-plugin's `events` action's `run()` is never called, no error appears at INFO log level.

Direct repro from a shell inside the container:

```python
import os, django, logging
os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'dispatcharr.settings')
django.setup()
logging.getLogger('apps.connect.utils').setLevel(logging.DEBUG)
from apps.connect.utils import trigger_event
trigger_event('channel_start', {'channel_name': 'TEST'})
```

## Trace

```
File "/app/apps/connect/utils.py", line 102, in trigger_event
    logger.debug(f"Skipping disabled plugin id={plugin.key} name={plugin.name}")
                                                ^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'key'. Did you mean: 'keys'?
```

(`logger.debug(f"...")` evaluates the f-string eagerly regardless of effective log level — Python doesn't lazy-evaluate the argument to a deferred logger call.)

## Impact

- Any installation with at least one disabled plugin sorted alphabetically (per discovery order) before any enabled plugin that uses `events: [...]` will silently receive zero events for that plugin.
- The plugin's UI test/buttons still work, so the failure looks like the plugin is buggy when in fact dispatch is broken.
- Affects every event type, not just channel events.

Confirmed against current `main` of `Dispatcharr/Dispatcharr` at `apps/connect/utils.py:102`.

## Suggested fix

Replace attribute access with dict access on the same line:

```python
-            logger.debug(f"Skipping disabled plugin id={plugin.key} name={plugin.name}")
+            logger.debug(f"Skipping disabled plugin id={plugin['key']} name={plugin['name']}")
```

Same loop has a related (separate) inefficiency one line below: `logger.debug(json.dumps(plugin))` is also evaluated eagerly per event per enabled plugin even at INFO. Happy to address that in the same PR if wanted, or open a follow-up.

## Environment

- Dispatcharr v0.24.x (confirmed against `main` source as of 2026-05-09).
- Python 3.13 (container default).
- Repro independent of OS / DB backend.

Happy to open a PR against `dev` with the fix + a regression test if this looks right.

R3XCHRIS:
## What

Fixes a two-character bug in `apps/connect/utils.py:102` that silently aborts the plugin event dispatch loop on the first disabled plugin encountered.

## Why

Closes #1231.

`trigger_event` iterates `pm.list_plugins()` and on disabled plugins emits a debug log that accesses dict items as if they were attributes:

```python
logger.debug(f"Skipping disabled plugin id={plugin.key} name={plugin.name}")
```

Python evaluates f-string arguments eagerly even when the logger discards the message at INFO level, so this raises `AttributeError: 'dict' object has no attribute 'key'` the moment a disabled plugin is iterated. The exception bubbles out of `trigger_event` with no `try/except` in the loop, so every plugin sorted after the disabled one silently receives zero events.

The user-visible effect: a plugin subscribed to events via `events: [...]` on an action will never receive them whenever an alphabetically-earlier plugin is disabled. Manual button actions on the same plugin still work, which makes the failure look like a plugin bug rather than a dispatch one.

## How

Replace the two attribute accesses with dict access on the same line:

```diff
- logger.debug(f"Skipping disabled plugin id={plugin.key} name={plugin.name}")
+ logger.debug(f"Skipping disabled plugin id={plugin['key']} name={plugin['name']}")
```

This is the minimum-diff fix. The same loop has a related (but separate) inefficiency one line below — `logger.debug(json.dumps(plugin))` is also evaluated eagerly per event per enabled plugin even at INFO log level. I left that alone to keep this PR focused; happy to follow up.

## How was it tested

Added `apps/connect/tests/test_trigger_event.py` with two cases:

1. **Regression test for this bug.** Mocks `PluginManager` to return `[disabled_plugin, enabled_plugin_with_events_action]` and asserts `pm.run_action` is called with the enabled plugin's `(key, action_id, params)`. This test fails against the original code with the exact `AttributeError` from the trace, and passes with the fix.
2. **Sanity check.** A plugin whose action `events` list doesn't include the fired event (and an action with no `events` key at all) is not dispatched.

Verified inside a running Dispatcharr container with `python manage.py test apps.connect.tests.test_trigger_event`:

- Against current `dev` (buggy): `FAILED (errors=1)` with `AttributeError: 'dict' object has no attribute 'key'`.
- Against this PR: `Ran 2 tests in 0.005s — OK`.

## PR checklist

- [x] I have read CONTRIBUTING.md
- [x] Linked issue: #1231
- [x] I understand every line in this PR
- [x] No new endpoints / no migrations needed
- [x] Frontend: no changes
- [x] Tests included (regression test + sanity check)
- [x] PR targets `dev`
- [x] CLA: I have read and agree to the terms in CONTRIBUTING.md


SergeantPanda:
This looks good. Thank you! 

If you'd like to submit for the other issue you found, you are more than welcome.

PR #1237, opened 2026-05-11
Reduce OCI image bloat by making large optional dependencies actually optional

Matthew-Beckett:
## Description

This pull request introduces optional installation of machine learning (ML) dependencies, making ML features easier to enable or exclude in local, Docker, and Debian-based setups. The main change is moving heavy ML packages (like `torch` and `sentence-transformers`) into an optional dependency group, and updating installation scripts and documentation to support this new workflow.

Dependency management improvements:

* Moved `torch` and `sentence-transformers` from the main `dependencies` list in `pyproject.toml` to a new `[project.optional-dependencies] ml` group, making them optional and only installed when needed. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L30-L31) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R39-R44)

Installer and environment updates:

* Updated Docker build (`docker/DispatcharrBase`) and initialization scripts (`docker/init/99-init-dev.sh`) to accept a `DISPATCHARR_INSTALL_ML` flag. When set to `true`, ML dependencies are installed via the `--extra ml` option to `uv sync`. [[1]](diffhunk://#diff-37c17d7a8c67a652b9fa00cebe2ed520f5c7b6741e119b2a7a1c397f2406acddR9) [[2]](diffhunk://#diff-37c17d7a8c67a652b9fa00cebe2ed520f5c7b6741e119b2a7a1c397f2406acddL31-R34) [[3]](diffhunk://#diff-27106270658cfbe5ba212625df3a9e9061943baad7cbbae4a0b94a9f70fdedb7L19-R24)
* Modified the Debian installation script (`debian_install.sh`) to install ML dependencies only if `DISPATCHARR_INSTALL_ML` is set.

Documentation:

* Updated `CONTRIBUTING.md` to document how to install ML dependencies locally using the `--extra ml` option with `uv sync`.

## Related Issue

A 1.5GB OCI image is beyond suboptimal, it needs no justification that artifacts that excessively large are not shipped

## How was it tested?

Existing code gates against failure to load ML dependencies, I've just exercised our ability for them to never exist instead of forcing them to.

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,6 +81,12 @@ uv run python manage.py migrate
 uv run python manage.py runserver
 ```
 
+If you need ML-enhanced channel matching locally, install the optional ML extra:
+
+```bash
+uv sync --extra ml
+```
+
 ### Frontend
 
 ```bash
--- a/debian_install.sh
+++ b/debian_install.sh
@@ -192,7 +192,11 @@ setup_python_env() {
   env/bin/python -m ensurepip --upgrade
 
   export UV_PROJECT_ENVIRONMENT="$APP_DIR/env"
-  uv sync --no-dev
+  UV_EXTRA_ARGS=""
+  if [ "${DISPATCHARR_INSTALL_ML:-false}" = "true" ]; then
+    UV_EXTRA_ARGS="--extra ml"
+  fi
+  uv sync --no-dev ${UV_EXTRA_ARGS}
 
   env/bin/python -m pip install -q gunicorn
 EOSU
--- a/docker/DispatcharrBase
+++ b/docker/DispatcharrBase
@@ -6,6 +6,7 @@ ENV VIRTUAL_ENV=/dispatcharrpy
 ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 ENV UV_COMPILE_BYTECODE=1
 ENV UV_LINK_MODE=copy
+ARG DISPATCHARR_INSTALL_ML=false
 
 # --- Install Python 3.13 and build dependencies ---
 # Note: Hardware acceleration (VA-API, VDPAU, NVENC) already included in base ffmpeg image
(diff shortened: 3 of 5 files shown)


SergeantPanda:
I am not opposed to having a slim base image but I am definitely opposed to changing the default to not having sentence transformers and torch included in the base image.

This PR does not let docker users use fuzzy matching for epg auto match.

At the end of the day, apps take space, we aren't asking for 20GB of space to run this app, it does quite a bit under the hood and WILL require storage to do so.

If you do not want some of the features, that is understandable but we aren't going to remove features from everyone in the process.

If you would like to refactor this PR a bit to instead of the default being to remove those dependencies to having them be included plus a slim base image that gets built, I would probably approve that.

lareeth:
@SergeantPanda Your image is quite bloated because of it, removing it would be an advantage to users who run on SBCs and mini-pcs with disk limitations. With the increasing cost of storage, 1.5GB for a container image is quite large compared to other projects.

If this is just used for fuzzy matching, I would recommend looking into other projects designed for this task, instead of using an ML library. This is like using a sledgehammer to crack a nut.

This image size reduction would benefit all the users of your project

SergeantPanda:
> @SergeantPanda Your image is quite bloated because of it, removing it would be an advantage to users who run on SBCs and mini-pcs with disk limitations. With the increasing cost of storage, 1.5GB for a container image is quite large compared to other projects.
> 
> If this is just used for fuzzy matching, I would recommend looking into other projects designed for this task, instead of using an ML library. This is like using a sledgehammer to crack a nut.
> 
> This image size reduction would benefit all the users of your project

It's not 'quite bloated'. 

It *may* be large in comparison to other docker images you are used to but that doesn't make it bloated. Everything has a purpose in the image so it's not 'bloat'.

"Storage costs" may be "high" in comparison to the last few years but it is still very affordable to store an extra gigabyte of space so I'm not a fan of that argument.

The long term goal would be to use the LLM for various things and not just how it is being used currently. So while it may seem like a sledgehammer approach, it is there more for future use than anything.

Also, removing it from ALL users because a FEW want a smaller image is not the correct approach. I mentioned I would probably approve a PR that built a slim base for users that can't be bothered to have an extra 700MB - 1GB of space on their device.

SergeantPanda:
Closing due to no response on change suggestions. If you would like to build out a 'slim' docker container please open a new PR.

PR #1239, opened 2026-05-11

cedric-marcoux:
## Summary

Adds a native `/timeshift/{user}/{pass}/{stream_id}/{timestamp}/{duration}.ts` endpoint that proxies Xtream Codes catch-up sessions to clients like iPlayTV (Apple TV) and TiviMate. Promotes the external [`dispatcharr_timeshift`](https://github.com/cedric-marcoux/dispatcharr_timeshift) plugin into first-class core integration with a one-shot migration for existing users.

**Based on v0.24.0.**

## What's new

(full details in `CHANGELOG.md` under `[Unreleased]`)

- **`/timeshift/` endpoint** reuses the same `xc_password` auth as `/live/.../{id}.ts` — no new auth model.
- **`/stats` integration**: catch-up sessions appear next to live sessions with a violet `TIMESHIFT` badge so admins can see active replays at a glance.
- **`xc_get_live_streams` / `xc_get_epg`** now expose catch-up flags (`tv_archive`, `tv_archive_duration`) and emit the provider's `stream_id` **only when the channel has `tv_archive=1`** — non-catchup channels keep `stream_id = channel.id` so existing XC client favorites resolve unchanged. The dual-resolution logic in `stream_xc` handles both forms.
- **New `Settings → Timeshift` panel** with four knobs (defaults are neutral: `UTC`, `en`, off):
  - `default_timezone` (provider local time for `xc_get_epg` JSON + timeshift URL building)
  - `default_language` (XMLTV `<lang>`)
  - `xmltv_prev_days_override` (force a specific lookback window; 0 = auto-detect)
  - `debug_logging` (verbose per-request logs)
- **`/epg.xml` continues to emit UTC times** for Jellyfin / Plex / Channels DVR compatibility — the timeshift timezone setting **does NOT** affect `/epg.xml`. Local-time conversion applies only to `xc_get_epg` JSON.
- **XMLTV `prev_days` auto-detection** from the provider's largest `tv_archive_duration` (capped at 30 days), overridable via setting or `?prev_days=` URL param.
- **Byte path uses a dedicated producer thread + `os.pipe`**, mirroring the pattern in `apps/proxy/ts_proxy/http_streamer.py:HTTPStreamReader` for live streams. Throughput stays comfortably above the typical 5 Mbps FHD bitrate so clients don't buffer.
- **Migration `apps/timeshift/0001`** ports settings forward from the legacy plugin's `PluginConfig` (if present), then marks it `enabled=False` + `migrated_to_native: True`. Idempotent and a no-op on clean installs.

## Architecture

- **`apps/channels/utils.py`** gains two helpers used by 4 sites (xc_get_live_streams, xc_get_epg, stream_xc, timeshift_proxy):
  - `resolve_channel_by_provider_stream_id(provider_stream_id) → (Channel, Stream)`
  - `get_channel_catchup_info(channel) → dict | None`
- **`ChannelMetadataField`** gains `IS_TIMESHIFT`, `LOGO_ID`, `M3U_PROFILE` so the `/stats` card can render the right logo + profile + badge for both live and catch-up. (LOGO_ID is generally useful for live too.)
- **`ProxyServer._check_orphaned_metadata`** skips channel ids starting with `timeshift_` — those are managed by the WSGI request lifecycle, not by a worker heartbeat key.

## Test plan

- [ ] Catch-up playback in iPlayTV (Apple TV) on an XC provider — programme replays without buffering
- [ ] Catch-up playback in TiviMate — same
- [ ] Live `/live/.../{id}.ts` regression: stream_id remap is gated by `tv_archive=1`, so non-catchup XC client favorites still resolve via the legacy internal-id path
- [ ] `/stats` page: catch-up viewer appears with violet `TIMESHIFT` badge, disappears within `CLIENT_RECORD_TTL` (60 s) when client disconnects
- [ ] `/epg.xml` consumed by Jellyfin / Plex / Channels DVR: programme times remain in UTC regardless of the timeshift `default_timezone` setting
- [ ] Settings → Timeshift: round-trip persistence works for all 4 fields
- [ ] Fresh install (no `dispatcharr_timeshift` plugin record): migration `0001` is a no-op, no errors
- [ ] Existing install with the legacy plugin: settings are copied forward, plugin is disabled, `migrated_to_native: True` marker present

## Notes

- A previous prototype of this work used a `subprocess.Popen` to bypass a uWSGI/gevent throughput cap discovered during profiling. The shipped version replaces that with the same `threading.Thread + os.pipe` pattern used by `HTTPStreamReader` for live, keeping all upstream-friendly properties (no hardcoded interpreter path, no `python -c` invocation, no per-request fork) while still measuring 6.4–7.2 Mbps end-to-end via local curl — comfortably above the 5 Mbps FHD bitrate.
- The legacy external plugin remains available at https://github.com/cedric-marcoux/dispatcharr_timeshift for users on older Dispatcharr versions; the migration in this PR is the documented upgrade path for them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **XC catch-up (timeshift) support, built-in**. Adds a native `/timeshift/{user}/{pass}/{stream_id}/{timestamp}/{duration}.ts` endpoint that proxies replay sessions from an Xtream Codes provider to clients such as iPlayTV (Apple TV) and TiviMate. Auth reuses the same `xc_password` custom-property the live `/live/.../{id}.ts` endpoint already uses. The new endpoint integrates with the existing `/stats` page (catch-up sessions appear with a violet `TIMESHIFT` badge alongside live sessions) and respects per-channel access rules.
+  - **`xc_get_live_streams` / `xc_get_epg` now expose catch-up flags** (`tv_archive`, `tv_archive_duration`) and emit the provider's `stream_id` *only when the channel actually has catch-up support* — for channels without catch-up, `stream_id` keeps its existing internal-id value so existing XC client favorites continue to resolve unchanged. The dual-resolution logic in `stream_xc` already handles both forms.
+  - **`xc_get_epg` JSON timestamps and timeshift URLs** are emitted in the provider's local timezone via a new `Settings → Timeshift` panel (timezone, language, debug logging, XMLTV prev_days override). Defaults are neutral: `UTC`, `en`, off. The `/epg.xml` XMLTV endpoint continues to emit UTC times for backward compatibility with Jellyfin / Plex / Channels DVR consumers — the timeshift timezone setting does NOT influence `/epg.xml`.
+  - **EPG XMLTV `prev_days` auto-detection** from the provider's largest `tv_archive_duration` (capped at 30 days), overridable via the new `xmltv_prev_days_override` setting or the `?prev_days=` URL parameter.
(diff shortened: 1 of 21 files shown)


SergeantPanda:
Please target `dev` and fill out the contributing form.

cedric-marcoux:
to dev

PR #1241, opened 2026-05-12

cedric-marcoux:
## Summary

Adds a native `/timeshift/{user}/{pass}/{stream_id}/{timestamp}/{duration}.ts` endpoint that proxies Xtream Codes catch-up sessions to clients like iPlayTV (Apple TV) and TiviMate.

**Based on v0.24.0.** Single self-contained commit (re-opened after a cleanup pass that dropped ~190 lines; supersedes #1239).

## What's new

(full details in `CHANGELOG.md` under `[Unreleased]`)

- **`/timeshift/` endpoint** reuses the same `xc_password` auth as `/live/.../{id}.ts` — no new auth model.
- **`/stats` integration**: catch-up sessions appear next to live sessions with a violet `TIMESHIFT` badge.
- **`xc_get_live_streams` / `xc_get_epg`** now expose catch-up flags (`tv_archive`, `tv_archive_duration`) and emit the provider's `stream_id` **only when the channel has `tv_archive=1`** — non-catchup channels keep `stream_id = channel.id` so existing XC client favorites resolve unchanged. The dual-resolution logic in `stream_xc` handles both forms.
- **New `Settings → Timeshift` panel** with four knobs (defaults: `UTC`, `en`, `0`, off):
  - `default_timezone` (provider local time for `xc_get_epg` JSON + timeshift URL building)
  - `default_language` (XMLTV `<lang>`)
  - `xmltv_prev_days_override` (force a specific lookback window; `0` = auto-detect)
  - `debug_logging` (verbose per-request logs)
- **`/epg.xml` continues to emit UTC times** for Jellyfin / Plex / Channels DVR compatibility. The timeshift timezone setting only affects `xc_get_epg` JSON + the catch-up URL building.
- **XMLTV `prev_days` auto-detected** from the provider's largest `tv_archive_duration` (capped at 30 days), overridable via setting or `?prev_days=`.
- **Byte path uses a dedicated producer thread + `os.pipe`**, mirroring `apps/proxy/ts_proxy/http_streamer.py:HTTPStreamReader` for live streams. Throughput stays comfortably above the typical 5 Mbps FHD bitrate so clients don't buffer.

## Architecture

- **`apps/channels/utils.py`** gains two helpers used by `xc_get_live_streams`, `xc_get_epg`, `stream_xc`, and the new `timeshift_proxy` view:
  - `resolve_channel_by_provider_stream_id(provider_stream_id) → (Channel, Stream)`
  - `get_channel_catchup_info(channel) → dict | None`
- **`ChannelMetadataField`** gains `IS_TIMESHIFT`, `LOGO_ID`, `M3U_PROFILE` so the `/stats` card can render the right logo + profile + badge for both live and catch-up. (`LOGO_ID` is generally useful for live too.)
- **`ProxyServer._check_orphaned_metadata`** skips channel ids starting with `timeshift_` — those are managed by the WSGI request lifecycle, not by a worker heartbeat key.

## Test plan

- [ ] Catch-up playback in iPlayTV (Apple TV) on an XC provider — programme replays without buffering
- [ ] Catch-up playback in TiviMate — same
- [ ] Live `/live/.../{id}.ts` regression: stream_id remap is gated by `tv_archive=1`, so non-catchup XC favorites still resolve via the legacy internal-id path
- [ ] `/stats` page: catch-up viewer appears with violet `TIMESHIFT` badge, disappears within `CLIENT_RECORD_TTL` (60 s) when client disconnects
- [ ] `/epg.xml` consumed by Jellyfin / Plex / Channels DVR: programme times remain in UTC regardless of the timeshift `default_timezone` setting
- [ ] Settings → Timeshift: round-trip persistence works for all 4 fields

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **XC catch-up (timeshift) support**. Adds a native `/timeshift/{user}/{pass}/{stream_id}/{timestamp}/{duration}.ts` endpoint that proxies replay sessions from an Xtream Codes provider to clients such as iPlayTV (Apple TV) and TiviMate. Auth reuses the same `xc_password` custom-property the live `/live/.../{id}.ts` endpoint already uses. Catch-up sessions appear on `/stats` with a violet `TIMESHIFT` badge alongside live sessions and respect per-channel access rules.
+  - **`xc_get_live_streams` / `xc_get_epg` now expose catch-up flags** (`tv_archive`, `tv_archive_duration`) and emit the provider's `stream_id` *only when the channel actually has catch-up support* — for channels without catch-up, `stream_id` keeps its existing internal-id value so existing XC client favorites continue to resolve unchanged. The dual-resolution logic in `stream_xc` already handles both forms.
+  - **`xc_get_epg` JSON timestamps and timeshift URLs** are emitted in the provider's local timezone via a new `Settings → Timeshift` panel (timezone, language, debug logging, XMLTV `prev_days` override). Defaults are neutral: `UTC`, `en`, off. The `/epg.xml` XMLTV endpoint continues to emit UTC times for backward compatibility with Jellyfin / Plex / Channels DVR consumers — the timeshift timezone setting does NOT influence `/epg.xml`.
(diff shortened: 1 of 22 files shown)


SergeantPanda:
It appears you opened a new PR but this one still targets the main branch instead of dev. Also, could you please fill out the PR form?

cedric-marcoux:
Closing in favour of a fresh branch rebased onto `dev` with the upstream `ts_proxy` → `live_proxy` refactor integrated. New PR coming on `dev`.

Issue #569, opened 2025-10-19
[Bug]: Series VOD Internal Error -- more than one M3UEpisodeRelation


daiycman:
### Describe the bug

When trying to watch a series that has multiple sources/Stream IDs dispatcharr throws an internal server error about returning 2 vods

2025-10-19 13:54:26,000 INFO uwsgi.requests Worker ID: 1 GET 500 /series/UN/pass/2926.mkv 9ms
2025-10-19 13:54:26,781 ERROR django.request Internal Server Error: /series/UN/pass/2926.mkv
Traceback (most recent call last):
  File "/dispatcharrpy/lib/python3.13/site-packages/django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "/dispatcharrpy/lib/python3.13/site-packages/django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/app/apps/output/views.py", line 2207, in xc_series_stream
    episode_relation = M3UEpisodeRelation.objects.select_related('episode').get(**filters)
  File "/dispatcharrpy/lib/python3.13/site-packages/django/db/models/query.py", line 636, in get
    raise self.model.MultipleObjectsReturned(
    ...<5 lines>...
    )
apps.vod.models.M3UEpisodeRelation.MultipleObjectsReturned: get() returned more than one M3UEpisodeRelation -- it returned 2!

They all work when using the source directly not through Dispatcharr. Seems to be only impacting Series at this time. couldnt break Movies in the same way

### How can we recreate this bug?

Have VODs turned on for which the source has multiple streams of
Try to play from device
Check Logs

### Dispatcharr Version

 v0.10.4-20251019022013

### Docker Version

Docker version 27.0.3, build 7d4bcd8

### What's in your Docker Compose file?

**Unraid Docker**

docker run
  -d
  --name='Dispatcharr'
  --net='br0'
  --ip='192.168.89.243'
  --pids-limit 2048
  -e TZ="America/Chicago"
  -e HOST_OS="Unraid"
  -e HOST_HOSTNAME="unraid"
  -e HOST_CONTAINERNAME="Dispatcharr"
  -e 'TCP_PORT_9191'='9191'
  -e 'DISPATCHARR_ENV'='aio'
  -e 'REDIS_HOST'='localhost'
  -e 'CELERY_BROKER_URL'='redis://localhost:6379/0'
[...]

PR #1242, opened 2026-05-12

cedric-marcoux:
## Description

Built-in XC catch-up (timeshift) support — proxy the provider's `/streaming/timeshift.php` endpoint so IPTV clients (iPlayTV, TiviMate, iptvnator) can play archived programmes directly through Dispatcharr.

**Key features:**
- Timeshift HTTP proxy view with URL format cascade (5 timestamp shapes) and per-account format cache
- Denormalized catchup fields on Stream and Channel models (zero-cost reads in `_xc_channel_entry`)
- EPG catch-up attributes (`tv_archive`, `tv_archive_duration`, `has_archive`) in player_api and XMLTV output
- Stats integration — timeshift viewers appear on `/stats` with a badge
- Settings UI — Timeshift tab in Settings (timezone selector, debug toggle, XMLTV prev-days override)
- Stream management — timeshift is a first-class citizen in the stream-limits system (same stop-key pattern as VOD)

**Bug fixes since initial submission:**
- Channel resolution: resolve by `Channel.id` first (what the XC API emits), provider stream_id as fallback
- Stream management: timeshift integrated into `check_user_stream_limits()` / `attempt_stream_termination()` — fixes infinite retry loops on `max_connections=1` providers
- Streaming: direct `iter_content`+`yield` instead of thread+pipe (gevent deadlock fix, 74 B/s → 13+ MB/s)
- TS preamble: pre-sync bytes stripped before streaming
- Credential safety: `_redact_url()` truncates to `scheme://host/...`

**All review feedback from previous rounds addressed:**
1. Clients only see Dispatcharr's own IDs, never the provider's stream_id
2. Timeshift endpoint looks up `catchup_provider_stream_id` internally
3. `resolve_channel_by_provider_stream_id` is fallback only
4. Catchup fields fully denormalized
5. No hardcoded user-specific defaults (timezone defaults to UTC)
6. `compute_provider_archive_days_capped` uses `Stream.aggregate(Max)`, no JSON scan

## Related Issue

This implements catch-up/timeshift support, which has been on the roadmap. No single issue number — addresses the general need for XC catch-up streaming through Dispatcharr.

## How was it tested?

- **Unit tests**: 21/21 timeshift backend tests pass (`apps.timeshift.tests.test_helpers`, `apps.timeshift.tests.test_views`)
- **Docker build**: clean build from Dockerfile, image pushed to GHCR
- **Credential scan**: automated scan of all changed files + git diff — clean
- **py_compile**: all 29 changed files compile without warnings
- **Import verification**: all new modules resolve correctly in the container
- **Live testing on iPlayTV (Apple TV)**:
  - Cold start timeshift → plays immediately
  - Live → timeshift transition → live stream terminated via stream limits, timeshift starts within ~1s
  - Timeshift → timeshift switch → previous stream receives stop signal, new stream starts
  - Throughput verified: 13+ MB/s sustained
- **Live testing on TiviMate (Android)**: catch-up playback and seek confirmed working
- **Provider**: 900900.eu with `max_connections=1` — the most restrictive scenario

**Docker image for testing:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

## Checklist


cedric-marcoux:
**Update — additional fixes in progress**

Since the initial PR submission, real-world testing on iPlayTV (Apple TV) and TiviMate has revealed several edge cases that need addressing before this is production-ready. Fixes are implemented and tested locally, will be pushed shortly:

1. **Status code mapping** — upstream 404 now returns client 404 ("Catch-up not available yet") instead of 400, stopping IPTV player retry loops. Upstream 403 → client 403. Fallback cascade widened to try alternative URL on 404 too.

2. **Timestamp format** — XC servers have two timestamp parsers (a strict modern one and a legacy lenient one). The PR now builds 3 candidate URLs per request (Format A dash-shape, Format A SQL-datetime shape, Format B path-style) and tries them in order with a per-account cache so subsequent seeks are single-request.

3. **MPEG-TS preamble stripping** — some providers inject PHP warnings before the TS stream; strict demuxers (ExoPlayer in TiviMate) reject this. The producer thread now scans for the first 0x47 sync chain and strips any preamble.

4. **XC EPG `has_archive` flag** — was hardcoded to 0; now computed dynamically (1 for past programmes within the archive window). Past programmes are also auto-included when the channel supports catch-up, so clients that don't pass `prev_days` still get the catch-up programme list.

5. **XMLTV + JSON timezone conversion** — programme times are now converted to the configured local timezone for both `/xmltv.php` and `get_simple_data_table` responses, matching what XC clients display verbatim.

All changes include unit tests (20 tests total). Updated commits coming soon.

SergeantPanda:
Thank you! I will review this. It will probably take me some time to go through all of the code but I do want to support catchup and it's been on my to-do list. 

This is slightly off-topic but I reviewed your release notes for the timeshift plugin and noticed you had issues with the code in 0.24.0 when I changed the plugins to not load discovery on every uwsgi worker. Please know, that was just an oversight I didn't consider how it could affect plugins that do monkey patching (not really a standard). Anyway, I do *think* I fixed that issue in the latest dev build that I just pushed.

SergeantPanda:
Not sure if you are ready for this to be reviewed yet or not but please resolve the conflicts. I made some changes to the proxy that appear to have interfered a bit. I don't think I'll need to make any more substantial changes for a little while here.

cedric-marcoux:
## Heads up: this PR also includes VOD/series fixes

I initially planned to split timeshift and VOD into separate PRs, but both features modify the same files (`apps/output/views.py`, `apps/proxy/vod_proxy/views.py`), which would create conflicting PRs. Keeping them together avoids that.

**VOD-specific changes** (for review context):

| Change | File | What it fixes |
|--------|------|---------------|
| Null sanitization (`None` → `""`) | `apps/output/views.py` — `xc_get_series_info`, `get_vod_streams`, `get_vod_info`, `get_live_streams` | iPlayTV crashes on JSON `null` for string fields like `custom_sid`, `title` |
| Episode `id` → provider `stream_id` | `apps/output/views.py` — `xc_get_series_info` | Clients build playback URL from `id` field — internal PK doesn't route correctly |
| `stream_id`-first lookup | `apps/output/views.py` — `xc_series_stream` + `apps/proxy/vod_proxy/views.py` — `stream_xc_episode` | Dual lookup: provider stream_id first, internal episode_id fallback |
| Provider cascade in `stream_vod()` | `apps/proxy/vod_proxy/views.py` | When first provider is at capacity, try next by priority — mirrors live TV's `Channel.get_stream()` dispatch logic |

These fixes address upstream issues [#820](https://github.com/Dispatcharr/Dispatcharr/issues/820) and [#569](https://github.com/Dispatcharr/Dispatcharr/issues/569). Happy to split if you prefer — just let me know.

**Docker image for testing:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

Tested on iPlayTV (Apple TV), TiviMate (Android), and iptvnator (Desktop). 16/16 backend tests pass, Docker build clean, zero credentials in code.

SergeantPanda (changes requested):
Thanks for the work here, the catch-up streaming mechanism itself is solid and there are some genuinely good improvements. That said, I have some concerns about the data access patterns and one change I'd like pulled out into a separate PR entirely.

The series/episode ID change should be a separate PR

The xc_get_series_info and stream_xc_episode changes don't need to be here for live TV catchup to work, and bundling them in makes this harder to review and harder to merge cleanly. Could you split those out into their own PR? That way the catchup feature can move forward on its own merits without being blocked on a separate discussion about the episode ID scheme.

The specific concern: M3UEpisodeRelation.unique_together is (m3u_account, stream_id), not globally unique. Two providers can have the same stream_id pointing to completely different content. Switching xc_get_series_info to return the provider stream_id as the episode id and resolving by that in stream_xc_episode means the lookup just picks whichever account has higher priority on a collision, which could silently serve the wrong episode. Dispatcharr's internal episode PK is globally unique. The provider stream_id needed for upstream catch-up URL construction can be looked up at playback time from the M3UEpisodeRelation row that's already being fetched, so clients don't need to carry it at all.

Catchup detection is being computed on every channel list response

get_channel_catchup_info() is called inside _xc_channel_entry() for every channel on every xc_get_live_streams call. Each call fires channel.streams.order_by("channelstream__order") as a live DB query, and adding prefetch_related('streams') on the channel queryset won't fix it because the custom ordering defeats Django's prefetch cache.

The right approach is to denormalize. The M3U/XC importer already iterates every stream and writes custom_properties at refresh time, so adding is_catchup and catchup_days fields to Stream at that same point costs nothing extra. Rolling those up to Channel.is_catchup via a post_save signal on ChannelStream turns the output query into a zero-cost column read. User override is then just a toggle on the field.

resolve_channel_by_provider_stream_id being called unconditionally

In both stream_xc() and xc_get_epg(), this fires before the normal Channel.objects.filter(id=...) lookup on every single request, including channels that have nothing to do with catch-up. It should be a fallback for when the direct ID lookup fails, not a first step.

What's good

The timeshift/views.py proxy logic, auth (hmac.compare_digest), stats registration, and URL routing are all solid. The HTTPStreamReader extensions (1MB pipe buffer, _pipe_write_all() building on the existing O_NONBLOCK, TS preamble stripping, pre-opened response support for the cascade, 206 acceptance) are genuine improvements. The VOD provider cascade in stream_vod() is also a good addition. The main things that need rework before this merges are the catchup eligibility data path and the series ID scheme split.

Happy to discuss further.

SergeantPanda:
One other thing, do you mind filling out the PR form? It seems that was skipped and it is necessary to get the CLA agreement. Thank you.

cedric-marcoux:
Rebased and squashed into 1 commit. Changes from previous review:

**Addressed all 3 review points:**

1. **Series/episode ID changes removed** — `xc_get_series_info`, `xc_series_stream`, `stream_xc_episode`, `stream_vod` are now identical to upstream. VOD cascade split out for separate PR.

2. **Catchup detection denormalized** — `Stream.is_catchup` + `Stream.catchup_days` populated at both XC and M3U import time. Rolled up to `Channel.is_catchup` + `Channel.catchup_days` + `Channel.catchup_provider_stream_id` via `post_save`/`post_delete` signal on ChannelStream (UI path) and explicit SQL after `bulk_create` (import path). Migration 0038 backfills existing data. `_xc_channel_entry()` now reads denormalized fields — zero DB queries per channel. `is_catchup` exposed in `ChannelSerializer` for user override toggle.

3. **resolve_channel_by_provider_stream_id is now fallback only** — both `stream_xc()` and `xc_get_epg()` try `Channel.objects.filter(id=X)` first, provider lookup only when that fails.

**Additional improvements since last push:**
- `ChannelService.stop_channel()` called before timeshift upstream connect — prevents slot contention when switching from live to catch-up on same channel
- Stale comments cleaned up, unused imports removed
- `custom_sid: ""` only in `_xc_channel_entry` (live streams) — VOD endpoints untouched

**Docker image for testing:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

SergeantPanda (changes requested):
Hey, good work on this one. The series/episode ID stuff is gone, catchup fields are properly denormalized on both Stream and Channel, and the streaming path and format cascade logic look solid. A few things before this gets merged though:

One issue is along the same lines as the episode/series ID feedback - get_live_streams is emitting the provider's stream_id for catch-up channels instead of the internal Channel.id. Dispatcharr is a proxy and clients should only ever see Dispatcharr's own IDs. The provider's stream_id is a backend detail that has no business being sent to a client. If a provider changes their stream_ids or a channel moves providers, every client that stored that ID now has a broken reference because Dispatcharr handed out an ID it doesn't own. The fix is actually straightforward since catchup_provider_stream_id is already stored on the Channel - always emit channel.id in get_live_streams for all channels, and have the timeshift endpoint look up channel.catchup_provider_stream_id internally to build the upstream URL. The client never sees it, and the resolve_channel_by_provider_stream_id fallback in stream_xc and xc_get_epg goes away entirely along with the JSON scan it requires.

There's a silent failure in get_programme_duration. datetime.strptime() produces a naive datetime, but with USE_TZ=True the EPG start_time/end_time fields are timezone-aware, so that ORM filter will raise a TypeError at runtime. The broad except Exception catches it silently and always returns the 120-minute default, meaning EPG-based duration detection never actually works. Just add .replace(tzinfo=timezone.utc) to the parsed datetime and import timezone from datetime.

The migration backfill will break on dirty data. (custom_properties->>'tv_archive_duration')::int will throw a Postgres exception the moment it hits any row where that field isn't a clean integer (empty string, "7.5", null, anything). That means the entire migration fails for any user with even one bad row. Needs a safer cast before this ships.

compute_provider_archive_days_capped still iterates the JSON blobs. The whole reason we denormalized catchup_days onto Stream was to avoid exactly this scan. That function should now just be Stream.objects.filter(is_catchup=True, m3u_account__account_type="XC").aggregate(Max("catchup_days"))["val"] - one query, no JSON parsing, no loop.

get_channel_catchup_info hits the DB on every timeshift request. It walks channel.streams.order_by(...) when the channel already has catchup_provider_stream_id sitting right on it. Should be a direct stream lookup using that field instead of iterating.

A few smaller things: _convert_xmltv_to_local_timezone in output/views.py runs the full regex rewrite even when the timezone is already UTC, so an early return there would be a quick win. And from datetime import timezone as dt_timezone ends up inside the function body rather than at the top of the file with the rest of the imports.

One suggestion on _format_cache: since helpers.py in this same PR already uses django.core.cache with a get/set pattern, it would be worth doing the same here instead of the module-level dict. The lock goes away (Redis ops are atomic), and you get cross-worker sharing for free - right now every uWSGI worker has to rediscover the winning URL format independently on its first request. Small change for a real gain.

SergeantPanda:
Also, please fill out the PR form to give us permissions to use the code.

cedric-marcoux:
## Update: bug fixes and improvements

Pushed a new commit addressing issues found during real-world testing with iPlayTV on Apple TV:

### Bug fixes

1. **Channel resolution fix** — The timeshift view was calling `resolve_channel_by_provider_stream_id()` with `Channel.id` (what the XC API emits), which searches the wrong ID namespace. Now resolves by `Channel.id` first, provider stream_id as fallback.

2. **Stream management integration** — Timeshift is now a first-class citizen in the stream-limits system (same pattern as VOD):
   - `get_user_active_connections()` detects `type='timeshift'`
   - `attempt_stream_termination()` uses `RedisKeys.client_stop()` for timeshift
   - `timeshift_proxy()` calls `check_user_stream_limits()` before connecting upstream
   - `stream_generator()` checks the stop key every 100 chunks
   
   This fixes infinite retry loops on `max_connections=1` providers when switching live→timeshift or timeshift→timeshift.

3. **Streaming throughput** — Replaced `HTTPStreamReader` thread+pipe with direct `iter_content`+`yield` (same pattern as VOD proxy). The pipe approach deadlocked under gevent because the reader thread's `select()` competed with the greenlet's pipe read for hub scheduling. Throughput: ~74 B/s → 13+ MB/s.

4. **TS preamble stripping** — `find_ts_sync()` returns an offset that could be > 0, but the peek data was prepended without stripping pre-sync bytes. Fixed: `response._peek_data = peek[sync_offset:]`.

5. **Credential safety** — `_redact_url()` now truncates to `scheme://host/...` instead of keeping path segments (which contain XC credentials in format B URLs).

### Code cleanup
- Removed dead `import os` and `HTTPStreamReader` import
- Updated stale comments referencing the removed pipe approach
- Added `return False` guard in timeshift termination when Redis unavailable
- Protected `upstream.close()` in finally block
- Final heartbeat bytes flush before unregistering stats
- Fixed migration regex `SyntaxWarning`
- Updated 5 tests that referenced the removed `HTTPStreamReader`

### Test results
- 21/21 timeshift backend tests pass
- Docker build clean
- Credential scan: clean
- Live tested on iPlayTV (Apple TV)

**Docker image for testing:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

SergeantPanda (changes requested):
Thank you for filling out the PR form.

I feel this one is getting closer to going through more rigorous testing.

A couple things I can see still that I'd like resolved before moving forward though:
1) the timeshift_proxy still has a fallback to resolve_channel_by_provider_stream_id. I don't think we should be doing that at all, the client should not ever even know what the provider stream id is so I don't see the use-case for it (unless I'm missing something here, if so, please let me know)
2) This one is extremely minor but in views.py, _get_cached_format_index and _set_cached_format_index both import django.core.cache and that can be cleaned up a bit by importing at the top of the file.

ssjgesus:
let me know if I should be providing testing notes elsewhere but:

- Doesn't seem to reliably mark channels with tv_archive: 1 even when they contain streams which do have tv_archive: 1. 
Potentially it only marks them on initial migration from previous version, but not when new channels are created. Is there a way to retrigger the marking process?
It would also be a huge help with we could see tv_archive and tv_archive_duration in the streams and channel views as extra columns.

cedric-marcoux:
Thanks @SergeantPanda — addressed both remaining items from your latest review, plus a resync with `dev` and a small cleanup.

### 1. Removed the `resolve_channel_by_provider_stream_id` fallback in `timeshift_proxy()`
Agreed — the client should never see the provider stream_id. The catch-up endpoint now resolves strictly by Dispatcharr's internal `Channel.id` and returns `404` on miss. I also:
- dropped the dead `channel = None` sentinel and the now-unused import, and
- **deleted the now-orphaned `resolve_channel_by_provider_stream_id` helper** (and its local `Stream` import) from `apps/channels/utils.py` — it had zero remaining call sites.

The provider stream_id is looked up internally via `get_channel_catchup_info()`; the client only ever sends `Channel.id`.

### 2. Moved `from django.core.cache import cache` to module top-level
Out of `_get_cached_format_index` / `_set_cached_format_index` in `apps/timeshift/views.py`.

### Also in this push
- Fixed the stale comment in `dispatcharr/urls.py` (the `duration` slot carries `Channel.id`, not the provider stream_id).
- **Resynced with `dev`** (resolved 2 trivial conflicts: an import union in `apps/output/views.py`, one `booleanFields` entry in `SettingsUtils.js`) — the PR is mergeable again.
- Added migration `0039` to sync the `catchup_provider_stream_id` `help_text` (no schema change) so `makemigrations --check` is clean.
- Genericised a couple of provider-host references in code comments.

### Verification
- `apps.timeshift.tests` — **21/21 pass**.
- Rebuilt the Docker image and live-retested end-to-end through Dispatcharr's own endpoints: `404` on unknown id, `403` on bad creds / no access, and a real catch-up stream returns `200` with valid MPEG-TS (single upstream request).

**Updated test image:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

SergeantPanda (changes requested):
Ok I went ahead and played with this a little bit and make a few changes. 

I refactored how channels are marked as catchup, this is now its own function and runs after every m3u refresh.

I removed the catchup_provider_stream_id field as I don't see a need for that.

I also have it grabbing the catchup stream ordered by channelstream order.

Unfortunately I still can't get this to work. I am getting 500 responses from both providers I've tested with and don't have more time to diagnose what is going on.

SergeantPanda (changes requested):
Another thing I just thought of, it seems there are timeshift settings that maybe aren't necessary. Providers output what their timezone is and I would imagine we could use that to determine what the catchup timezone shoudl be instead of forcing one for all different account?

cedric-marcoux:
@SergeantPanda tracked down the 500s — root cause + fix pushed (kept your 3 commits on top).

### Why it 500s on your providers but streamed fine in my own testing
The candidate-format cascade in `_stream_from_provider` treated any upstream **5xx as a decisive failure and stopped**, never trying the remaining timestamp shapes. The catch: some XC servers run PHP with `display_errors` **off**, so a timestamp shape their parser can't handle comes back as a hard **HTTP 500** (a PHP "Undefined array key" fatal) instead of a `200` with the warning text inline in the body.

- On servers where it's a `200`+text → the cascade's "no TS sync → soft-reject → try next shape" path kicks in and the next shape succeeds.
- On servers that return `500` → the cascade bailed on the **first** shape and never reached the one that works, so it failed on every request.

That's why it 500'd for you but streamed for me — same bad request, two different server reactions.

### Fix — `fix(timeshift): keep cascading on upstream 5xx`
Treat 5xx like 400/404 (log + try the next candidate shape). Only short-circuit on genuinely decisive, **ban-sensitive** statuses: `401/403/406` and any `3xx` redirect (for XC a `302` is the first sign of an IP ban, so we must stop hammering there). Added regression tests: *500-then-success* and *all-candidates-500*.

Also re-synced the branch with `dev` so it's mergeable again. **22/22** timeshift unit tests pass and I confirmed a real catch-up stream plays end-to-end through Dispatcharr's endpoints.

### On the timezone idea
`default_timezone` currently drives the **XMLTV/EPG** output timezone (`output/views.py`), not the timeshift URL — `timeshift_proxy` passes the client's UTC timestamp straight through to the provider on purpose (converting it was the root of a long run of "wrong programme plays" bugs in the original plugin). Auto-deriving the EPG timezone from the provider's `server_info` is reasonable, but it's tangled up with that pass-through logic, so I'd suggest a separate follow-up — happy to take it on.

**Updated test image:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

cedric-marcoux:
On the timezone idea — dug into it. Two useful findings, plus one real risk I want to flag before we pick an approach.

The provider tz is already available. `core/xtream_codes.py` pulls `server_info.timezone` during auth and `refresh_account_profiles` stores it on `M3UAccountProfile.custom_properties['server_info']['timezone']` on every refresh — so per-account provider timezone is there today, no extra fetching.

But two things make "just use the provider's tz" trickier than it looks:

1. **Single timezone per connection.** The XC API hands a client one `server_info.timezone`, and the EPG `start`/`end` + `time_now` all have to match it (the "timezone triple"). A client sees channels from all its providers under that one zone, so a true per-provider timezone in one XC response isn't expressible — multi-provider setups still need one chosen zone.

2. **`server_info.timezone` is really a display zone, but the catch-up archive is UTC-indexed.** The proxy forwards the catch-up timestamp to the provider unchanged on the assumption the archive is indexed in UTC (the "never convert" rule from the earlier timezone bugs). `default_timezone` today only shifts the EPG wall-clock the user sees. So if we auto-derive the provider's reported tz (say `Europe/Paris`) into the triple, a client that computes its seek offset from `server_info.timezone` could land on the wrong hour unless that zone matches the provider's actual archive indexing — which providers don't really tell us (they report a display tz). That's the exact class of bug the UTC rule was protecting against.

So before I code anything: how do you see the split between the **display** timezone (what `default_timezone` does now) and the **archive-seek** timezone (which has to stay aligned with the provider's UTC indexing)? If the goal is just to auto-fill a sensible display default from the provider instead of forcing UTC, I can do that with the manual field kept as an override — I just want to make sure we don't reopen the seek-offset bugs or break multi-provider setups. What did you have in mind?

SergeantPanda (changes requested):
So, you'll have to excuse my ignorance on catchup as I haven't fully investigated how it works completely on the backend.

That being said, I really don't understand why everything wouldn't be handled in UTC already?

EPG Programs are already stored and output in UTC time so any client connecting should already have those in UTC. If the user sets their timezone in a client that really shouldn't matter to Dispatcharr anyway should it?

Am I missing something fundamental here?

cedric-marcoux:
You're not missing anything fundamental — the epoch/archive side is already all-UTC. Two clarifications:

**It already *is* UTC by default.** `default_timezone` defaults to `UTC`, so `server_info.timezone`, the EPG `start`/`end` strings and `time_now` all come out in UTC and `_convert_xmltv_to_local_timezone` is a no-op. Nothing is forced on anyone — the "forcing one for all" worry is really just that the *option* exists, not that it's imposed.

**Why the option exists at all:** per the code comments / the original plugin's history, a couple of catch-up clients (TiviMate, iPlayTV) don't honor the XMLTV offset — they show the `start`/`end` wall-clock verbatim and use `server_info.timezone` to compute the seek into the archive. For those, pure-UTC output shows the wrong local time and seeks an hour or two off. That machinery exists only as an opt-in escape hatch for those clients (the "6 timezone bugs" the comments mention).

So: UTC everywhere by default; the setting is an opt-in for client quirks. Happy to go either way — keep it as the escape hatch, or rip it out and declare Dispatcharr strictly-UTC and push display onto the client. I'd just do a quick real-client check (TiviMate/iPlayTV with the setting at UTC) before deleting anything. I can run that, or defer to your read since you know the client landscape better.

SergeantPanda (changes requested):
Thanks for the detailed replies. I dug into this again on our side (code plus how XC clients build catch-up URLs) and want to clarify a few things, because I think we were partly talking past each other on timezone.

### `xc_get_epg` should always output UTC

On `dev`, `xc_get_epg` already outputs UTC wall-clock in `start`/`end` (`start.strftime("%Y-%m-%d %H:%M:%S")` on UTC-aware DB datetimes). Programmes are stored in UTC on ingest. `start_timestamp` / `stop_timestamp` are already correct UTC epochs.

In this PR, `default_timezone` defaults to `"UTC"`, and when it is UTC the conversion paths are effectively no-ops. So the default behavior is already UTC. I do not understand why we would add optional timezone conversion to `xc_get_epg` at all.

Dispatcharr is a proxy. We do not know at EPG response time which provider will ultimately serve a given catch-up request (channel failover, stream priority, etc.). A global or per-response timezone choice in `xc_get_epg` cannot reliably match the upstream provider anyway.

The simple model:

- `xc_get_epg` always outputs UTC in `start`/`end`
- `server_info.timezone` is always `"UTC"`
- `start_timestamp` / `stop_timestamp` stay as UTC epochs
- Client builds the timeshift URL from those UTC values
- Proxy forwards UTC to the provider unchanged

If that chain is internally consistent, what problem are we solving by converting EPG strings to a local zone? Display timezone is the client's job (it has `start_timestamp` and its own timezone setting). We should not complicate server output for that.

I would drop the `default_timezone` setting and any `astimezone()` logic in `xc_get_epg` and `_convert_xmltv_to_local_timezone`. Strict UTC everywhere on the API surface unless we have a specific failing client test that proves otherwise.

Also: the Settings UI says timezone is used for "provider timeshift URLs", but `timeshift_proxy` and the helpers explicitly do not convert timestamps for upstream URLs. They only reshape format (`YYYY-MM-DD:HH-MM` to underscore/SQL). The UI text does not match the code.

### How XC clients actually build catch-up URLs

I looked at several XC client implementations to understand what they do with EPG data. The pattern is consistent: they use the human-readable `start`/`end` strings, not the epoch fields.

**open-tv** ([`xtream.rs`](https://github.com/Fredolx/open-tv/blob/main/src-tauri/src/xtream.rs)):
- Fetches EPG via `get_simple_data_table`
- Builds `timeshift.php` from the `start` and `end` string fields
- Parses them as `%Y-%m-%d %H:%M:%S` (naive wall-clock)
- Reformats to `YYYY-MM-DD:HH-MM` for the `start` query parameter
- Does **not** use `start_timestamp` when constructing the catch-up URL

**open-tv #173** ([issue notes](https://github.com/Fredolx/open-tv/issues/173)):
- Documents the same flow for TiviMate-style clients
- `start` must be obtained from the EPG `"start"` field and reformatted
- `duration` is derived from `"end"` minus `"start"`
- URL shape: `baseurl/streaming/timeshift.php?username=...&stream=...&start=YYYY-MM-DD:hh-mm&duration=...`

**Kodi PVR IPTV Simple / XC catch-up** ([forum thread](https://forum.kodi.tv/showthread.php?tid=351431)):
- Catch-up URL: `/streaming/timeshift.php?username=...&password=...&stream=...&start=YYYY-MM-DD:HH-MM&duration=NN`
- Stalker-style path format also documented: `/timeshift/{user}/{pass}/{duration}/{Y}-{m}-{d}:{H}-{M}/{channel_id}.ts`

**xipl XC API wiki** ([Timeshift EPG adjustment](https://github.com/zaclimon/xipl/wiki/Xtream-Codes-API)):
- Documents an `epg_shift` parameter: integer hour offset between provider and user timezone when obtaining catch-up EPG data
- Exists because timezone alignment between EPG and catch-up playback is a known pain point in the XC ecosystem

**tuliprox XC API docs** ([catchup/timeshift](https://euzu-tuliprox-11.mintlify.app/api/xtream)):
- Documents both path-based and query-based timeshift endpoints
- Notes that `start` is "the start time of the timeshift segment (format used by the provider)"

The takeaway: whatever wall-clock we put in the EPG `start`/`end` strings is what clients send back to us in the timeshift URL. They parse those strings as naive datetimes and reformat them. They do not reliably use `start_timestamp` for URL construction.

On real XC providers, `start`/`end` are typically in provider local wall-clock (matching `server_info.timezone`), while `start_timestamp`/`stop_timestamp` are the correct UTC epochs. That is how providers behave natively.

For Dispatcharr, the consistent approach is: always emit UTC in `start`/`end`, always set `server_info.timezone` to `"UTC"`, and let the client send UTC back to us. If we are internally consistent, the client side of the chain works. The only open question is whether the upstream provider accepts UTC in `timeshift.php`, and that is a proxy concern, not an EPG formatting concern.

### Provider timezone (what I was actually suggesting)

When I suggested using the provider's timezone, I did not mean a global manual override in Settings. I meant: we already store each provider's `server_info.timezone` on every account refresh (`XCClient.get_account_info()` into `M3UAccountProfile.custom_properties`, visible in the account info UI).

That is separate from EPG output. EPG should stay UTC. If an upstream provider turns out to expect local wall-clock in `timeshift.php` (some XC panels do), the conversion belongs at proxy time only, using the stored timezone for the `m3u_account` actually handling that request via `get_channel_catchup_info()`. Not in `xc_get_epg`, and not via a global manual setting.

The multi-provider `server_info.timezone` issue you raised is real for a single `server_info` block (one XC connection can only report one zone). But that does not argue for converting EPG output. It argues for keeping EPG UTC and handling any provider-specific upstream conversion at request time if testing shows we need it.

### Debug logging

This should not live under catch-up settings. Dispatcharr already has `DISPATCHARR_LOG_LEVEL`. Timeshift verbose logging should follow the standard logger level (DEBUG), not a separate `timeshift_settings.debug_logging` toggle.

### Bottom line

- `xc_get_epg` always UTC: yes, and I think we should remove the option to do otherwise.
- Proxy always UTC pass-through: consistent with the above; any provider-local conversion only at proxy time if testing shows we need it.
- Global manual timezone setting: I would remove it.
- 5xx cascade fix: looks good from our testing, thanks for pushing that.

cedric-marcoux:
@SergeantPanda — implemented everything from your last review, plus the result of a deep empirical investigation that settles the timezone question with data. Two new commits, resynced with `dev` (v0.26.0, zero divergence).

## Your review items — all done

| Your ask | Status |
|---|---|
| `xc_get_epg` always outputs UTC; drop the conversion option | ✅ `start`/`end` always UTC; `default_timezone` removed everywhere |
| Drop `_convert_xmltv_to_local_timezone` | ✅ Deleted (function + regex + call site) |
| `server_info.timezone` always `"UTC"` | ✅ Hardcoded; `time_now` UTC; epochs untouched |
| Remove the global manual timezone setting | ✅ Gone from backend + Settings UI |
| Provider-local conversion, if needed, at proxy time only — via the account serving the request | ✅ Implemented exactly there (see below) |
| Settings UI text doesn't match the code | ✅ Field removed; remaining text now describes actual behavior |
| Debug logging should follow the standard logger level | ✅ `debug_logging` removed; gated on `logger.isEnabledFor(DEBUG)` |

Also removed `default_language` (it was never read; the EPG `lang` field is hardcoded). `timeshift_settings` is down to a single key: `xmltv_prev_days_override`.

## The proxy-time conversion is not speculative — your "if testing shows we need it" test exists

You wrote: *"any provider-local conversion only at proxy time if testing shows we need it"*. I ran that test against a real XC provider, pulling a few seconds of catch-up TS and identifying the actual frame content with ffmpeg against the channel's known schedule:

- Requesting `17-00` (the UTC instant of the 19:00 local news) returned the **17:00-local afternoon programme** — wrong.
- Requesting the same instant converted to the provider's reported zone returned the **19:00 news** — correct.

So this provider (and per the xipl `epg_shift` ecosystem note, it's not alone) interprets the timeshift URL timestamp as **its own local wall clock**, not UTC. Pure UTC pass-through seeks 1–2h off there.

The fix follows your prescription literally: `timeshift_proxy` converts the client's UTC timestamp to the **serving** provider's zone — the `server_info.timezone` we already store on every account refresh, read from the `m3u_account` resolved via `get_channel_catchup_info()`. DST-correct via `ZoneInfo`, no-op when the zone is `UTC`/missing/unknown. The EPG duration lookup keeps the original UTC value. Exactly one conversion in the whole chain, per-provider, multi-provider safe — and the EPG surface stays strictly UTC as you asked. After the change, the same ffmpeg check confirms the requested UTC instant plays the right programme end-to-end through Dispatcharr.

## Second finding from the same investigation: the URL *form* matters

The same frame-level testing exposed why catch-up could silently play live content: on this provider the `timeshift.php?start=` **query** form returns HTTP 200 with the **live** stream, ignoring `start` entirely — valid MPEG-TS, so the cascade happily accepted it. Only the **path** form (`/timeshift/{user}/{pass}/{dur}/{start}/{id}.ts`) actually seeks. The candidate cascade now tries the path form first and keeps `timeshift.php` as a last-resort fallback (rationale documented in `build_timeshift_candidate_urls()`).

## Hardening in the same push

- Redacted the upstream URL in the connection-error log path (requests exceptions embed the full URL, i.e. the XC credentials).
- Timestamp validated up front (400 on malformed input instead of forwarding it into the upstream URL).
- XC credentials URL-encoded in both builders; `Accept-Encoding: identity` so the TS-sync peek reads raw bytes.
- Stop-key polled on the 5-second heartbeat (was every ~25 MB), so stream-limit terminations free the provider slot quickly.
- `HTTPStreamReader` restored to its upstream form — the timeshift view streams via `iter_content` directly, so only the small `find_ts_sync()` helper remains in that file (+28 lines instead of +235).
- `catchup_days` semantics unified everywhere: MAX over a channel's catch-up streams (rollup SQL `FILTER`, same aggregation in the ChannelStream signal).
- New tests for the new behavior: proxy timestamp wiring (converted value → URL builder, original UTC → duration lookup), URL redaction, decisive 3xx break, invalid-timestamp rejection. **42/42** timeshift tests green.

## Verification

- Resynced with `dev` (v0.26.0) — mergeable, no migration conflicts (`0038` chains on `0037_auto_sync_overhaul`).
- 42/42 backend tests; frontend suite green (4 pre-existing locale-formatting failures on a French-locale host pass under `LANG=en_US`, unrelated to this PR).
- Live end-to-end on the rebuilt image: EPG strings UTC, `server_info.timezone=UTC`, and the frame-content check confirms the requested programme is what actually plays. Also verified on a real TiviMate client (guide times correct, catch-up plays the right programme).

**Updated test image:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

SergeantPanda (changes requested):
This is definitely looking better, thank you!

A few more things to look at and then I'll start testing this more thoroughly I think:

1) The access control for timeshift_proxy is missing network_access_allowed (should use the same as streams)
2) No failover for catchup:  get_channel_catchup_info() picks the first catch-up stream by channelstream__order with no capacity/failover logic. Live TV has multi-provider dispatch; catch-up does not. On max_connections=1 providers this is partly handled by stream-limit termination, but if the primary provider’s archive is unavailable, there’s no attempt on backup catch-up streams.
3) The changes to live_proxy > views.py seem unnecessary now. Mind making sure no old code isn't around making unnecessary changes?
4) Now that we are down to just 1 setting, it can probably be moved in to proxy settings.
Overall this is getting close I think. Thank you for digging in to the actual behavior with the providers. I think overall this is going to be MUCH better when we use UTC for everything.

cedric-marcoux:
@SergeantPanda — all four items from your review are implemented and pushed (one commit on top of a fresh resync with `dev`). Verified end-to-end on real clients (TiviMate + iPlayTV) before pushing.

## Your four points

### 1. `network_access_allowed` on `timeshift_proxy` ✅
Added with the same key and placement as the live XC endpoint: `network_access_allowed(request, "STREAMS", user)` right after credential auth, before any channel lookup or upstream contact. Covered by a test that asserts the `"STREAMS"` key and that nothing else runs when the gate denies.

### 2. Catch-up failover ✅
`get_channel_catchup_info()` is gone; the proxy now walks the channel's catch-up streams in `channelstream` order (`get_channel_catchup_streams()`), mirroring live playback:

- Each attempt carries **its own provider context**: account credentials, provider stream id, the account's reported `server_info.timezone` (the UTC→provider conversion is recomputed per attempt), user-agent, and the per-account URL-format cache.
- First streamable response wins; if every provider fails, the client gets the last failure.
- **Ban-safety is per account**: a decisive auth/ban-class failure (401/403/406) marks that account and skips its remaining streams (e.g. FHD/HD variants of the same channel) instead of hammering a banning provider — while other accounts, being different hosts, are still tried.
- Streams from **disabled M3U accounts are excluded**, same as live dispatch.

On the "capacity" half: per-user limits already apply through `check_user_stream_limits` (terminates the oldest session, same as live/VOD). Catch-up doesn't yet reserve per-profile provider slots the way live's `reserve_profile_slot` rotation does — I kept that out to avoid growing this PR further, but happy to wire it up here or as a follow-up if you want it.

### 3. `live_proxy/views.py` ✅
Restored byte-identical to upstream (`git diff` is zero lines) — the channel-id wrapper was indeed a leftover from the removed provider-stream-id fallback. I also swept the whole delta for other leftovers and reverted two unrelated drive-by hunks, so the diff is feature-only now.

### 4. Setting moved into proxy settings ✅
`timeshift_settings`, `TIMESHIFT_DEFAULTS`, `get_timeshift_settings()` and the Settings → Timeshift form/tab are all deleted. `xmltv_prev_days_override` now lives in `proxy_settings` (default in `get_proxy_settings()`, consumed by the XMLTV `prev_days` resolution) and shows up as one numeric field under **Settings → Proxy Settings** (0 = auto-detect, capped at 30). The pre-existing proxy-defaults unit test is updated accordingly.

## Test coverage grown to 69 backend tests
New coverage for everything above: failover success/exhaustion/skip semantics, decisive-account skip vs soft-failure retry on the same account, per-stream timezone conversion (two providers with different zones in one request), 406/connection-error cascade paths, network-gate 403, stream-limit and no-eligible-stream outcomes, the `server_info` strict-UTC guarantee, EPG duration-window resolution, and DB-backed tests for `xc_password` auth, `user_level` access and the failover stream ordering (catch-up-only, active accounts, `channelstream` order).

## Verification
- Resynced with `dev` (including the User-Agent centralization refactor — no interaction with this PR's paths) — mergeable, zero divergence.
- Backend: 69/69. Frontend: full suite green apart from 4 pre-existing locale-formatting tests unrelated to this PR (they pass under an en-US locale; my dev machine runs fr-BE).
- Live end-to-end on the rebuilt image through Dispatcharr's endpoints, plus real-client validation on TiviMate and iPlayTV: guide times correct, catch-up plays the requested programme.

**Updated test image:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

SergeantPanda (changes requested):
Thanks for the iteration on this. The UTC surface, failover, network gate, and test coverage are in good shape. Before I’m comfortable merging, there are three connection/accounting concerns I’d like addressed (or at least explicitly designed for). These aren’t nitpicks; they touch how Dispatcharr tracks provider capacity.

### 1. Catch-up rollup after bulk stream removal

`rollup_channel_catchup_fields()` only updates channels that **still have at least one stream from the account being refreshed**. That works well when streams are updated in place (e.g. `tv_archive` flips from 1 → 0).

The gap is **stale cleanup**: `cleanup_streams()` deletes streams via queryset bulk delete, which does **not** fire the `ChannelStream` post_delete signal that recalculates `is_catchup` on manual channels. After cleanup, if a channel no longer has any streams from that account, it also drops out of the rollup SQL entirely — so `is_catchup` / `catchup_days` can stay stale on the channel row.

**Symptom:** XC API still advertises `tv_archive: 1` while timeshift correctly returns “not supported” (no catch-up streams left).

Auto-created channels are mostly fine because auto-sync deletes the whole channel when streams disappear. Manual / multi-provider channels are the risky case.

**Ask:** After stale cleanup (or as part of rollup), also reset catch-up flags on channels that **lost** their last stream from this account, or extend rollup to recompute from all remaining `ChannelStream` rows regardless of whether the account still contributes a stream. Happy to discuss the simplest SQL approach.

### 2. Profile slot reservation (must integrate with `connection_pool`)

Timeshift currently enforces **Dispatcharr user** stream limits via `check_user_stream_limits()`, but it does **not** participate in the **provider profile** connection pool that live and VOD use:

- No `reserve_profile_slot()` before upstream connect  
- No `release_profile_slot()` when the session ends  
- No walk through alternate profiles when default is at capacity  
- No participation in shared credential pool (`credential_full`) logic  

Live does this in `Channel.get_stream()`; VOD does it in `MultiWorkerVODConnectionManager._check_and_reserve_profile_slot()`. Catch-up bypasses that entirely and opens upstream with account credentials directly.

That breaks the model where Dispatcharr knows exactly how many connections each profile is using. On `max_connections=1` providers, multi-user setups, or when live + catch-up overlap, the Redis pool counter and real upstream usage can diverge — which is how we get excess provider connections, 403s, and ban risk.

**Ask:** Reuse the existing helpers in `apps/m3u/connection_pool.py` (`reserve_profile_slot`, `release_profile_slot`, and the same default-first profile walk as live). Reserve before connect, release in the generator `finally` (and on error paths before streaming starts), and store which profile was reserved for cleanup (analogous to live’s `stream_profile:{stream_id}` pattern). Fail over to the next profile / catch-up stream when `profile_full` or `credential_full`, same as live — not only when the archive URL fails.

I don’t think catch-up needs to call `Channel.get_stream()` wholesale (different stream selection and URL building), but it **must** honor the same slot reservation contract.

### 3. Seek / FF / RW — new upstream connection per request (VOD-like risk)

Each `timeshift_proxy` request gets a new random `client_id`, runs stream limits again, and opens a **new** upstream `requests.get()`. There is no session persistence like VOD’s `vod_persistent_connection:{session_id}`.

XC clients typically seek in two ways:

- **Programme jump:** new timeshift URL with a different `start` timestamp → entirely new Dispatcharr request, new upstream connection.  
- **In-programme seek:** same URL with `Range: bytes=...` → Range is forwarded upstream, but still on a **new** HTTP connection (no reuse of the previous upstream socket).

The per-account URL format cache only avoids re-probing timestamp **shapes**; it does not reuse connections or profile slots across seeks.

This is structurally similar to the VOD problem where providers hold connections for some time after open, and players issuing rapid seeks/scrubs can stack multiple upstream sessions. User stream limits may eventually kill the old session via stop key, but that’s not the same as coordinated pool accounting, and `ignore_same_channel` explicitly excludes non-live types.

**Ask:** At minimum, with profile slot integration (#2), each seek request should reserve/release correctly so the pool stays accurate even when clients open overlapping requests.

**Ideally / worth discussing:** Should catch-up have a **session concept** (stable ID keyed by user + channel + programme, or similar) so FF/RW within one programme reuses one reserved profile slot and, where possible, one upstream connection for Range requests — similar in spirit to VOD persistence? Or should a new timeshift request for the same programme explicitly terminate and release the previous timeshift session before opening upstream?

Happy to re-test once these are addressed. The feature direction is solid; I think these three items are what’s left to align catch-up with how the rest of Dispatcharr manages provider connections.

cedric-marcoux:
@SergeantPanda — all three connection/accounting items are implemented and pushed (one commit, `c91f4f9a`, on a fresh resync with `dev` — zero divergence). Validated end-to-end on real `max_connections=1` accounts with Redis counter traces, plus TiviMate and iPlayTV.

## 1. Catch-up rollup after bulk stream removal ✅ (+ one finding from the investigation)

Implemented your second alternative: `rollup_channel_catchup_fields()` now ends with a **self-heal pass** that resets `is_catchup`/`catchup_days` on any channel left flagged with no remaining catch-up stream — independent of whether the account still contributes a stream, and regardless of how the link rows disappeared. Cheap (`is_catchup` is indexed) and idempotent.

One factual note from verifying the mechanism: `cleanup_streams()`' queryset bulk delete **does** fire the `ChannelStream` `post_delete` signal — registered signal listeners disable Django's fast-delete path, so cascaded through-rows are collected as instances and signaled (I verified this empirically with a `Collector.collect()` probe: 5 streams → 5 ChannelStream instances collected, `fast_deletes=[]`). A regression test now locks that behavior, and the self-heal covers every path the signal can't see (raw SQL, signal-less bulk ops, historic staleness).

While auditing this I did find a real route to your exact symptom: an account that is **disabled** (not deleted). XC output reads the denormalized `channel.is_catchup`, while the timeshift dispatcher filters `m3u_account__is_active=True` — so a disabled account's channels keep advertising `tv_archive: 1` until the account is re-enabled or refreshed. I left that as-is deliberately, because it matches live behavior (a disabled account's channels also remain listed yet unplayable). If you'd rather the catch-up flags follow account activation, happy to wire the rollup/signal to `is_active` and trigger recompute on the toggle.

## 2. Profile slot reservation ✅

Catch-up now honors the same slot contract as live and VOD, using the `connection_pool` helpers as-is:

- **Reserve before connect**: `reserve_profile_slot(profile, redis)` before any upstream request, walking the account's active profiles default-first (same ordering as `Channel.get_stream()`). `profile_full` / `credential_full` move to the next profile, then to the next catch-up stream — and **never** mark the account decisive (capacity is transient; the 401/403/406 ban-class semantics are untouched). When every eligible stream is blocked on capacity alone, the client gets **503** (the VOD pool-exhausted status). One deliberate asymmetry: an *upstream* failure does not retry the same account through alternate profiles — that would re-probe the same server (wasteful and ban-adjacent); only *reservation* failures walk profiles.
- **Per-profile credentials**: the upstream URL is built from `get_transformed_credentials(account, reserved_profile)` — the same credential-extraction live uses — so the connection consuming a profile's slot always authenticates as that profile, and the `credential_full` fingerprint accounting stays truthful. (Applying the profile regex to the built timeshift URL would silently no-op: user patterns target `/live|movie|series/` shapes.)
- **Release exactly once**: slot ownership is a one-shot Redis token (`timeshift_slot:{client_id}`) consumed with a transactional GET+DEL. The generator's `finally`, the response-close path, failed failover attempts and the takeover below all share it — no path can double-decrement. A client disconnecting **before the first chunk** still releases: the streaming iterator is wrapped so its `close()` (which Django registers as a resource closer and WSGI always invokes) tears the session down even when the generator never started. Any failure between reservation and the response owning the slot releases before propagating.
- **Which profile is reserved is stored** for observability/cleanup: in the ownership token and in the session's stats metadata (the `/stats` entry shows the actual reserved profile, not just the default).

Live evidence on a real `max_connections=1` provider profile: `profile_connections:{id}` reads **1 during the stream and 0 after close**, token consumed exactly once; with two overlapping seek requests the counter **never exceeds 1** (trace below in point 3).

One honest limitation, inherited from the pool itself: the counters have no TTL anywhere upstream, and timeshift doesn't yet have a reaper equivalent to VOD's stale-connection cleanup — so a hard worker kill (SIGKILL/OOM) mid-session strands the slot until the Redis flush at container start (`wait_for_redis`), same recovery story as the rest of the pool. The ownership token makes a reconcile pass straightforward (scan tokens whose client stats are gone, release through the same GET+DEL); I kept it out of this PR to avoid growing it further, but I'm happy to add it here or as a follow-up.

## 3. Seek / FF / RW ✅ (terminate-previous, deliberately widened to user+channel)

With #2, every request reserves/releases correctly, so overlapping seeks keep the pool exact — your minimum ask. On top of that I implemented your terminate-previous alternative with **one active catch-up session per user per channel**: a new timeshift request finds the user's previous session on the same channel, releases its provider slot **synchronously** (consuming its token, so the dying generator's own release becomes a no-op), unregisters its stats, and stops its generator via the standard stop-key.

Two scope notes, stated explicitly because they go beyond your sketch:

- I keyed displacement on **user+channel**, not user+channel+programme: a programme *jump* on the same channel also displaces the previous session (it's the same physical seek pattern), and re-requesting the same programme obviously does. Side effect: two devices sharing one Dispatcharr user on the same channel's catch-up displace each other — which seemed right for capacity accounting, but it's a one-line change to narrow if you disagree.
- The takeover runs **before** `check_user_stream_limits`, so a seek displaces its own predecessor instead of being denied when `terminate_on_limit_exceeded` is off.

Residual windows, for completeness: the displaced session's accounting is corrected synchronously, but its TCP socket closes lazily (stop key is polled on the 5-second heartbeat), so the provider can see a few seconds of physical overlap per seek; and two *near-simultaneous* seeks can both reserve briefly if the second one's takeover scans before the first registered — transient, self-healing, and the tokens keep the count exact throughout.

On VOD-style upstream connection reuse: I'd argue it doesn't transfer to catch-up — XC seeks are timestamp-indexed, so every FF/RW produces a *new* upstream URL (an open TS socket can't be repositioned), unlike VOD re-fetching one `final_url` with different Ranges. In-programme byte-Range scrubs are the only candidate and they're marginal in practice (responses are chunked, no Content-Length to scrub against). If your testing shows real Range-reuse traffic, I'm happy to revisit with a session abstraction.

## Tests & verification

- Backend suite grown to **94 tests**: reservation before upstream, release on every failover outcome (soft failure, decisive failure, exception paths before and during the upstream call, ownership-token write failure), profile walk on `profile_full`/`credential_full`, capacity-never-decisive, mixed capacity-vs-upstream-failure precedence (real failure wins over 503, both orders), all-capacity → 503 with zero upstream contact, exactly-once token semantics, never-started-generator release (locks the Django resource-closer contract), takeover scoping (same user+channel only, prefix-safe against channel 8 vs 80) and ordering (before the limit check), rollup self-heal + the bulk-delete signal regression.
- Resynced with `dev` — mergeable, zero divergence.
- Live end-to-end through Dispatcharr's endpoints with frame-level content checks (the requested programme is what actually plays, before and after a takeover) and the Redis counter traces above; real-client validation on **TiviMate and iPlayTV**.

**Updated test image:**
```
docker pull ghcr.io/cedric-marcoux/dispatcharr:timeshift-pr1242
```

doriancroatia-cyber:
Great feature !
Is there a way to have a "flag" in streams panel on stream that have catchup? So we easily know what stream to add, that would be really useful! 

Cheers

SergeantPanda (commented):
Sorry for the delay on this. I had some major regressions I had to get fixed up before I could look at PR's anymore. This is looking really good though. It looks like you addressed all of my concerns so far and I'm not seeing anything major right away. I will be playing with this in the next few days.

I may have one more bug fix release before this hits dev though.

I went ahead and merged in the current dev for you as I did create a conflict with my recent fixes. Please make sure it didn't cause any issues in the meantime.

Thanks again!

SergeantPanda (approved):
Ok. I spent quite a long time on this so far. 

I noticed there was no connection reuse logic built in yet so anytime I fired up a new catchup session to my provider it consumed all of my available connections.

I have fixed that now to work in a multi-worker environment (for the most part)

It may need some more work like VOD does but in my testing on 1 provider so far it does properly keep connection counts as you'd expect.

I still need to play with this a lot more and will end up refactoring a good amount of this into different modules but for now I think we are good to merge.

Thank you for your work on this, all of the research, initial code and testing with different clients definitely has helped get this feature production ready.

SergeantPanda (commented):
Before I do actually merge though, please look over what I've done and test it with your clients to make sure I didn't cause any regressions.

SergeantPanda:
Ok. I took quite a bit more time today to do some final testing on the xc parsing and noticed some significant slowness in how the is_archive was being handled so I went ahead and fixed the performance issues.

I think overall this is good enough to merge so I'm going to go ahead and do that. When you have time, please pull dev and make sure it still works with your providers. 

I am going to refactor the code a bit still but that can happen later.

SergeantPanda:
Thank you again for the work on this!

cedric-marcoux:
@SergeantPanda — pulled dev and re-validated everything on my providers (real `max_connections=1` profiles): **it works**. Full backend suite green (148 tests), and I traced the new session flow live: 301 → `session_id`, `profile_connections` at 1 during playback and 0 after close, the stored `provider_timestamp` confirms the per-provider timezone conversion rides along correctly, and frame-level checks confirm the requested programme is what actually plays. Also verified on real clients — TiviMate and iPlayTV both play catch-up fine through the new session flow, including seeks and programme changes.

Nice catch on the `is_archive` slowness — that deferred-field N+1 was mine (the `.only()` lists omitted the two catch-up fields the `bulk_update` wrote), and your session pool is a clean answer to the connection stacking. The per-chunk stop-key polling and the 416 passthrough are strict improvements too.

Three small things I spotted while reviewing your changes — offered as follow-up material, not objections:

1. **Reuse-failure drops the decisive flag.** When `_stream_reused_session` fails upstream it discards the pool session and returns `None`, and the fresh failover loop then re-reserves and re-connects **the same account** within the same request with an empty `decisive_accounts`. On a 403 (ban onset) that means a second ban-class hit at exactly the wrong moment. Propagating `timeshift_decisive` from the discarded reuse response into the loop's skip-set would restore the one-attempt guarantee — happy to send a small PR.

2. **Displaced-slot release is now asynchronous.** `_terminate_previous_timeshift_sessions` only sets the stop key now, so a cross-programme zap on the same channel can race its own takeover on `max_connections=1` providers: the new reserve can hit `profile_full` (first-attempt 503) if the dying generator hasn't polled its stop key yet. Per-chunk polling keeps the window small and my clients retried through it without visible failures — but if reports of first-attempt zap errors show up, preempt-and-wait on same-media busy sessions (like the mid-file scrub path already does) would close it.

3. **The sibling stream-limit exemption is gated on `ignore_same_channel_connections`** (default off). With `stream_limit=1` and default user-limit settings, an overlapping same-session probe can terminate the user's own in-flight catch-up stream and then 503 itself against the busy pool. Was gating it on that setting intentional, or should same-`client_id` siblings always be exempt?

One-line question: timeshift bytes now gate on `XC_API` while `stream_xc` live and XC VOD bytes gate on `STREAMS` — intended asymmetry for operators who restrict `STREAMS` to LAN?

Thanks for taking this over the line — the session pool plus the import fixes made the feature better than the PR I submitted. The crash-worker slot reconciliation I offered earlier still applies to the pool counters if you'd like it as a follow-up.

SergeantPanda:
Thank you for going over the changes, you are correct on these and I did make a couple of changes.

1) This should be fixed now. Reuse failures now feed into the same last_response / decisive_accounts path as the main loop, and _stream_reused_session returns the failure response instead of discarding it. Covered by a new test.
2) I think maybe we just monitor this and see if it becomes an issue? I'd be open to a PR if you think we should, I may be missing something on this one.
3) I agree on this and it is now fixed as well. We should not be using the same logic as live streams for that, wasn't intended.
4) I do agree on this as well, there is an inconsistency but I decided since timeshift is XC only it made the most sense. My long term goal would be to rework ALL XC api endpoints to use the xc api network access. This would be a larger refactor though so I thought I'd just start with the new one. I know I had told you to use stream (as that would be consistent with the other xc endpoints for streams).

cedric-marcoux:
Verified both fixes in `a9087207` — the reuse-failure path now lands in `last_response`/`decisive_accounts` exactly as the fresh loop does (one-attempt-per-banning-account guarantee restored), and the same-session probe exemption is ungated. Both look right to me, and the new failover/probe tests pin them down nicely.

Agreed on monitoring #2 — my clients ride through it and the scrub path already covers the common case. If first-attempt zap errors ever show up in reports, I'll send the preempt-and-wait patch.

Makes sense on the XC_API direction too. Thanks for the quick turnaround — running the merged dev in production here, all good.

SergeantPanda:
Great, happy to hear it's working correctly!

I know in Discord a user had mentioned something about FF/RW https://discord.com/channels/1340492560220684331/1443678482709086208/1520934112628703422

I haven't had a chance to look at that yet. Have you?

cedric-marcoux:
Looked at it — found it, reproduced it, and it's a one-liner class of bug. Repro without any client: request `start=X` → follow the 301 → play → then request a **different** `start=Y` with the **same** `session_id`: the two responses are byte-identical (same MD5), and the pool entry still shows X's `provider_timestamp`.

Root cause: `_stream_reused_session` replays the **stored** `descriptor["provider_timestamp"]` and only computes from the requested timestamp as a fallback, and `_acquire_idle_pool_session` never compares the request's media_id to the descriptor's. TiviMate keeps the `?session_id=` query when it rebuilds the seek URL with a new `start`, so every timestamp-jump FF/RW inside a session re-serves the session's original position — exactly the "snaps back to the rewind anchor" report from Discord. Direct-to-provider has no session_id, which is why it doesn't reproduce there.

Fix: in the reuse path, always recompute `provider_timestamp` from the **requested** timestamp (the timezone is per-account, not per-position) and update `media_id`/`provider_timestamp` on the pool entry at re-acquire — slot continuity stays, position follows the request. Happy to send the PR with a regression test (same session, different timestamp → different upstream position) if you want, or it's a quick one on your side.

SergeantPanda:
I'm fine with either. I feel like since you did the hard work you are the correct guy for the job though. 🙂 

Thank you!

cedric-marcoux:
Fix is up: #1425 — reuse path now always serves the requested timestamp and moves the pool descriptor with the seek (slot continuity untouched). 3 regression tests incl. an end-to-end same-session timestamp-jump; verified live (the two-curl repro now returns different content and the descriptor follows: `19-00 → 19-30`). Details in the PR.

PR #1245, opened 2026-05-13
fix: register plugin @shared_tasks on worker startup

### LINKED ISSUE #1244: Plugin @shared_task cron jobs silently miss ticks after worker restart
## Summary

Celery workers don't import plugin code at startup, so any plugin that uses `@shared_task` for cron-scheduled work loses its task registration after every worker restart. Beat fires the periodic task on time; the worker rejects it with `Received unregistered task`; nothing in the user-facing logs at INFO level explains why the schedule "didn't run". The task only becomes registered once *something else* (an event firing through `apps.connect.utils.trigger_event`, or any other call that hits `PluginManager.discover_plugins`) lazily imports `plugin.py` in the worker process — after which subsequent ticks fire correctly.

## Reproducer

1. Install a plugin that registers a periodic `@shared_task` at module level — e.g. [R3XCHRIS/telegram-alerts](https://github.com/R3XCHRIS/telegram-alerts), where `Plugin.SCHEDULED_TASK_CELERY_NAME = "telegram_alerts.send_daily_report"` is registered via `@shared_task(name=...)` at the bottom of `plugin.py`. (`VOD2MLIB` exhibits the same shape — its `vod2mlib.scheduled_rescan` task hits the same bug.)
2. Enable the plugin and register a `PeriodicTask` pointing at the task name (e.g. via the plugin's "Apply Schedule" action).
3. `docker restart dispatcharr` (or any other worker restart — Watchtower image updates count).
4. Wait for the next scheduled tick **without triggering any plugin events in between**.

Expected: the task fires.
Actual: the worker logs `ERROR celery.worker.consumer.consumer Received unregistered task of type '<plugin_name>.<task_name>'` and the task is silently dropped. `PeriodicTask.last_run_at` advances anyway, hiding the failure from `beat`'s perspective.

A subsequent channel/stream event (or any explicit `PluginManager.get().discover_plugins(...)` call) imports the plugin module in the worker process, registering the task via `@shared_task`. After that, `celery -A dispatcharr inspect registered` shows the task — but beat already missed the tick.

### Trace from a real worker, just after restart at 08:00 UTC:

```
2026-05-13 08:00:00,004 ERROR celery.worker.consumer.consumer Received unregistered task of type 'telegram_alerts.send_daily_report'.
KeyError: 'telegram_alerts.send_daily_report'
```

### Confirming the worker has the task *after* an event has fired:

```
$ docker exec dispatcharr celery -A dispatcharr inspect registered
    * telegram_alerts.send_daily_report
    * vod2mlib.scheduled_rescan
```

## Root cause

`apps.plugins.apps.PluginsConfig.ready()` deliberately skips eager plugin discovery in Celery worker processes:

```python
# Celery workers and management commands skip the eager pass: the post_migrate
# handler covers the DB sync, and `connect/utils.py` lazy-discovers on first
# event using the loader's cache.
```

This is correct for event dispatch (which is lazy by design), but `@shared_task` decorators only register with the worker's Celery app **at import time**. If the worker process never imports `plugin.py`, the task is never registered — until something else triggers the import.

## Impact

- Every plugin using `@shared_task` for cron work silently misses ticks after worker restart.
- For nightly cron schedules this often goes unnoticed because (a) restarts are infrequent, (b) one missed tick just delays a job by 24h.
- For more time-sensitive schedules (e.g. daily reports at a user-chosen time) the missed tick is user-visible — the report never arrives.
- The failure mode is silent at the default INFO log level. `PeriodicTask.last_run_at` advances even though the task didn't actually execute, so beat-side telemetry doesn't catch it.

## Suggested fix

Hook Celery's `worker_ready` signal in `dispatcharr/celery.py` to call `PluginManager.discover_plugins(sync_db=False)` on every worker boot. The signal is already imported and used for other initialization in that file. The discovery call imports every plugin's `plugin.py`, which triggers `@shared_task` registration before beat starts firing.

```python
@worker_ready.connect(weak=False)
def discover_plugins_on_worker_ready(**_kwargs):
    try:
        from apps.plugins.loader import PluginManager
        PluginManager.get().discover_plugins(sync_db=False)
    except Exception:
        logger.exception("plugin discovery on worker_ready failed")
```

Verified against the running Yoda container — pre-patch, the worker registry didn't include `telegram_alerts.send_daily_report` until an event fired. Post-patch (applied via `docker exec` for testing), the registry shows the task immediately after worker startup with no event needed.

## Environment

- Dispatcharr at `0704ec9` (current `dev` after #1232 merged).
- Python 3.13.
- Reproduces against any plugin shipping a module-level `@shared_task`.

Happy to open a PR against `dev` with this fix + a regression test if it looks right.

R3XCHRIS:
## What

Hook Celery's `worker_ready` signal in `dispatcharr/celery.py` to eagerly call `PluginManager.discover_plugins(sync_db=False)` on every worker boot. Importing the plugin modules runs their module-level `@shared_task` decorators, registering the tasks with the worker before beat starts firing.

## Why

Closes #1244.

Plugin code lives in `/data/plugins/<slug>/plugin.py`, outside `INSTALLED_APPS`. `celery.autodiscover_tasks()` only scans installed apps, so plugin modules are never imported in worker processes. The intentional skip of eager plugin discovery in `apps.plugins.PluginsConfig.ready()` (documented in code) is correct for the event-dispatch path (`apps.connect.utils.trigger_event` lazy-discovers on first event), but `@shared_task` decorators only register at import time — so plugins shipping module-level `@shared_task` code have their tasks unregistered with the worker until something else triggers `PluginManager.discover_plugins` in the worker process.

User-visible failure: beat fires the periodic task on time, the worker rejects it with `Received unregistered task`, and `PeriodicTask.last_run_at` advances anyway. Silent miss until *something* warms up the worker (a channel event firing, etc.) imports plugin.py, after which subsequent ticks fire correctly.

Both confirmed plugin schedulers shipping in the wild hit this:
- `telegram_alerts.send_daily_report` ([R3XCHRIS/telegram-alerts](https://github.com/R3XCHRIS/telegram-alerts))
- `vod2mlib.scheduled_rescan` ([R3XCHRIS/VOD2MLIB](https://github.com/R3XCHRIS/VOD2MLIB))

For nightly schedules the symptom is "the cron didn't run today" which is easy to chalk up to a flaky upstream M3U or a one-off; for time-of-day schedules (e.g. a daily report at 9am) it's noticeably user-visible.

## How

One block added to `dispatcharr/celery.py`, right after the existing `app.autodiscover_tasks()` call:

```python
@worker_ready.connect(weak=False)
def discover_plugins_on_worker_ready(**_kwargs):
    try:
        from apps.plugins.loader import PluginManager
        PluginManager.get().discover_plugins(sync_db=False)
    except Exception:
        logger.exception("plugin discovery on worker_ready failed")
```

`weak=False` so the handler isn't GC'd if Python ever decides the module-level function is unreachable. `sync_db=False` intentionally — discovery on every worker boot must not touch the DB schema, just import plugin modules.

The exception handling is intentionally broad: a broken `plugin.py` (import error, syntax error) shouldn't keep the worker from coming up. Errors get logged at ERROR level via `logger.exception`, the rest of the workers proceeds.

## How was it tested

`tests/test_celery_plugin_discovery.py` covers four cases:

1. Handler calls `PluginManager.discover_plugins(sync_db=False)` exactly once.
2. Handler swallows exceptions from `PluginManager.get()`.
3. Handler swallows exceptions from `discover_plugins()` itself (e.g. one plugin failing to import).
4. Handler is actually wired into Celery's `worker_ready` signal (guards against a future refactor accidentally dropping the `@connect`).

Verified inside a running Dispatcharr container with `python manage.py test tests.test_celery_plugin_discovery`:

- Against this PR: `Ran 4 tests — OK`.
- Against current `dev` (no fix): `FAILED (errors=4)` — `ImportError: cannot import name 'discover_plugins_on_worker_ready'`.

Also verified end-to-end against the live worker:

```
$ docker exec dispatcharr celery -A dispatcharr inspect registered
    * telegram_alerts.send_daily_report
    * vod2mlib.scheduled_rescan
```

Pre-fix: this listing was empty until an event lazily imported plugins. Post-fix (applied via `docker exec` to validate): tasks appear immediately after `docker restart dispatcharr`, no event needed.

## PR checklist

- [x] I have read CONTRIBUTING.md
- [x] Linked issue: #1244
- [x] I understand every line in this PR
- [x] No new endpoints / no migrations needed
- [x] Frontend: no changes
- [x] Tests included (regression test + signal connection sanity)
- [x] PR targets `dev`
- [x] CLA: I have read and agree to the terms in CONTRIBUTING.md


SergeantPanda:
Thank you for this. It makes sense to me. I made one small change to the test and modified the verbiage of your comment slightly. Otherwise, it looks good.

Issue #1246, opened 2026-05-17
[Feature]: Implement Schedules Direct EPG Integration


Shokkstokk:
## Summary

The codebase contains a `fetch_schedules_direct()` function in `apps/epg/tasks.py` that is entirely non-functional. It has `api_url = ''` and uses incorrect Bearer token authentication — any attempt to use Schedules Direct as an EPG source silently fails. This issue proposes completing this implementation.

## Current State

`fetch_schedules_direct()` exists as a stub:
- `api_url` is an empty string
- Authentication uses Bearer token, which is incorrect — the SD API requires a session token obtained via a separate auth endpoint with a SHA1-hashed password
- No station fetching, no program data fetching, no lineup management, no progress reporting
- The EPG source type `schedules_direct` is selectable in the UI but produces no data

## Proposed Implementation

A complete Schedules Direct integration using their [JSON API (v20141201)](https://github.com/SchedulesDirect/JSON-Service/wiki/API-20141201), following all documented client requirements.

### Two-Phase Fetch
**Phase 1 — On source creation (~4 API calls, fast):** Authenticate, fetch lineups, retrieve station metadata, persist EPGData entries. Fast enough to complete before the user runs Auto-match.

**Phase 2 — On manual refresh (variable API calls, MD5-delta):** On first run, fetches 20 days of schedule grid and program metadata. On subsequent runs, uses MD5 hashes to detect changes — only downloading schedules and programs that have actually changed. A steady-state refresh with no changes uses as few as 3-4 API calls.

### In-App Lineup Management
Users can search, add, and remove SD lineups directly from the EPG source form without visiting schedulesdirect.org. Includes:
- Country picker (fetched dynamically from SD API)
- Postal code search returning available headends/lineups
- Active lineup list with remove buttons
- Lineup count badge (X/4)
- Daily addition limit tracking with 24-hour lockout and automatic reset

### Rate Limit Compliance
- 2-hour minimum refresh interval enforced server-side
- Full SD error code handling (4100, 4101, 4102, 2100, 3000, 4001, 4004, 4007, 4008, 4009)
- `changesRemaining` tracked and persisted to database after every add
- Lockout UI with reset time shown when daily add limit is reached
- Cron schedule warning in UI for schedules more frequent than every 2 hours

### SD-Priority Matching
Before existing fuzzy/ML matching runs, checks each channel against SD stations via:
1. Gracenote ID exact match (via `tvc-guide-stationid` M3U attribute)
[...]

PR #1247, opened 2026-05-17
Implement Schedules Direct EPG Integration

Shokkstokk:
## PR Checklist

- [x] I opened a GitHub issue for this change before writing code — see #1246
- [x] I understand — line by line — every change in this PR
- [x] Backend: migrations are included (0023_schedules_direct)
- [ ] Backend: new endpoints are documented in the OpenAPI schema — *new endpoints are viewset actions on the existing EPGSourceViewSet, auto-documented by DRF router*
- [ ] Frontend: ESLint and Prettier pass cleanly — *see Notes below*
- [x] Tests are included for new functionality — 15 backend tests in `apps/epg/tests/test_schedules_direct.py`
- [x] No debug artifacts are left in the code
- [x] My PR targets the `dev` branch

---

## What

Implements the non-functional `fetch_schedules_direct()` stub that existed in the codebase. The stub had `api_url = ''` and used incorrect Bearer authentication — it would silently fail on any invocation. This PR delivers a complete, production-ready Schedules Direct (SD) integration including:

- Full EPG data fetch (20 days of program data)
- MD5-delta fetching — only downloads schedules and programs that have actually changed since the last refresh
- In-app lineup management — users can search, add, and remove SD lineups without leaving Dispatcharr
- Daily addition limit tracking with automatic 24-hour lockout and reset
- Comprehensive SD API error code handling
- 2-hour minimum refresh interval enforcement to protect against rate limit abuse

---

## Why

Schedules Direct is a widely used, high-quality EPG data source in the HTPC/self-hosted media community. The stub's existence implied intent to support it — this PR fulfills that intent. Linked issue: #1246.

**Note on SD approval:** Schedules Direct requires approved software status for API access (https://www.schedulesdirect.org/approvedsoftware). A request for Dispatcharr to be listed as approved software has been submitted to grabber@schedulesdirect.org. This PR should not be merged until that approval is confirmed. This requirement was identified during development and is referenced in SergeantPanda's comment on the previously closed PR #1034.

---

### Two-Phase Fetch Design

The integration uses a deliberate two-phase approach to respect SD's API rate limits and provide smooth UX:

**Phase 1 — Source creation (stations only, ~4 API calls, fast):**
On creation of an SD EPG source, the `fetch_schedules_direct_stations` Celery task fires automatically. It authenticates, checks system status, fetches lineups, retrieves station metadata, and persists EPGData entries (one per station). This completes quickly enough that users can immediately run Auto-match before committing to a full program fetch.

**Phase 2 — Manual refresh (full fetch, variable API calls):**
When the user triggers a manual refresh, the full fetch runs with MD5-delta logic (see below). On first run, all data is downloaded. On subsequent runs, only changed schedules and programs are fetched — dramatically reducing API call counts.

The intended user workflow: Create SD source → Add lineup(s) via Manage Lineups → Stations load automatically → Run Auto-match → Review matches → Manual refresh for full program data.

### MD5-Delta Fetching

SD provides MD5 hashes for both schedules (per station per date) and programs (per airing). This implementation uses both to minimize API calls:

**Schedule MD5 delta:** Before downloading schedules, `POST /schedules/md5` is called to retrieve the current MD5 hashes for all stations. These are compared against locally cached hashes in the new `SDScheduleMD5` model. Only stations/dates where the hash has changed are downloaded.

**Program MD5 delta:** Each schedule airing includes an MD5 hash for its program. These are compared against MD5 hashes cached in the `SDProgramMD5` relation table, keyed by program ID. Only programs with changed hashes are downloaded via `POST /programs`.

On a steady-state refresh where little has changed, this can reduce API calls from ~20 (1 lineup) to as few as 3-4 (auth + status + MD5 check). If nothing has changed, the fetch exits early with "No schedule changes detected."

### Lineup Management

Users can manage their SD account lineups directly from the EPG source form without visiting schedulesdirect.org. The "Manage Lineups" section provides:

- **Active lineups** — lists currently configured lineups with transport type, location, and lineup ID. Remove button per lineup.
- **Search** — country picker (fetched dynamically from `GET /available/countries` per SD's recommendation) + postal code input + search button. Results shown as a scrollable flat list of headends/lineups.
- **Add** — one-click add via `PUT /lineups/{lineup}`. Triggers `fetch_schedules_direct_stations` automatically so new lineup stations are available for matching immediately.
- **Lineup count badge** — shows `X / 4` (SD accounts support up to 4 lineups).

Three new viewset actions on `EPGSourceViewSet`:
- `GET/POST/DELETE /api/epg/sources/{id}/sd-lineups/` — list, add, remove lineups
- `POST /api/epg/sources/{id}/sd-lineups/search/` — search headends by country + postal code

### Daily Addition Limit Tracking

SD allows 6 lineup additions per 24-hour rolling window. This implementation tracks and enforces this:

- `changesRemaining` is captured from every successful `PUT /lineups/{id}` and `DELETE /lineups/{id}` response and persisted to `EPGSource.custom_properties`
- When SD returns code 4100 (`MAX_LINEUP_CHANGES_REACHED`), a lockout is set via `EPGSource.custom_properties` with a reset time of the next midnight UTC (00:00Z), matching SD's documented reset schedule
- The lockout clears automatically when midnight UTC passes — no manual intervention needed
- The UI shows warning alerts at 2 and 1 remaining adds, and a full lockout with reset time when exhausted
- Add and Search buttons are disabled during lockout
- If a user arrives with zero `changesRemaining` history (e.g. first time using Dispatcharr with an existing SD account), the first failed add triggers the lockout gracefully

### 2-Hour Minimum Refresh Interval

SD enforces ~200 API requests per 2-hour window. `fetch_schedules_direct()` checks `source.updated_at` at the start of every full fetch and returns early if less than 2 hours have elapsed, setting a friendly idle status message. Manual refreshes are subject to this check equally — no automated bypass is possible.

The EPG form enforces this in the UI:
- **Interval mode:** shows a validation error if the interval is set below 2 hours
- **Cron mode:** shows a persistent orange warning that schedules more frequent than every 2 hours risk account bans, with a link to SD documentation

### SD-Priority EPG Matching

A new matching pass runs in `match_epg_channels()` before the existing fuzzy/ML logic:

1. **Gracenote exact match** — compares `tvc_guide_stationid` (from M3U `tvc-guide-stationid` attribute) against SD station IDs
2. **Callsign/name match** — compares normalized channel name against station callsign stored in EPGData

SD-matched channels are merged with standard-matched channels for the bulk DB update and WebSocket payload.

### Authentication and Error Handling

- SHA1 password hashing is required by the SD API specification, not a design choice
- One token is obtained per task run and reused for all subsequent calls in that run (per SD spec — avoids `TOO_MANY_LOGINS` code 4009)
- Full SD error code handling: 4001 (expired), 4004 (lockout), 4007 (application disabled), 4008 (inactive), 4009 (too many logins), 4100 (add limit), 4101 (max lineups), 4102 (no lineups), 2100 (duplicate lineup), 3000 (service offline — idle status set, per SD spec do not retry for 1 hour)

### SD API Compliance

This implementation was developed against the SD JSON API specification (https://github.com/SchedulesDirect/JSON-Service/wiki/API-20141201) and follows the documented client requirements:

- Token obtained once per session, reused throughout
- `Content-Type: application/json` set on all requests
- System status checked before proceeding — disconnects if offline
- Lineup list fetched; stations fetched per lineup
- Schedules requested in batches ≤5000 stationIDs
- Programs requested in batches ≤5000 programIDs
- MD5 hashes used to minimize re-downloads (design goal explicitly stated in SD API docs)
- 20 days of schedule data fetched (SD JSON API provides ~20 days per their FAQ)
- Error responses handled to prevent account blockage (per SD's explicit warning in docs)

**Note on countries list:** SD recommends not hardcoding the country list. This implementation fetches `GET /available/countries` dynamically on form open, with a fallback hardcoded list if the request fails (no token required for this endpoint).

### XMLTV Guard

A guard at the top of `parse_programs_for_tvg_id` detects SD sources and returns early, preventing the XMLTV parser from attempting to parse SD data. An additional guard in `channels/signals.py` prevents the same issue during `refresh_epg_programs` signal handling.

### Real-Time UI Update

After auto-match completes, the `epg_match` WebSocket handler calls `fetchEPGData()` followed by `API.requeryChannels()`. The EPG store must be refreshed first so the channels table can cross-reference new `epg_data_id` assignments — without this, the table shows stale data until page reload.

---

### New Files (1)
| File | Description |
|------|-------------|
| `apps/epg/migrations/0023_schedules_direct.py` | Renames `api_key` → `password`; adds `username`; updates `custom_properties` help_text; creates `SDScheduleMD5` and `SDProgramMD5` models |

### Modified Files (11)
| File | Description |
|------|-------------|
| `apps/epg/models.py` | Renames `api_key` → `password`; adds `username`; removes SD-specific nullable columns in favour of `custom_properties`; adds `SDScheduleMD5` and `SDProgramMD5` models |
| `apps/epg/api_views.py` | Adds `_sd_authenticate`, `sd_lineups` (GET/POST/DELETE), and `sd_lineups_search` actions to `EPGSourceViewSet`; SD changes tracking stored in `custom_properties` |
| `apps/epg/serializers.py` | Adds `username` and `password` to `EPGSourceSerializer` fields; `password` marked `write_only` to prevent exposure in API responses |
| `apps/epg/tasks.py` | Full SD implementation; fixes program MD5 delta key mismatch; fixes schedule change counter; upserts `SDProgramMD5` relation table after each fetch; `programs_to_fetch` built as a set for O(1) membership checks in the MD5 cache upsert |
| `apps/epg/signals.py` | Fires `fetch_schedules_direct_stations` on SD source creation |
| `apps/channels/signals.py` | Skips XMLTV parser for SD sources in `refresh_epg_programs` |
| `frontend/src/api.js` | Adds `getSDLineups`, `addSDLineup`, `deleteSDLineup`, `searchSDLineups` methods |
| `frontend/src/components/forms/EPG.jsx` | Full lineup management UI; credential fields; rate limit warning; cron/interval validation; dynamic country list; `changesRemaining` alerts |
| `frontend/src/components/tables/EPGsTable.jsx` | Refactors status cell into standalone `EPGStatusCell` component for real-time progress bar updates; password masking |
| `frontend/src/WebSocket.jsx` | Adds `fetchEPGData` + `requeryChannels` after `epg_match` for real-time channel table update |

---

## How Tested

Tested end-to-end against a live Dispatcharr v0.24.0 instance using an active Schedules Direct subscriber account (US lineup, NJ market):

| Test | Result |
|------|--------|
| Source creation → stations-only fetch | ✅ 650 EPGData entries created; guidance message shown in table |
| Lineup management — list active lineups | ✅ Shows configured lineups with transport/location/ID |
| Lineup management — search by postal code | ✅ Returns flat list of available headends/lineups |
| Lineup management — add lineup | ✅ Lineup added; stations re-fetched automatically |
| Lineup management — remove lineup | ✅ Lineup removed; list refreshes |
| Daily add limit — 4100 handling | ✅ Friendly lockout message shown; Add/Search disabled; 24-hour reset persisted to DB |
| Daily add limit — lockout persists across sessions | ✅ Lockout survives modal close/reopen and page reload |
| Auto-match → real-time channel table update | ✅ 97 via SD Gracenote + 92 via standard = 189 matched; table updated without page reload |
| Full manual refresh → program data | ✅ 43,909 programs committed; progress bar accurate throughout |
| MD5-delta fetch — second run | ✅ "No schedule changes detected" — exits early without downloading |
| Second auto-match (already matched) | ✅ 0 new matches; exits cleanly |
| 2-hour minimum interval enforcement | ✅ Early return with idle status if triggered within 2 hours of last refresh |
| Zero-lineup account state | ✅ Idle status with guidance message instead of error |
| XMLTV guard | ✅ SD EPGData entries correctly skipped in parse tasks |
| Password masking in table and API | ✅ Shows "User: [username]", password not exposed in UI; `write_only` constraint on serializer prevents exposure in all API responses |
| Migrations — fresh apply | ✅ Single migration applies cleanly |
| XMLTV sources unaffected | ✅ Existing EPG sources continue to work normally |

---

## Notes

**On SD approval:** Schedules Direct requires approved software status before API use. A request has been submitted. This PR documents the full implementation so maintainers can review it in parallel with the approval process, but it should not be merged until approval is confirmed.

**On tests:** 15 automated backend tests are included in `apps/epg/tests/test_schedules_direct.py`, covering authentication, credential validation, SHA1 hashing, signal behavior, and serializer fields. All pass cleanly against the dev image.

**On AI assistance:** This implementation was developed with AI assistance (Claude). Per the CONTRIBUTING.md requirement: I have read every line of this diff, understand the reasoning behind each decision, and can explain any part of it during review.

**On ESLint:** `npm run lint` was run. Our changed files introduce no new errors. `EPG.jsx` is clean. `WebSocket.jsx` and `EPGsTable.jsx` have pre-existing warnings/errors on lines unrelated to this PR. The codebase has 204 pre-existing lint problems — none introduced here. `npm run format` was not run to avoid unrelated diff noise.

**On SD API behaviors discovered empirically (not documented):**
- `GET /lineups` returns HTTP 400 with code 4102 when no lineups are configured (not a 200 with empty array)
- `changesRemaining` is only available in `PUT /lineups/{id}` responses — not in `GET /status` despite documentation claiming otherwise
- SD error 4100 states "lineup changes for today" and resets at midnight UTC (00:00Z) per SD's documented behavior — not on a rolling 24-hour window
- SD returns `changesRemaining` on `DELETE /lineups/{id}` responses as well as `PUT` — this value is persisted to keep the count accurate


SergeantPanda:
Thank you for taking the time to submit to SD. I have not read through all of the changes yet but did read through what you wrote.

I have no problem at all with identifying Dispatcharr with a version during SD api hits, makes sense to me.

Could you please combine those migrations into 1 migration?

Shokkstokk:
Migrations consolidated — The three migrations (0023, 0024, 0025) have been squashed into a single 0023_schedules_direct.py migration that handles all SD-related schema changes in one operation.

User-Agent updated — The SD API requests now identify as Dispatcharr/{version} where the version is pulled dynamically from version.py at runtime, so it stays in sync automatically as the version bumps.

Both changes are pushed to the branch. Happy to address any other feedback as you work through the rest of the diff.

Cheers.

SergeantPanda (changes requested):
Thank you for your time on this.

While I do not have SD so I cannot fully test this code, I did read through what I could and had AI assist with review.

Bugs found:

The backend API endpoints for the lineup manager do not exist. I would expect changes to api_views.py for the frontend GET/POST/DELETE lineup functionality to work at all.

The program MD5 delta does not appear to be working correctly. It looks like it is writing correctly, but when reading back it builds the cache keyed by tvg_id (station ID) and then looks values up by program ID, which will never match. Every refresh will re-download all program metadata regardless of what is cached. On the same note, existing_program_md5s is constructed a few lines earlier but is never actually used anywhere.

The schedule change logging counter will always be wrong, sum(len(v) for v in changed_by_station.items()) iterates over key-value tuples so len() always returns 2, making the logged count 2 × number of changed stations instead of the actual total of changed station/date combinations.

Architectural feedback:

username is fine as its own column, no issue there. I would also like to see api_key renamed to password since the field was never used before this PR and the rename better reflects its purpose. It would also make it more reusable if other provider types need credentials in the future.

sd_changes_remaining and sd_changes_reset_at should be moved into custom_properties rather than adding nullable columns to EPGSource for all non-SD source types.

sd_program_md5 on ProgramData has the same problem, it is a nullable column on every program row regardless of source type. If program-level delta detection is wanted, a dedicated relation table (similar to SDScheduleMD5) keyed by program ID would be the right design and would also fix the key mismatch bug above.

Thank you again!

Shokkstokk:
@SergeantPanda - corrected and re-submitted for review.

Thank you.

SergeantPanda:
Thank you. I will take a look probably tomorrow more in depth, initial quick glance looks better so far 😃

SergeantPanda (changes requested):
Thank you for the quick turnaround, the new commit looks much better. A few remaining issues to address.

The lineup search endpoint will crash with a NameError: name 'dispatcharr_version' is not defined. Both sd_lineups and _sd_authenticate import it as a local variable inside their method bodies but sd_lineups_search skips that import and then uses it in the headers dict. Adding from version import __version__ as dispatcharr_version alongside the other imports at the top of that method fixes it.

The DELETE handler does not persist changesRemaining from the SD response. The POST handler correctly reads it from the response and saves it to custom_properties, but the DELETE handler returns the already-stored value without updating it. SD returns changesRemaining on deletes as well, so the count will drift out of sync if lineups are removed.

The lockout reset time is calculated incorrectly. The user-facing message says "Resets at midnight UTC" which matches SD's documented behavior. Error 4100 says "lineup changes for today" and every other SD rate limit counter explicitly resets at 00:00Z. However _save_sd_lockout sets reset_at = now + timedelta(hours=24), a rolling window. If a user hits the limit at 11:55pm UTC the message tells them it resets in 5 minutes but the code holds the lockout for another 24 hours. The reset time should be calculated as the next 00:00Z, not now plus 24 hours.

Minor: SDProgramMD5 has the same redundant index as SDScheduleMD5, unique_together = ('epg_source', 'program_id') already creates that index in Postgres so the explicit Index declaration is redundant. Not a bug, just wastes a small amount of storage.

Shokkstokk:
Sorry Sarge, been out on the road today. Just got back, saw your update. All changes have been adjusted/made as you directed. Please let me know if you happen to find anything else. 

Cheers.

SergeantPanda:
No need to apologize at all. Thank you for working on it, I'll take a look probably tomorrow sometime otherwise Thursday.

SergeantPanda (changes requested):
Thank you for addressing the issues with all the previous feedback! I have a couple more things I'd like to see changed:

The SD account password is exposed in the API response. EPGSourceSerializer lists password in fields but has no write_only constraint on it. The list and retrieve actions on EPGSourceViewSet are open to standard users (not just admins), so any authenticated user calling GET /api/epg/sources/ receives every SD account's plaintext password. Adding extra_kwargs = {'password': {'write_only': True}} to the serializer's Meta class fixes it; the field will still be accepted on create and update but will never be serialized out. I see I also need to do the same for the m3u server password, I will take care of that.

The programs_to_fetch list membership check in fetch_schedules_direct is O(N) per lookup. The list comprehension that builds md5_records on line 2761 filters with if pid in programs_to_fetch, and programs_to_fetch is a plain list at that point. With your test data of ~44,000 programs on a first full refresh, this iterates the full list once per entry in schedule_program_md5s, which also covers all programs. Converting programs_to_fetch to a set immediately after it is built (one line) reduces the membership check to O(1) and avoids the quadratic behavior on large lineups.

Have you heard back from Schedules Direct on approval by chance?

Thank you again!

Shokkstokk:
Both issues are addressed:

**Password exposure in API responses**
Added `extra_kwargs = {'password': {'write_only': True}}` to `EPGSourceSerializer.Meta`. The field is still accepted on create and update but will never be serialized out — no GET response will include it.

**O(N) membership check in `programs_to_fetch`**
Changed the list comprehension to a set comprehension so the `if pid in programs_to_fetch` check in the `md5_records` build runs in O(1). One downstream consumer (`program_id_list`) needs slicing for batch iteration, so that assignment is now `list(programs_to_fetch)` — the set is built once and used for the membership check; the list is derived from it for the batch loop.

**On SD approval:** I followed up with Schedules Direct this morning as I did not hear back earlier; they got back to me this afternoon! Long story short - we're approved, but before they deem us as *official*, I need to provide them with some additional info. I've pinged you on discord with that, so as soon as you get back to me, I'll get that over to them in short order.

bshiershome:
I can test this if needed, just let me know how.  I do have a SD account.

SergeantPanda:
> I can test this if needed, just let me know how. I do have a SD account.

Thank you for offering. There is no build of it so you'd need to spin up a dev build.

bshiershome:
Sounds good I'll give it a go with the dev build, i'm newish to docker so it's fighting me lol.  I tried the docker compose for the dev build and it errors out.

SergeantPanda:
It's definitely not for the faint of heart! Be sure to map the code you want to run into the correct path.

brownbe1992:
Passerby here - noticed Dispatcharr on the official approved apps list on the SD website this morning and came across this request. Just want to give a huge thank you to @Shokkstokk for the time and effort you've put in to building out this implementation and to @SergeantPanda for the time spent on review. Speaking for myself, this will improve my Dispatcharr experience immensely and I'm sure it will for others as well. I will try to spin up a Dev build to test but make no promises I will get it working. Looking forward to this going live in prod!

SergeantPanda:
I played around with this quite a bit and found some things to change for reliability and better extraction of data.

I pushed all of my changes directly into your PR. I did modify the migration so you'll need to reset your database to test.

I do ask that you look it over, play with it and make sure I did not break anything but I *believe* you'll find my changes a net positive.

One thing I noticed when playing with it that I'm not sure how best to tackle is the program posters. Currently they are not pulled in and it seems SD makes it difficult for some reason to get them (can't do in the same request as getting programs).

Another thing that could be improved, after I read through the SD API documentation, I saw that they offer 4 different logo versions, this PR has it hardcoded to `dark` but I think we could actually expose that to the user so they can select other versions if they'd like.

Overall I think it is really close to merging, just looking for some feedback on those things that I changed and your thoughts on the above.

Thanks again!

Shokkstokk:
Regarding your changes, all sound great. I did not get a chance to pull in and test your update(s) yet, I'll get around to that later tonight. To answer your ideas, I think I can program in a post-fetch to give the user an option to pull in posters, perhaps via a simple toggle in the setup (I'll have to dig into SD's API further). It would be a simple EPG retrieval process like normal followed by a 2nd retrieval for posters? Not sure how that would work - I'll figure it out regardless.

Regarding the Icons/Logos, I'm thinking possibly dropdown and options are "Disabled" which would use existing if the user has pre-existing logos already, followed up by the four options beneath it. I'll touch base with you on discord when I have some more information to run with. Cheers.

brownbe1992:
I was able to get a Dev build running on my system. A few notes from my testing:

- Successfully entered my SD username and password. It creates the EPG source and pulled my four selected lineups as I would have expected. The "Maximum of 4 lineups reached" notice is nice to have. However, the button at the bottom of the window remains "Create EPG Source". If I make a change, it gives me an error that the EPG Source already exists. I need to close the window and edit the source for the button to change to "Update EPG Source" and allow for changes to be made. A minor inconvenience.
- The SD password is required to make any changes to the EPG source. Otherwise, the configuration is wiped. Guide data isn't lost but the credentials need to be entered again to re-enable it and pull down lineup data.
- I was able to pull down guide data and link channels to guide data. I haven't done much more than this yet, but other than the issues I noted above it is good so far.

Shokkstokk:
Hey Sarge,

Addressing your feedback...

**Program posters** — Done. Implemented a separate artwork fetch that runs after program data is committed. Uses SD's `/metadata/programs/` endpoint in batches of 500, selects portrait (2x3/3x4) images, and caches them to disk so subsequent requests don't hit SD. The poster proxy endpoint handles auth, error caching (to avoid hammering SD if the account gets blocked), and uses `tokenExpires` from SD's response for token caching instead of a hardcoded TTL. One note on image selection: I switched the artwork preference from "Iconic" (generic actor photos with no show branding) to "Banner-L1" (key art with show title/logo). Significant visual improvement — the posters are now immediately recognizable in the guide.

**Station logo variants** — Done. Added a logo style selector to the SD configuration UI with all 4 options SD provides. There's also a toggle to enable/disable SD logos entirely — when disabled, the style selector is greyed out. When enabled, logos are applied to matched channels automatically after each refresh.

## Bug fixes

**`channel_group_id` not updated during M3U stream refresh** — This was a pre-existing bug your migration reset exposed. When groups get recreated with new IDs, existing streams matched by `stream_hash` go through the update path, but `channel_group_id` was never in the comparison, assignment, or `bulk_update` field list. The cleanup function then sees streams pointing to non-existent groups and deletes all of them. Added `channel_group_id` to the comparison, assignment, `only()` fetch, and `bulk_update` fields in both the standard and XC processing paths. **File:** `apps/m3u/tasks.py`

**SD content rating showing wrong region** — The original code took `contentRating[0]` blindly, which sometimes returned Scandinavian ratings (e.g., "12 ÅR") instead of US ratings ("TV-14") depending on array ordering from SD. Now extracts the lineup country from the lineup ID prefix (e.g., "USA" from "USA-NJ29486-X"), filters the `contentRating` array by the `country` field, and falls back to USA then first-available. Full ratings array is also stored for potential future region selection. **File:** `apps/epg/tasks.py`

**SD season/episode data not displaying** — The Gracenote metadata lookup only checked `metadata[0]`. SD's metadata array can have multiple entries in any order (TVmaze at index 0, Gracenote at index 1, etc.). If Gracenote wasn't first, season/episode was silently lost. Now iterates all metadata entries, falls back to TVmaze if Gracenote is absent. **File:** `apps/epg/tasks.py`

**Lineup reset timing** — Changed from midnight UTC calculation to a rolling 24-hour window per SD's API documentation ("6 adds in a 24 hour period"). **File:** `apps/epg/api_views.py`

**Error handling per SD docs** — SD's documentation explicitly states that failing to handle errors will get accounts blocked. Added error caching to the poster proxy: auth failures cache for 1 hour, network errors for 5 minutes. Prevents repeated requests from hammering a blocked account. **File:** `apps/epg/api_views.py`

## SD data enrichment

Audited the full SD program response schema against what we were actually extracting. We were leaving a lot of data on the table — all of it already present in the API response we download, zero additional API calls needed. Here's what's new:

- **Content advisory** — `contentAdvisory` array (e.g., "Adult Language", "Violence"). Displayed as an orange italic line below the description.
- **Entity type / Show type** — `entityType` ("Episode", "Movie", "Team Event") and `showType` ("Series", "Feature Film", "Sports event") now prepended to categories. Result: "Episode, Series, Drama, Medical" matching XMLTV parity instead of just "Drama, Medical".
- **Sports event details** — `eventDetails` with venue, teams (name + home/away). Displayed in a new section in the program detail modal for sports content.
- **Original air date** — Full `YYYY-MM-DD` instead of year-only. Displayed as "First aired" only for reruns (not shown alongside "NEW" badge — that'd be redundant).
- **Country** — Production country from SD's `country` array.
- **Runtime** — `duration` / `movie.duration` in seconds, converted to minutes and stored using the same `length` key format as the XMLTV parser. Displayed in the metadata section.
- **Movie quality ratings** — `movie.qualityRating` mapped to `star_ratings` using the same format as XMLTV's star-rating parser.
- **Character names** — `cast[].characterName` stored alongside actor name for potential future "as Character" display.
- **Full content ratings** — Complete `contentRating` array preserved in `content_ratings` for potential future region selection.

**Files:** `apps/epg/tasks.py`, `apps/epg/serializers.py`, `frontend/src/components/ProgramDetailModal.jsx`

## Cast cleanup

SD returns the entire cast including all guest stars with role labels like "(Actor)", "(Guest Star)". Compared to XMLTV which typically only lists top-billed names, this was visually overwhelming. Now:
- Capped to top 6 main cast by `billingOrder`
- Guest stars excluded from default display
- Removed `role` field from display (it's noise — of course they're actors, they're in the cast section)
- Result matches XMLTV presentation: clean, concise cast line

## Fields intentionally skipped

- `videoProperties` / `audioProperties` — Describes original broadcast quality, not what IPTV providers deliver after re-encoding. Would be misleading.
- `recommendations`, `keyWords`, `awards` — No current UI surface. Can revisit in future PRs.
- `episodeImage` — Would require per-episode image fetches. Not worth the API calls.
- `cast[].personId` — Would require additional calls to SD's celebrity endpoint for actor headshots.

## Non-SD change (flagged for your review)

**Logo cache-busting** — Added a `?v=` parameter to logo cache URLs based on a hash of the source URL. When a channel's logo changes (e.g., M3U logo replaced by SD logo), the browser fetches the new version instead of serving stale cache. The backend ignores the parameter — it's purely for browser cache invalidation. This touches core code (`apps/channels/serializers.py`) and affects all logo sources. Happy to revert if you'd prefer a different approach.

## Testing notes

- All changes tested against live SD API with a US lineup
- Verified: correct US content ratings, season/episode badges, entity/show type in categories, sports event details (NBA Finals with venue/teams), runtime, content advisory, cast truncation, Banner-L1 poster artwork
- XMLTV programs unaffected — all new fields conditionally rendered, simply absent for non-SD sources
- No new migrations — everything uses existing `custom_properties` JSONField
- No additional API calls for data enrichment — purely parsing more from existing responses

Let me know what you think.

SergeantPanda (changes requested):
I just spent quite a bit of time trying to get this to work and I think I have it back to partially working.

Seems you tried to merge dev into this PR but in doing so you overwrote a LOT of code that is required as well as had duplicate files somehow.

I am going to hand this back to you to continue playing with.

Please ensure it works fully before marking as ready.

I do not see the recent changes you made, I don't even see changes to the frontend that would be required for the picker you said there was for logos?

Honestly not sure what is going on with this branch at this point. 

I did add a couple changes for how caching should work though (using nginx like we do for channel logos) but haven't ran through a full test because there were a lot of other issues.

SergeantPanda (changes requested):
Also, I see this commit: 763d1430fdd10801f59c52c960f1751f9737dc9c I'm curious why this was included in this PR?

Shokkstokk:
Sarge,

Saw your feedback...one big UGH, sorry about that. I have several dev containers and I must have been going off the wrong (old) one that I had as a backup during the push.

Long story short, lack of sleep the last few days is likely catching up with me. We have a monster project where we're migrating our entire org off of vmware...this has been a painful week so far since last Friday. I'm really sorry you had to waste time fixing the mess; I just wanna get this done and wrapped finally...

I went back and nuked my dev containers then did a fresh clone of the branch (with all your fixes intact), rebuilt from a clean database, and verified everything end-to-end. Should be g2g now.

About your "Curious why commit 763d143 was included"...
This is a pre-existing bug I discovered during my further SD testing and your migration reset made me realize it - when channel groups get recreated with new IDs, existing streams matched by `stream_hash` go through the update path, but `channel_group_id` was never in the comparison, assignment, or `bulk_update` field list. The cleanup function then sees streams pointing to non-existent groups and deletes all of them. It's not SD-specific but it surfaced during the testing. It was driving me nuts and I just did the work here mainly because it was intefering with the testing/resetting. If you want me to move it to a separate PR, let me know.

Cheers.

SergeantPanda (changes requested):
Thank you, your explanation makes sense. I gotta admit, I was a bit frustrated when I started looking at it :joy:

Things look MUCH better now, thank you!

I made a couple small changes to improve the experience in dev mode that will help with troubleshooting. Also added a function to backfill posters if the setting is enabled afterwards. I also made a couple small changes to how actors and guest stars are saved so they will be properly output during xml generation. As well as added user-agent headers to program poster fetches.

One thing I noticed while playing around, if the "Use SD Station Logos" is not toggled the user can't select the logo type they'd like. In my mind, that should ALWAYS be enabled as it is still good to have them stored in the database so the user can pick and choose which logos get used.

If you want to tackle that great, I think we are really close to getting this merged. I may tweak a couple of things afterwards but overall it's looking pretty solid to me.

Thanks again!

Shokkstokk:
We're good to go. Toggle now only controls whether logos get applied; the style picker is always available so the preference is stored regardless.

Cheers.

SergeantPanda:
I made a few more changes to this. Now users can select their preferred poster style. Did some minor refactoring as well.

Thanks again for your work on this!

PR #1254, opened 2026-05-20
Enforce shared credential connection pools across M3U accounts (#1137)

### LINKED ISSUE #1137: [Feature]: Shared Connection Pool for Providers with Identical Credentials

Hey! I wanted to suggest a feature that could really help with more advanced setups.

**Feature Request: Shared Connection Pool / Credential-Aware Load Balancing**

Right now, Dispatcharr treats each provider independently. However, in some setups, multiple providers may actually use the **same IPTV credentials (same username/password)** — meaning they share a **single connection limit on the provider side**.

This can lead to issues where:

* Dispatcharr unknowingly uses multiple “providers” that are actually the same backend account
* Connection limits are exceeded unintentionally
* Stream selection doesn’t distribute load efficiently

### Suggested solution:

Introduce a way to group providers into a **shared connection pool**, where:

* Multiple providers can be linked together (manually or via matching credentials)
* A global **max connections limit** is enforced across the group
* Stream selection becomes **connection-aware**, distributing streams intelligently within the pool
* Prevents selecting streams if the shared pool is already at capacity

### Optional enhancements:

* Detect identical credentials automatically (with override option)
* Real-time tracking of active connections per pool
* Smarter fallback logic that considers pool usage instead of just provider priority

This would help users with:

* Duplicate provider setups
* Multi-line IPTV accounts
* More efficient and predictable stream distribution

Would love to hear your thoughts on whether something like this could be feasible in the future!

Goldenfreddy0703:
## Description

Enforces shared connection limits across M3U accounts that share the same provider login, using the existing **manual `ServerGroup`** model (per maintainer feedback).

**Problem:** Providers limit concurrent streams per login, but Dispatcharr counted connections per account/profile separately. Two accounts (e.g. XC + M3U playlist) with the same embedded credentials could exceed the provider limit.

**Solution:**

- Migration **`0020`** — `max_streams` on `ServerGroup`
- **`apps/m3u/connection_pool.py`** — centralized `reserve_profile_slot` / `release_profile_slot` for live TV and VOD
- When an account belongs to a group with `max_streams > 0`, gate on a group Redis counter at reserve/release; otherwise use per-profile counters as before
- **Profile rotation preserved** — selection uses each profile's own counter; group enforcement applies at reserve time
- Group counters scoped by **resolved provider login at runtime** (URL transforms / XC credentials), not a DB fingerprint column — multi-profile accounts with different logins keep independent limits
- **Settings → Server Groups** UI — create/edit/delete groups and set limits
- **M3U account form** — dropdown to assign accounts to a group
- Removed migration **`0021`**, auto-created `credential-pool-*` groups, and post-save auto-assignment hooks
- Stale `channel_stream` cleanup when the live proxy is no longer active (prevents stuck pool counters)

## Related Issue

Closes #1137

## How was it tested?

**Automated**

```bash
python manage.py test apps.m3u.tests.test_connection_pool -v 2 --keepdb
python manage.py check
python manage.py makemigrations --check
```

- `apps.m3u.tests.test_connection_pool` — 8 tests, all passing locally

**Frontend (PR files only)**

```bash
cd frontend
npx prettier --check src/api.js src/components/forms/M3U.jsx src/components/forms/ServerGroup.jsx src/components/tables/ServerGroupsTable.jsx src/pages/Settings.jsx src/store/auth.jsx src/store/serverGroups.jsx src/utils/forms/M3uUtils.js
npx eslint src/components/forms/ServerGroup.jsx src/components/tables/ServerGroupsTable.jsx src/store/serverGroups.jsx src/utils/forms/M3uUtils.js
```

- Prettier — pass (after formatting PR files)
- ESLint — pass on new files; `api.js` / `M3U.jsx` only touch pre-existing lint findings outside this PR's additions

**Manual (Docker dev stack)**

- [x] XC account with 3 profiles (`max_streams=1` each) → rotation across different provider logins works
- [x] STD + XC accounts with same URL login, manually assigned to same ServerGroup → shared limit enforced
- [x] Live channel then VOD → alternate profile selected when default is full
- [x] VOD then live → works
- [x] At-capacity requests return clean **429/503**, not provider "invalid HTTP status" errors
- [x] Server Groups panel and M3U account group dropdown work as expected
- [ ] `python manage.py migrate m3u` on fresh upgrade (maintainers / CI environment)

## Checklist

- [x] Backend: new API endpoints appear correctly in the OpenAPI schema *(no new endpoints — uses existing ServerGroup CRUD API)*
- [x] Frontend: ESLint and Prettier pass cleanly on files changed in this PR


JCBird1012:
@Goldenfreddy0703 - I guess I'm curious to your use case as to why you'd have multiple M3U accounts in Dispatcharr with the same credentials?

Goldenfreddy0703:
short answer: **I don't add the same account twice to duplicate channels.** I use **two different account types in Dispatcharr that both authenticate with the same IPTV login**, because each side does something the other doesn't do well.

### How my setup is split

1. **Standard M3U account** — usually a modified playlist + custom EPG for live TV (cleaner channel list, better guide/naming).
2. **Xtream Codes (XC) account** — same subscription, but used for **VOD and channel groups that aren't available (or aren't practical) through the M3U path**, and because Dispatcharr's VOD workflow is built around XC rather than plain M3U.

So in Dispatcharr that's **two providers**. On the IPTV side it's still **one username/password and one max-connection limit**.

### Why that causes a problem today

Dispatcharr enforces limits **per account/profile**. The provider enforces limits **per login**. So I can be "under limit" on both accounts in the UI while the provider is already at capacity — or pick streams on one account without knowing the other is using the same backend login.

That's exactly what I described in #1137: duplicate provider setups where Dispatcharr sees separate sources, but the service sees one shared pool.

### Why profiles aren't the answer (for this case)

Profiles are great when you want **different logins or URL rewrites on a single account**. They don't help when you intentionally have **two accounts** (M3U + XC) that still resolve to the **same credentials** at playback time. You'd still double-count against the provider.

### What the PR is trying to do

Make Dispatcharr **credential-aware**: if two accounts/profiles use the same effective login, they share one connection pool and selection respects that — without forcing users to pick one access method or the other.

Hope that clarifies the use case.

SergeantPanda (changes requested):
Hey, thanks for putting this together. The problem you're solving is real and worth fixing. Before we go further with the implementation though, I'd like to talk through the design.

The ServerGroup table is already there for this

ServerGroup already exists as a model with a FK on M3UAccount, a serializer, a CRUD API endpoint, and a form field. It's a user-managed grouping concept and it fits this use case directly. What this PR needs to do is complete that original intent by adding connection enforcement on top of the existing relationship, not work around it with an auto-created hidden layer.

The design I'd prefer

ServerGroup gets a max_streams field (your migration 0020 is exactly right for this).
The connection enforcement checks: if the account belongs to a group with max_streams > 0, gate against the group Redis counter; otherwise fall back to the per-profile counter as before. That's the entire change to the pool logic.
Users explicitly assign accounts to a group. Accounts from different types (XC + M3U) sharing the same login get assigned to the same group manually. That's one configuration step per provider.
Why I don't want auto-fingerprinting as the primary mechanism

The SHA256 fingerprint idea is clever, but making it the primary enforcement path creates a hidden system: rows appear in ServerGroup with names like credential-pool-9f8a3b... that aren't visible or manageable in the React frontend (there are currently zero frontend references to ServerGroup). It also runs expensive DB work on every profile save. Auto-detection can be a convenience layer on top, surfaced as a suggested group assignment during account setup, but the enforcement model should rest on an explicit user-owned relationship.

The frontend gap is actually an opportunity

Since the React UI has no ServerGroup panel at all today, this PR is a natural place to add one. Something like:

A Server Groups settings panel where you can create and name groups and set a max_streams limit for each.
On the M3U account edit form, a dropdown to assign the account to a group (or leave it unassigned). The dropdown would list existing groups with their limit shown, e.g. My IPTV Provider (max: 2).
That's a small amount of frontend work but makes the feature fully transparent and user-controllable. Users can see at a glance which accounts share a pool and what the limit is, and can adjust or reassign without touching anything invisible.

What I'd like to see

Keep migration 0020 (max_streams on ServerGroup).
Drop migration 0021 (fingerprint column and backfill). No auto-created rows.
reserve_profile_slot / release_profile_slot check profile.m3u_account.server_group and gate on the group counter when present. A handful of lines.
A Server Groups settings panel and a group assignment dropdown on the account form.
The stale channel_stream key fix is genuinely good. Please split it into a separate PR so it can go in cleanly on its own.
If you want to keep the fingerprint detection as an optional "suggest a shared group" hint in the UI (e.g. "these two accounts appear to use the same login, assign them to a group?"), that's welcome as an enhancement on top. I just don't want it as the invisible enforcement backbone.

Happy to discuss further. Appreciate the effort here.

Goldenfreddy0703:
@SergeantPanda — thanks for the thoughtful review. That direction made a lot of sense, and I pushed an update (`bb9b95b`) that refactors the PR to match it. Wanted to walk through what changed.

---

## What I changed based on your feedback

### Kept

- Migration **`0020`** — `max_streams` on `ServerGroup`
- **Manual `ServerGroup` assignment** as the enforcement model
- **`reserve_profile_slot` / `release_profile_slot`** gating on the group Redis counter when `account.server_group.max_streams > 0`, otherwise per-profile counters as before
- **Stale `channel_stream` cleanup** when the proxy is no longer active (fixes stuck assignments / bad pool accounting)

### Removed

- Migration **`0021`** (fingerprint column + backfill)
- Auto-created **`credential-pool-*`** `ServerGroup` rows
- Post-save hooks that auto-assigned accounts to fingerprint pools
- Auto-fingerprinting as the **primary** enforcement path

### Added (frontend)

- **Settings → Server Groups** panel — create/edit/delete groups and set `max_streams`
- **M3U account form** — dropdown to assign an account to a group (or leave unassigned), showing the group limit

---

## How enforcement works now

1. **Profile selection** (live + VOD) still rotates across profiles using each profile's own `profile_connections:{id}` counter — same as pre-pool behavior.
2. **On reserve/release**, if the account belongs to a `ServerGroup` with `max_streams > 0`, we also increment/decrement a group counter.
3. **Group counter is scoped by resolved provider login** (runtime fingerprint from `get_transformed_credentials` / stream URL transforms — *no DB column*). That way:
   - XC + M3U accounts manually assigned to the same group **share one limit when they use the same login**
   - Multi-profile accounts with **different URL transforms / logins** keep independent limits and can still rotate (this was broken when the whole group shared one counter)

> Happy to simplify further if you'd rather a flat per-group counter with no login scoping — I kept the scoping because it matches how providers actually count connections and it fixed real failures in testing.

---

## Re: stale `channel_stream` fix

Still bundled in this PR for now since it was tightly coupled to pool accounting bugs I hit while testing. I can **split it into a separate PR** if you'd prefer that for review/merge order — just say the word.

---

## Re: fingerprint as "suggest group" hint

Not implemented in this pass — wanted to land the manual model cleanly first. Fine to add as a follow-up if useful.

---

## Testing

- [x] `python manage.py test apps.m3u.tests.test_connection_pool` — 8 tests, passing locally
- [x] Manually verified on my Docker dev stack:
  - XC account with 3 profiles (`max_streams=1` each) → rotation across different logins works
  - Live + VOD concurrently → alternate profile picked when default is full
  - Two accounts in the same manual group with the same login → shared limit enforced
  - Clean **429/503** when at capacity (not provider "invalid HTTP status" errors)

---

Let me know if anything still feels off with the design or if you want the group counter flattened before merge. Appreciate you taking the time on this.

SergeantPanda:
I'll take another look at this soon. In the meantime, please fill out the PR form.

SergeantPanda (changes requested):
Before I look at this any further, please fill out the PR form.

Goldenfreddy0703:
## Description

Enforces shared connection limits across M3U accounts that share the same provider login, using the existing **manual `ServerGroup`** model.

Providers limit concurrent streams per login, but Dispatcharr previously counted connections per account/profile separately. This PR adds:

- Migration **`0020`** — `max_streams` on `ServerGroup`
- **`apps/m3u/connection_pool.py`** — centralized `reserve_profile_slot` / `release_profile_slot` for live TV and VOD
- Group Redis counter enforcement when an account belongs to a `ServerGroup` with `max_streams > 0`; otherwise per-profile counters as before
- **Profile rotation preserved** — selection uses each profile's own counter; group enforcement applies at reserve time
- Group counters scoped by **resolved provider login at runtime** (URL transforms / XC credentials), not a DB fingerprint column
- **Settings → Server Groups** UI and **M3U account form** group assignment dropdown
- Removed migration **`0021`**, auto-created `credential-pool-*` groups, and post-save auto-assignment hooks
- Stale `channel_stream` cleanup when the live proxy is no longer active

Refactored per review feedback in commit `bb9b95b`.

## Related Issue

Closes #1137

## How was it tested?

**Automated**

```bash
python manage.py test apps.m3u.tests.test_connection_pool -v 2 --keepdb
python manage.py check
python manage.py makemigrations --check
```

- `apps.m3u.tests.test_connection_pool` — 8 tests, all passing locally

**Frontend (files changed in this PR)**

```bash
cd frontend
npx prettier --check src/api.js src/components/forms/M3U.jsx src/components/forms/ServerGroup.jsx src/components/tables/ServerGroupsTable.jsx src/pages/Settings.jsx src/store/auth.jsx src/store/serverGroups.jsx src/utils/forms/M3uUtils.js
npx eslint src/components/forms/ServerGroup.jsx src/components/tables/ServerGroupsTable.jsx src/store/serverGroups.jsx src/utils/forms/M3uUtils.js
```

**Manual (Docker dev stack)**

- XC account with 3 profiles (`max_streams=1` each) → rotation across different provider logins works
- STD + XC accounts with same URL login, manually assigned to same ServerGroup → shared limit enforced
- Live channel then VOD → alternate profile selected when default is full
- VOD then live → works
- At-capacity requests return clean 429/503, not provider "invalid HTTP status" errors
- Server Groups panel and M3U account group dropdown work as expected

## Checklist

- [x] I have read the [CONTRIBUTING.md](https://github.com/Dispatcharr/Dispatcharr/blob/dev/CONTRIBUTING.md) in full
- [x] I agree to the [Contributor License Agreement](https://github.com/Dispatcharr/Dispatcharr/blob/dev/CONTRIBUTING.md#contributor-license-agreement)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema *(no new endpoints — uses existing ServerGroup CRUD API)*
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) on files changed in this PR

SergeantPanda (changes requested):
Thanks for the rework - the shift to manual ServerGroup assignment is exactly the right direction. Before this can merge there are a few things that need addressing, plus a design question worth discussing.

Required fixes

1. update_stream_profile leaks counters when a ServerGroup is enforced

The new early-return branch in Channel.update_stream_profile (when get_enforced_server_group_for_profile(new_profile) is truthy) rewrites stream_profile:{stream_id} to the new profile ID but skips the DECR/INCR pipeline on the per-profile counters. This means:

The old profile counter is never decremented and stays inflated, making that profile appear permanently at capacity until Redis is flushed
release_profile_slot at stream end will decrement the new profile counter, which was never incremented, clamping it negative then back to 0
The per-profile counter swap still needs to happen for all profile switches regardless of group membership. The group counter does not change (it is still one stream), but the profile counters do. The pipeline that was there before should run unconditionally.

2. The fingerprint counter uses the wrong cap

In _reserve_server_group_slot_for_profile, the fingerprint counter is checked against group.max_streams. But the fingerprint counter's purpose is to enforce the per-credential limit across accounts, and that limit is profile.max_streams, not the group total.

Concrete example:


Group cap = 2, each username1 profile has max_streams = 1Stream 1 -> M3U username1 profile: profile_counter = 1/1, fingerprint_counter(username1) = 1, group = 1 -- all passStream 2 -> XC  username1 profile: profile_counter = 1/1, fingerprint_counter(username1) = 2 <= group.max_streams(2) -- PASSESProvider now sees 2 connections on username1 against a real limit of 1.
The fingerprint sub-counter needs to be capped at profile.max_streams, not group.max_streams.

3. Missing tests for the above paths

update_stream_profile with a ServerGroup assigned (would have caught #1)
Reserve behavior when fingerprint resolution returns None (falls back to "unknown" and unrelated profiles collide on the same counter)
Release when the profile row no longer exists
Design discussion - group profiles

Setting aside the cap bug for a moment, there is a deeper question about whether credential fingerprinting is the right mechanism here at all.

The fingerprinting for STD/M3U accounts is fragile. The current code samples a stream URL from the DB, applies the profile's URL transform, then tries to parse credentials out of an XC-style path. If the M3U playlist does not embed credentials in that format the fingerprint silently falls back to account.username/password, which may or may not match what the XC account resolves to after its own transforms. There is no reliable way to detect this has gone wrong.

An alternative worth considering is group profiles. They would work exactly like M3U account profiles - same modal, same fields (name, max_streams, search/replace patterns) - except they belong to the group and apply across all accounts under it.

To understand how the limits interact, it helps to look at how accounts and profiles already work today. M3UAccount.max_streams seeds the default profile limit via a post-save signal and stays synced to it. There is no separate account-level Redis counter - the limit is enforced through the profile counter. Additional profiles each have their own independent counter. These are not nested ceilings where the account limit is an umbrella over the profiles; they are independent checks that each need to pass at reserve time.

Group profiles would follow the exact same pattern:

ServerGroup.max_streams gets its own independent Redis counter, checked the same way a profile counter is checked today
Each ServerGroupProfile gets its own independent counter as well
So when a stream comes in, reserve checks each applicable counter in sequence: per-account-profile counter (handles rotation within an account), per-group-profile counter (enforces the per-credential cap across all accounts in the group), group counter (the group-wide limit). All of them need to pass. None is an umbrella over the others - they are just independent limits that all apply simultaneously, same as how things work today at the profile level.

The user would create a group profile for each credential slot they want to cap, name it something meaningful, set max_streams to the provider's limit for that credential, and then assign the relevant account profiles across their M3U and XC accounts to it. Dispatcharr enforces it without needing to inspect or guess at credentials.

The tradeoffs compared to the current approach:

More upfront work: new model, migration, serializer, API endpoints, and a way to link account profiles to group profiles in the UI
The ServerGroup CRUD, UI, and store work in this PR carries over directly and is not wasted
No credential guessing, works correctly and predictably for any account type, no DB queries on the reserve/release hot path
The current fingerprinting approach requires less user configuration but is unreliable for STD/M3U accounts and needs the cap fix above to be correct even for XC accounts.

Worth deciding which direction to go before another rework. Let us know and we can give more specific guidance either way.

Goldenfreddy0703:
@SergeantPanda — thanks again for the thorough review. Latest push on `feature/1137-credential-pools` is commit `6e07dce3`, which addresses all three required fixes and includes manual testing on a real XC + M3U setup. Notes below on what changed, what we tested, and the design question.

---

## Required fixes (done)

### 1. `update_stream_profile` counter leak

Removed the early-return branch when a `ServerGroup` is assigned. Profile counters now **always** DECR/INCR on switch via the existing pipeline; the credential-scoped group counter stays unchanged (still one active stream on that login).

Added `UpdateStreamProfileTests.test_switch_updates_profile_counters_when_group_assigned`.

### 2. Fingerprint counter cap

The credential-scoped counter (`server_group_connections:{group_id}:{fingerprint}`) is now capped at **`profile.max_streams`**, not `group.max_streams`.

This fixes the example where group max = 2 but each login only allows 1 — a second stream on the same username is blocked at the credential counter even though the group limit has headroom.

Added `test_same_credential_capped_at_profile_max_not_group_max`.

**Note:** We briefly tried adding a separate group-wide total counter capped at `group.max_streams`, but it broke multi-login profile rotation (e.g. live on login A + VOD on login B blocked when group max = 1). That total counter was removed. Today, `group.max_streams > 0` enables credential pooling for accounts in the group; the per-login cap comes from each profile's `max_streams`.

### 3. Missing tests

Added coverage for:

- `update_stream_profile` with a grouped account
- Same credential capped at `profile.max_streams`, not `group.max_streams`
- Different logins both streaming when group max = 1 (regression guard for rotation)
- Fingerprint resolution returning `None` → credential counter skipped (no `"unknown"` bucket collision)
- `release_profile_slot` when the profile row no longer exists

All **11** tests in `apps.m3u.tests.test_connection_pool` pass locally.

---

## Manual testing (passed)

Tested with two accounts in the same Server Group (XC + STD M3U, same provider login on overlapping profiles):

| Scenario | Result |
|----------|--------|
| Live on account A, second stream on account B with same login | Blocked (expected) |
| Live on profile A (login 1), VOD/second stream on profile B (login 2) via rotation | Works (expected) |
| Profile switch while streaming | Counters swap correctly |

---

## Design question — group profiles vs fingerprinting

We prototyped the group profiles approach (explicit `ServerGroupProfile` model, linking account profiles to shared slots, three independent counters). It matched your proposal architecturally, but the extra configuration layer felt confusing in practice — easy to mix up with M3U account profiles — and we expect other users would hit the same wall.

**Decision for this PR:** stick with **manual Server Groups + runtime fingerprinting** for cross-account enforcement, with the cap/switch fixes above. Rationale:

- Satisfies the required fixes and real-world testing without adding a new model + linking UI
- Server Groups CRUD/settings panel work carries forward if group profiles are pursued later
- Less upfront setup: assign accounts to a group, set profile `max_streams`, fingerprint handles same-login detection on reserve

Happy to pivot to group profiles in a follow-up if you'd rather not merge fingerprinting at all — just say the word.

---

## Summary of current behavior

1. User creates a **Server Group** and assigns M3U/XC accounts to it (`max_streams > 0` enables pooling).
2. **Profile selection** (live/VOD rotation) uses each profile's own Redis counter only — unchanged pre-pool behavior.
3. On **reserve/release**, a credential fingerprint counter enforces the per-login cap across accounts in the group (`profile.max_streams`).
4. No auto-created groups, no migration `0021` fingerprint column — manual assignment only.

Let me know if anything in `6e07dce3` still looks off. Thanks again for the clear feedback.

SergeantPanda (changes requested):
The refactor looks solid - the three issues flagged before (counter leak in `update_stream_profile`, wrong cap on credential counter, fingerprint fallback) are all addressed. The test coverage is much more thorough now. However, a few things worth discussing:

**1. The deleted profile test documents a real leak, not correct behavior**

The new `test_release_when_profile_row_deleted` reserves a slot (credential_key = 1), deletes the profile, releases, and asserts `cred_key == 1`. That's the bug, not the fix. When `release_profile_slot` is called with a deleted profile, `_release_server_group_slot_for_profile` is skipped (`if profile:` is False at line 256), so the credential counter is never decremented. This leaks one slot permanently.

The test should either:
- Assert `cred_key == 0` (correct behavior after release), which means the code needs to change to handle the deleted case
- Or remove the test since the deletion edge case is sufficiently rare that it may not be worth defending

If a user deletes a profile while streams are actively using it, that's already a mess - but the counter leak would persist until restart.

**2. Profiles with `max_streams=0` bypass credential enforcement**

`group_has_capacity_for_profile` returns True immediately for `profile.max_streams == 0` (line 166). This is intentional but worth noting clearly: assigning an unlimited profile to a group that has credential limits effectively opts that profile out of cross-account enforcement. 

Example: Group has Account A (unlimited profile) and Account B (max_streams=1) on the same login. Both can stream simultaneously because A's unlimited status skips the credential counter. The provider still sees both connections.

This may be fine by design, but it's a footgun that should probably be documented - either in a code comment or as a limitation in the ServerGroup help_text.

**3. Extra DB hits on the failure logging path in `Channel.get_stream`**

After `reserve_profile_slot` fails, the logging block (lines 731-741) calls `group_has_capacity_for_profile`, which calls `_credential_counter_key`, which calls `get_profile_credential_fingerprint`. For STD/M3U profiles that's a `Stream.objects.filter().first()` DB query plus URL transform. This runs on every failed reserve in the rotation loop.

The information is already implicitly known (reserve returned False), so the logging happens on the hot path doing redundant work. Minor, but on a saturated system with many profiles in a group you'd see extra DB traffic on failures. You could have `reserve_profile_slot` return the reason for failure instead of just the boolean, so the caller doesn't need to re-check

Goldenfreddy0703:
@SergeantPanda — thanks for the follow-up review. Agreed on all three points — pushed fixes below.

---

## 1. Deleted profile release leak (fixed)

You were right: `test_release_when_profile_row_deleted` was documenting the bug (`cred_key == 1`), not correct behavior.

**Fix:** On successful credential reserve, we now store the credential Redis key at `profile_credential_release:{profile_id}`. `release_profile_slot` always releases via that stored key first, so credential counters decrement even when the profile row is gone.

The test now asserts `cred_key == 0` after release.

---

## 2. Unlimited profiles (`max_streams=0`) bypass (documented)

Agreed this is a footgun by design. Added documentation in:

- `connection_pool.py` module docstring + comment on `group_has_capacity_for_profile`
- `ServerGroup.max_streams` model `help_text`
- Server Group form UI description
- M3U account server group dropdown description

Unlimited profiles in a pooled group still stream without participating in cross-account credential enforcement.

---

## 3. Redundant DB work on failure logging (fixed)

`reserve_profile_slot` now returns `(reserved, profile_count, failure_reason)` with `failure_reason` of `"profile_full"` or `"credential_full"` when reserve fails.

`Channel.get_stream` uses that reason for logging instead of re-calling `group_has_capacity_for_profile` / fingerprint resolution on the failure path.

---

All **12** tests in `apps.m3u.tests.test_connection_pool` pass locally (including updated deleted-profile test and new `test_reserve_returns_failure_reason_without_extra_checks`).

Let us know if anything else looks off. Thanks again.

SergeantPanda (approved):
I spoke to you on Discord about this already but I'm going to mention it here too.

I went ahead and made a few changes to clean up the code a bit before we merge here.

I removed max_streams since we weren't really using it in this PR anyway, it was always using the linked profiles limits.

I moved settings out of system settings and into a modal that lives in the m3u manager.

I refactored the table for server groups.

I removed the username from the connection cards, I didn't see any value in them as the profile name is already listed which would tell you what username is being used.

I removed some unnecessary db calls that were only used for logging.

I modified relevant tests and added a couple of new ones.

Overall, I think this is now good to merge.

Thank you!

PR #1271, opened 2026-05-25
docker: split base image into builder and runtime stages

### LINKED ISSUE #1270: [Feature]: Shrink the base Docker image by building it in two stages

Yes. `docker/DispatcharrBase` installs the C/C++/Fortran build toolchain (build-essential, gcc, g++, gfortran, libopenblas-dev, ninja-build and the
`-dev` headers) to compile the virtual environment and the legacy NumPy wheel, then runs `apt-get remove` + `autoremove --purge` to delete it again.

Because the removal runs in a **separate layer** from the install, the bytes are never actually reclaimed: the multi-hundred-MB install layer stays baked into the image and the remove step only writes deletion markers on top (slightly *adding* to the total). The toolchain ships in every published image even though nothing at runtime uses it.

`docker history` on the current base image shows this clearly: 
- `apt-get install ... build-essential gcc g++ gfortran libopenblas-dev ...` → **847 MB** layer
- `apt-get remove ...` → **1.49 MB** layer (deletion markers only; reclaims nothing)

Convert `DispatcharrBase` into a multi-stage build:

* **builder stage**: installs the toolchain, runs `uv sync`, and builds the legacy (`cpu-baseline=none`) NumPy wheel.
* **runtime stage**: same `lscr.io/linuxserver/ffmpeg` base, installs only the runtime libraries (no compilers), and `COPY --from=builder` brings over the prebuilt `/dispatcharrpy` virtual environment and `/opt/numpy-*.whl`.

The set of packages present in the runtime image stays identical to what the image ships today *after* its cleanup step, so there is no behavior change; the compilers simply never enter the final stage in the first place.

**Expected impact:** base image **4.28 GB → 3.51 GB** (~770 MB / ~18% smaller), which carries through to the published application image. No runtime functionality change.

### Additional context

Nothing Extra

kensac:
## Description

Split `docker/DispatcharrBase` into a builder stage (build toolchain, `uv sync`, legacy NumPy wheel build) and a runtime stage that installs only runtime libraries and copies the prebuilt `/dispatcharrpy` venv and `/opt/numpy-*.whl` from the builder.

Previously the toolchain was installed, used, then removed in a separate layer, so its ~850 MB was never reclaimed and shipped in every image. The runtime stage installs the same package set that survived the old cleanup, so runtime contents are unchanged. Base image drops from 4.28 GB to 3.51 GB (~18%). No changes needed to `docker/Dockerfile` or `.github/workflows/base-image.yml`.

## Related Issue

Closes #1270 

## How was it tested?

Built the image locally for `linux/arm64` and ran a smoke test:

- All Python dependencies import; NumPy and Torch operations execute.
- nginx, redis-server, postgresql-17, comskip, vlc, uWSGI 2.0.31, and uv are present.
- Legacy NumPy wheel swap from the entrypoint (`uv pip install /opt/numpy-*.whl`) works.
- Compilers (gcc, g++, gfortran, cc, make, ninja) confirmed absent from the runtime image.

## Checklist

- [x] I understand - line by line - every change in this PR and can explain it if asked
- [x] Backend: migrations are included if any models were changed (N/A, no model changes)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A, no endpoint changes)
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) (N/A, no frontend changes)
- [x] Tests are included for new functionality (N/A, Docker build change)
- [x] Existing tests still pass (N/A, no code changes)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Docker base image now uses a multi-stage build.** The builder stage installs compilers and dev headers (`gcc`, `g++`, `gfortran`, `build-essential`, `libopenblas-dev`, `libpcre3-dev`, `python3.13-dev`, `ninja-build`) to create the virtual environment and compile the legacy NumPy wheel (cpu-baseline=none for old hardware). The final runtime image starts from the same ffmpeg base, copies only the prebuilt venv and wheel from the builder, and installs only the runtime libraries needed at runtime - eliminating compiler binaries from production containers and reducing final image size. — Thanks [@kensac](https://github.com/kensac)
+- **`libpq-dev` removed from both build and runtime stages.** The project uses `psycopg2-binary`, which bundles its own statically linked copy of libpq inside the wheel. No system libpq headers or shared library are needed at build or runtime.
+
 ### Performance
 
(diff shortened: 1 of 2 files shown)


SergeantPanda (commented):
Good find! I will play with this a bit but makes sense to me so far.

SergeantPanda (approved):
This looks good. Thank you!

PR #1255, opened 2026-05-21
perf: reduce intermittent 503s on stats page and API requests

JCBird1012:
## Description

Sometimes when navigating to the Stats page, I would see a bunch of 503 errors from Dispatcharr.

`GET /proxy/ts/status` was being called twice simultaneously on every Stats page mount due to two useEffect hooks both firing on mount. The first one sets up the polling interval (and fetches immediately), and the second that also fetched immediately. 

This doubled the request rate, and each request was expensive: `get_basic_channel_info()` was making `O(channels × clients)` Redis round-trips due to 4 separate `hget` calls per connected client and a redundant `hget` for TOTAL_BYTES (we already had fetched that in the preceding `hgetall` result).

Under load this could hold a uWSGI worker long enough for concurrent requests from other pages to time out, producing a 503.

**Changes:**
- **Stats.jsx**: Removed the redundant mount-time useEffect that duplicated the initial fetch already performed by the polling effect.
- **channel_status.py**: Replaced 4× hget per client with a single hmget for the 4 fields actually needed (user_agent, ip_address, connected_at, user_id), avoiding the 2 unused fields (last_active, worker_id).
- **nginx.conf**: Added uwsgi_read_timeout 300s and uwsgi_send_timeout 300s to the main / location. The global proxy_read_timeout only applies to proxy_pass directives, not uwsgi_pass - the `/` location was silently using nginx's 60s default while the `/proxy/` location already had explicit 300s timeouts.

## How was it tested?

Navigated to the Stats page repeatedly (like 200 times, seriously) with ~10ish active streams running and monitored the browser Network tab to confirm `GET /proxy/ts/status` is now called once on mount instead of twice. I also verified stats continue to update on the configured polling interval. It's tricky because the original 503s would happen very intermittently, so I wanted to be sure the issue was really solved.

## Checklist

Please check every item before marking this PR as ready for review. Unchecked items will be flagged during review.

- [X] I have read the [CONTRIBUTING.md](../blob/dev/CONTRIBUTING.md) in full
- [X] I agree to the [Contributor License Agreement](../blob/dev/CONTRIBUTING.md#contributor-license-agreement)
- [X] I understand — line by line — every change in this PR and can explain it if asked
- [X] This PR targets the `dev` branch
- [X] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`)
- [X] Tests are included/modified for new functionality
- [X] Existing tests still pass
[...]

PR #1272, opened 2026-05-25
perf: replace per-request DB connections with geventpool

JCBird1012:
## Description

(This is sorta a re-hash of #1255, but I'll quickly re-describe)...

So, occasionally, when navigating to the Stats page (or some other pages), Dispatcharr would either return a 503 or a 500 - this would happen very infrequently and it made it difficult to get a grasp for the conditions that would cause it. After investigating a little deeper after addressing #1255, I finally got a [stack trace](https://github.com/Dispatcharr/Dispatcharr/pull/1255#issuecomment-4529796572) that revealed that the underlying issue was that gevent was making per-greenlet connections to the DB - for something like logos, that would very quickly exhaust the default Postgres connection limit of 100. From a performance standpoint as well, that would mean gevent would have to open a new connection for each greenlet request (CONN_MAX_AGE = 0) - that was a full SSL handshake + auth round-trip on every DB query.

image

A pool keeps connections warm and hands them out to greenlets on demand. Under the load pattern that a typical Dispatcharr user creates while navigating the web interface, that's a pretty hefty reduction in per-request latency.

## Related Issue

Closes #1255 (sorta) - the underlying root cause for that issue was DB connection exhaustion.

## How was it tested?

On paper, this shouldn't regress anything - the `gevent` and `psycogreen` infrastructure is already correctly configured - this is a really just a swap of the DB backend. I tested this myself by running this branch as my production Dispatcharr instance for a few hours and didn't experience any issues. Of course, this should definitely live on `dev` for a hot second just to be sure (as is the Panda way).

## Checklist


JCBird1012:
Also side note - there was a discussion about adding native connection pooling to Django last year - https://forum.djangoproject.com/t/proposal-exploring-native-database-connection-pooling-support-in-django-orm/40514 - but it seems as of now the best way to get pooling is either via pgbouncer or a lib like [django-db-geventpool](https://github.com/jneight/django-db-geventpool) - which this PR does.

SergeantPanda (commented):
I will play with this for sure. I've been highlly considering doing exactly this so I expect to merge this. Just need to test a few things first.

Thank you!

SergeantPanda (approved):
Hey, so, I kind of went a bit overboard probably on this one but I went down a rabiit hole...

A few things.
1) I could not get this to work at all in its current state, it seems the ENGINE was missing the `_psycopg2` at the end. Adding that allowed it to boot. Not sure how you were able to get it.
2) I modifed the values you had set (which I think were probably default and generally OK). Considering up to 6 celery workers and by default 4 uwsgi workers, that was a lot of connections to keep open. It may still need some tweaking but this is a good in between I think (8 and 3)
3) I took this opportunity to update to psycopg3 as it seems django is going to be deprecating psycopg2 soon and 3 has some good benefits where we can drop other dependencies.

Thank you for the work you put into this. It will be a good change overall.

SergeantPanda (commented):
I just noticed now that this targets the MAIN branch. It should target DEV in the future, please be mindful of that (it's in the checklist)

PR #1274, opened 2026-05-26
Tests/frontend unit tests

nick4810:
## Description

Updated additional components within `components/forms/` for readability and added unit tests.

- frontend/src/components/forms/ProgramRecordingModal.jsx
- frontend/src/components/forms/Recording.jsx
- frontend/src/components/forms/RecordingDetailsModal.jsx
- frontend/src/components/forms/RecurringRuleModal.jsx
- frontend/src/components/forms/ScheduleInput.jsx
- frontend/src/components/forms/SeriesRecordingModal.jsx
- frontend/src/components/forms/SeriesRuleEditorModal.jsx
- frontend/src/components/forms/Stream.jsx
- frontend/src/components/forms/StreamProfile.jsx
- frontend/src/components/forms/SuperuserForm.jsx
- frontend/src/components/forms/User.jsx
- frontend/src/components/forms/UserAgent.jsx
- frontend/src/components/forms/VODCategoryFilter.jsx

Summary of changes:

- Added unit tests

## Related Issue

NA

## How was it tested?

NA

## Checklist

- [x]  I have not reformatted or refactored code outside the scope of this change


frontend/src/components/forms/User.jsx
```diff
@@ -248,13 +133,13 @@ const User = ({ user = null, isOpen, onClose }) => {
         payload.user_id = user.id;
       }
 
-      const resp = await API.generateApiKey(payload);
+      const resp = await generateApiKey(payload);
       const newKey = resp && (resp.key || resp.raw_key);
       if (newKey) {
         setGeneratedKey(newKey);
-        setUserAPIKey(newKey);
+        console.log(setUserAPIKey(newKey));
```
SergeantPanda: Mind removing this? I'm assuming you just had it here for testing?

frontend/src/components/forms/RecordingDetailsModal.jsx
```diff
@@ -599,7 +601,7 @@ const RecordingDetailsModal = ({
         title: { color: 'white' },
       }}
     >
-      {isSeriesGroup ? Series() : Movie()}
+      {isSeriesGroup ? <Series /> : <Movie />}
```
SergeantPanda: Both of these are functions defined inside the component body on every render. Switching from calling them as functions (Series() / Movie()) to using them as JSX elements (<Series /> / <Movie />) means React sees a brand-new component type on every render, so it will unmount and remount that entire subtree every single time the parent re-renders. The original function-call approach was actually correct here. Should be reverted to {isSeriesGroup ? Series() : Movie()}.

 Please correct me if I'm wrong here.

frontend/src/pages/__tests__/guideUtils.test.js
```diff
@@ -1277,32 +1065,44 @@ describe('guideUtils', () => {
...
     });
   });
 
+  // ── evaluateSeriesRulesByTvgId ────────────────────────────────────────────
+
   describe('evaluateSeriesRulesByTvgId', () => {
     it('should evaluate series rules via API', async () => {
+      vi.mocked(API.evaluateSeriesRules).mockResolvedValue({});
+
       await guideUtils.evaluateSeriesRulesByTvgId('tvg-1');
```
SergeantPanda, line 1094: This seems to be a duplicate of line 810 unless I'm missing something.

frontend/src/components/forms/__tests__/RecurringRuleModal.test.jsx
```diff
@@ -0,0 +1,969 @@
...
+
+  describe('form initialization', () => {
+    it('calls getFormDefaults with rule when opened', () => {
+      render(<RecurringRuleModal {...defaultProps()} />);
+      expect(RecurringRuleModalUtils.getFormDefaults).toHaveBeenCalledWith(
+        makeRule()
+      );
+    });
+
+    it('calls form.setValues with getFormDefaults result', () => {
```
SergeantPanda, line 585: This always passes regardless of what the component does - if formMock is undefined the fallback creates a fresh spy, and a spy is always defined. It's not actually testing anything. Should assert that setValues was called with the expected values.

SergeantPanda (changes requested):
Thank you again for the PR Nick!

I did find a few minor things to fix in this one, or at least clarify for me if I am misunderstanding before merge.

If you could please look at the 4 findings above and fix/report back I'd appreciate it.

Thanks again!

frontend/src/components/forms/RecordingDetailsModal.jsx
nick4810: Yeah I think `{isSeriesGroup ? Series() : Movie()}` is ok here, but `<EpisodeRow />` has a similar issue. I'd recommend moving it to outside of the parent.

nick4810:
@SergeantPanda Thanks for the feedback! I addressed your findings so this should be ready for review again.

frontend/src/components/forms/RecordingDetailsModal.jsx
SergeantPanda: Sounds good. I will look into it.

SergeantPanda:
Thank you Nick! I'll look at it tomorrow.

SergeantPanda (approved):
Everything looks good. Thanks again!

PR #1282, opened 2026-05-28
fix(output): preserve non-standard port in M3U URLs behind reverse proxy

### LINKED ISSUE #1267: [Bug]: Ver 0.25.1 ports missing from m3u links
### Describe the bug

After Upgrade it seems that the M3U is not generating the channel links with the port for dispatcharr.
it looks like the specific port has been stripped which is causing streams not to load.

You can download the m3u from the homepage and see that the channel list does not have any ports for the links generated.

### Dispatcharr Version

0.25.1

### Docker Version

Proxmox LXC

### Client Information

nemesbak:
## Description

`get_host_and_port()` in `apps/output/views.py` has an early-exit that drops the port number whenever a reverse-proxy header (`X-Forwarded-Proto` or `X-Forwarded-For`) is present, regardless of whether the port is standard or not. When Dispatcharr runs on a non-standard port (e.g. 9191) behind nginx or Caddy, every URL in the generated M3U playlist is missing the port, causing IPTV players to connect to port 80/443 instead.

The fix reads `HTTP_SERVER_PORT` (or `SERVER_PORT`) before returning when behind a proxy, and only omits the port when it matches the standard port for the scheme (80 for http, 443 for https).

## Related Issue

Closes #1267

## How was it tested?

- Dispatcharr running on port 9191 behind nginx with `X-Forwarded-Proto: https` and `X-Forwarded-For` headers set. Before the fix: M3U URLs had no port. After the fix: URLs contain `:9191`.
- Verified that standard-port configs (80/443) still omit the port correctly.
- Verified direct access without a proxy is unaffected.

## Checklist

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -3159,13 +3159,19 @@ def get_host_and_port(request):
         # Omit standard ports from URLs
         return host, None if port == standard_port else port
 
-    # 4. Check if we're behind a reverse proxy (X-Forwarded-Proto or X-Forwarded-For present)
-    # If so, assume standard port for the scheme (don't trust SERVER_PORT in this case)
-    if request.META.get("HTTP_X_FORWARDED_PROTO") or request.META.get("HTTP_X_FORWARDED_FOR"):
+    # 4. Check if we're behind a reverse proxy (X-Forwarded-Proto or X-Forwarded-For present).
+    # Still check SERVER_PORT: a non-standard port (e.g. 9191) must be preserved even behind
+    # a proxy so that generated M3U URLs remain reachable. Only skip SERVER_PORT when it
+    # matches the standard port for the scheme (80/443), which means the proxy handles TLS
+    # termination and the port can be safely omitted.
+    behind_proxy = request.META.get("HTTP_X_FORWARDED_PROTO") or request.META.get("HTTP_X_FORWARDED_FOR")
+    port = request.META.get("SERVER_PORT")
+    if behind_proxy:
+        if port and port != standard_port:
+            return host, port
         return host, None
 
     # 5. Try SERVER_PORT from META (only if NOT behind reverse proxy)
-    port = request.META.get("SERVER_PORT")
     if port:
         # Omit standard ports from URLs
         return host, None if port == standard_port else port


SergeantPanda:
This *may* be valid but we can't look at this without you filling out the PR form.

nemesbak:
Updated the PR description to follow the template and changed the base branch from `main` to `dev`. Apologies for missing those on the initial submission.

SergeantPanda:
Thanks for the detailed PR and for diagnosing the root cause correctly. There is a real bug here and the intent is right, but the fix has a regression that affects a very common deployment pattern. Here's the full picture:

**Why branch #4 is reached (the actual trigger)**

Our Docker nginx.conf sets `X-Forwarded-Host`, `X-Forwarded-Port`, and `X-Forwarded-Proto` using `proxy_set_header` in the server block. However, `proxy_set_header` only applies to `proxy_pass` upstreams and is silently ignored by `uwsgi_pass`. All three main application locations (`/`, `/hdhr`, `/proxy/`) use `uwsgi_pass unix:/app/uwsgi.sock` with only `include uwsgi_params;`. That means **our own nginx never injects any X-Forwarded headers into the WSGI environ**. The `HTTP_X_FORWARDED_PROTO` and `HTTP_X_FORWARDED_FOR` variables in WSGI are only present when a separate outer proxy (Traefik, Caddy, a user's own nginx, etc.) sent them. The standard `uwsgi_params` file *does* set `SERVER_PORT $server_port;`, so that value is nginx's listening port (e.g. 9191).

**The regression the PR introduces**

The very common outer-proxy pattern is:

```
Client → Traefik/Caddy/user-nginx (HTTPS :443) → Dispatcharr Docker nginx (:9191) → uWSGI
```

In this scenario:
- Outer proxy sets `X-Forwarded-Proto: https`, `X-Forwarded-For: ...` (no X-Forwarded-Port or X-Forwarded-Host, both are optional and rarely set by default).
- Host header = `example.com` (no port, since 443 is standard HTTPS).
- `SERVER_PORT` = `"9191"` (from `uwsgi_params $server_port;`, the internal port).
- `standard_port` = `"443"` (derived from `X-Forwarded-Proto: https`).

Branch #4 is reached. The PR returns `(example.com, "9191")`, producing `https://example.com:9191/...` in the M3U. That internal nginx port is not publicly reachable on that scheme. Before the PR, this returned `(example.com, None)`, producing `https://example.com/...`, which was correct.

This breaks any user running Dispatcharr on a non-standard internal port behind a TLS-terminating reverse proxy, which is an extremely common home-server pattern.

**The right fix**

The root cause is that our nginx doesn't get the correct headers into the WSGI environ. Replacing `proxy_set_header` with `uwsgi_param` for the uwsgi locations would make branches #1 and #3 handle everything correctly and eliminate the need for the fragile `SERVER_PORT` heuristic entirely. Adding these to nginx.conf (either in the server block or in each uwsgi location) would do it:

```nginx
uwsgi_param HTTP_X_FORWARDED_HOST $host:$server_port;
uwsgi_param HTTP_X_FORWARDED_PORT $server_port;
uwsgi_param HTTP_X_FORWARDED_PROTO $scheme;
uwsgi_param HTTP_X_FORWARDED_FOR $proxy_add_x_forwarded_for;
```

The debian_install.sh nginx config also needs `X-Forwarded-Host` and `X-Forwarded-Port` added to the main `location /` block alongside the existing `proxy_params`, since it uses `proxy_pass` to gunicorn (where `proxy_set_header` does work).

With those changes, branch #1 catches the direct-Docker case (`X-Forwarded-Host: myhost:9191`), branch #3 catches the outer-proxy case (`X-Forwarded-Port: 443` from the outer proxy), and branch #4 becomes a true last-resort fallback. The `SERVER_PORT`-based heuristic in branch #4 can then be reverted to the original `return host, None`.

PR #1286, opened 2026-05-28
fix(frontend): show Clear Overrides button when override value matches provider

### LINKED ISSUE #1276: [Bug]: For auto synced channels, the clear override buttons disappear from the edit channel screen when the data in the overridden field is identical to the default data in the auto synced field.
### Describe the bug

For auto synced channels, the clear override buttons disappear from the edit channel screen when the data in the overridden field is identical to data in the default auto synced field. But the yellow pencil indicating that there is an override in place remains.

I would expect to be able to clear the override, with no actual change in the field since it is all the same, then the pencil would disappear.

1) Set up an auto sync group in m3u manager
2) Go to your Channels 
3) Select a channel from the auto synced group and override the group to something different from the original group lets say "apple". The override marker will appear.
4) Go to M3U manager and then to the M3U Group Filter & Auto Channel Sync menu
5) Go to the group that you set up for auto sync and hit the gear icon to configure it
6) Set the Override channel Group to "apple" matching what you did on the channel list.
7) You will loose the ability to reset the override you originally did

### Dispatcharr Version

v0.25.1

### Docker Version

Docker version 29.4.3, build 055a478

services:
  dispatcharr:
    # build:
    #   context: .
    #   dockerfile: Dockerfile
    restart: unless-stopped
    container_name: dispatcharr
    ports:
      - 9191:9191
    volumes:
      - dispatcharr_data:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
      #- USE_LEGACY_NUMPY=true
    #
    #cap_add:
    #  - SYS_NICE
    #devices:
    #deploy:
    #  resources:
    #      reservations:
    #          devices:
    #                count: all

volumes:
  dispatcharr_data:
    external: true

### Client Information

Emby Server Version 4.9.5.0

nemesbak:
## Description

The **Clear All Overrides** button disappears from the channel edit form when the stored override value happens to be identical to the current provider value, even though the override is still active. The yellow pencil in the channel list correctly shows the override is present, but there is no way to remove it.

**Steps to reproduce:**

1. Auto-synced channel — set group override to "apple"

2. In M3U Group Filter, set the auto-sync group override to "apple" (matching)

3. Re-open channel edit → Clear button is gone, pencil remains

**Root cause:**

`overriddenFieldLabels` was computed with `isFormFieldOverridden()`, which returns `false` when the current form value equals the provider value — regardless of what is stored in `channel.override`. When the two values coincidentally match, the function reports "no override" and the button is hidden.

`listOverriddenFields()` already exists and does the right thing: it inspects `channel.override[field]` directly and returns labels for every field that has a non-null stored override.

**Fix:**

Replace the `isFormFieldOverridden()` call in the `overriddenFieldLabels` memo with `listOverriddenFields(channel)`. The button now reflects the actual database state rather than a value comparison.

## Related Issue

Closes #1276

## How was it tested?

- [ ] Set group override to the same value as the provider group → Clear button remains visible

- [ ] Clear the override → pencil disappears, button disappears ✓

- [ ] Overriding to a different value than provider → button still appears ✓

- [ ] Channel with no overrides → button hidden ✓

## Checklist

### DIFF
--- a/frontend/src/components/forms/Channel.jsx
+++ b/frontend/src/components/forms/Channel.jsx
@@ -52,6 +52,7 @@ import {
   getProviderHint,
   handleEpgUpdate,
   isFormFieldOverridden,
+  listOverriddenFields,
   matchChannelEpg,
   OVERRIDABLE_FIELDS,
   OVERRIDE_FIELD_LABELS,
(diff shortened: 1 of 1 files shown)


SergeantPanda:
This causes a new bug where the channel table does not show the override.

PR #1287, opened 2026-05-28
fix(vod): import director, actors and release_date from XC provider during sync

### LINKED ISSUE #1228: [Bug]: VOD metadata fields `director`, `cast`/`actors`, and `release_date` are not imported from XC provider during sync
### Describe the bug

When syncing a VOD library from an Xtream Codes provider, Dispatcharr imports only a subset of the metadata fields returned by `get_vod_streams`. The fields `director`, `cast` (also known as `actors`), and `release_date` are present in the provider's response but are never written to the database. As a result, Dispatcharr's own `get_vod_info` endpoint always returns empty values for these fields, even though the upstream XC provider supplies them correctly.

1. Set up an XC provider whose `get_vod_streams` response includes `director`, `cast`, and `release_date` for each movie
2. Add the provider in Dispatcharr and trigger a sync
3. Call Dispatcharr's own XC endpoint after the sync:
   `GET /player_api.php?username=USER&password=PASS&action=get_vod_info&vod_id=<id>`
4. Observe that `director`, `actors`/`cast`, and `release_date` are empty, even though they were present in the provider's `get_vod_streams` response

### Provider response (get_vod_streams) — correct data supplied

```json
{
  "stream_id": 1,
  "name": "13 Geister",
  "plot": "Der verwitwete Familienvater ...",
  "genre": "Horror, Thriller",
  "director": "Steve Beck",
  "cast": "Tony Shalhoub, Embeth Davidtz, Matthew Lillard",
  "release_date": "2001-10-25",
  "tmdb_id": "9378",
  "imdb_id": "tt0245674",
  "year": 2001,
  "rating": 1.9
}
```

### Dispatcharr get_vod_info response after sync — fields missing

```json
{
  "info": {
    "name": "13 Geister",
    "description": "Der verwitwete Familienvater ...",
    "plot": "Der verwitwete Familienvater ...",
    "genre": "Horror, Thriller",
    "director": "",
    "actors": "",
    "cast": "",
    "release_date": "",
    "tmdb_id": "9378",
    "imdb_id": "tt0245674",
    "year": 2001,
    "rating": "1.9"
  }
}
```

### Root cause

Dispatcharr reads `plot`, `genre`, `tmdb_id`, `imdb_id`, and `year` from `get_vod_streams` and writes them to the database during sync. However, `director`, `cast`/`actors`, and `release_date` are not mapped and never persisted.

Since Dispatcharr's own `get_vod_info` endpoint builds its response entirely from the local database — without forwarding the request to the upstream provider — these fields remain permanently empty after a sync.

### Expected behavior

All metadata fields provided by the XC provider in `get_vod_streams` should be imported into the database during sync. At minimum, `director`, `cast`/`actors`, and `release_date` should be persisted and returned correctly by Dispatcharr's `get_vod_info` endpoint.

### Dispatcharr Version

v0.24.0

### Docker Version

Proxmox VE Helper Script (bare-metal)

Not using.

### Client Information

Televizo 1.9.9.22-g on Android.

nemesbak:
## Description

After syncing a VOD library from an XC provider, `director`, `cast`/`actors`, and `release_date` are always empty in Dispatcharr's own `get_vod_info` / `provider-info` responses, even though the upstream `get_vod_streams` response contains them.

**Root cause:**

`process_movie_batch()` in `apps/vod/tasks.py` built `custom_properties` with only `trailer`, discarding the other fields the provider supplied.

Additionally, the update path for existing movies replaced `custom_properties` wholesale, wiping keys like `detailed_info` written by the advanced refresh task.

**Fix:**

1. Extract `director`, `actors`/`cast`, and `release_date` from `movie_data` during the batch-import phase and store them in `custom_properties`.

2. Change the update path to **merge** incoming `custom_properties` into the existing dict (`{**existing, **incoming}`) instead of overwriting it, preserving advanced-refresh data across subsequent basic syncs.

## Related Issue

Closes #1228

## How was it tested?

- [ ] Add XC provider with movies that include `director`, `cast`, `release_date` in `get_vod_streams`

- [ ] Trigger sync → call `GET /player_api.php?action=get_vod_info&vod_id=<id>` → fields are populated ✓

- [ ] Trigger advanced refresh, then basic sync again → `detailed_info` key is preserved ✓

## Checklist

### DIFF
--- a/apps/vod/tasks.py
+++ b/apps/vod/tasks.py
@@ -439,6 +439,28 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
             trailer = extract_string_from_array_or_string(trailer_raw) if trailer_raw else None
             logo_url = movie_data.get('stream_icon') or ''
 
+            # Extract optional metadata fields supplied by the XC provider in
+            # get_vod_streams. Stored in custom_properties so that the
+            # provider-info endpoint (and the XC vod_info passthrough) can
+            # return them without a separate per-movie API call.
+            director = extract_string_from_array_or_string(
+                movie_data.get('director') or ''
+            )
+            actors = extract_string_from_array_or_string(
+                movie_data.get('actors') or movie_data.get('cast') or ''
+            )
+            release_date = movie_data.get('release_date') or movie_data.get('releasedate') or ''
+
+            custom_props = {}
+            if trailer:
+                custom_props['trailer'] = trailer
+            if director:
+                custom_props['director'] = director
+            if actors:
+                custom_props['actors'] = actors
+            if release_date:
+                custom_props['release_date'] = release_date
+
             movie_props = {
                 'name': name,
                 'year': year,
@@ -448,7 +470,7 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
                 'rating': rating,
                 'genre': genre,
                 'duration_secs': duration_secs,
-                'custom_properties': {'trailer': trailer} if trailer else None,
+                'custom_properties': custom_props or None,
             }
 
             movie_keys[movie_key] = {
@@ -548,8 +570,14 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
 
             for field, value in movie_props.items():
                 if field == 'custom_properties':
-                    if value != movie.custom_properties:
-                        movie.custom_properties = value
+                    # Merge incoming props into the existing dict so that
+                    # detailed_info and other keys written by advanced refresh
+                    # are not lost when a basic sync runs afterwards.
+                    existing_cp = movie.custom_properties or {}
+                    incoming_cp = value or {}
+                    merged = {**existing_cp, **incoming_cp}
+                    if merged != existing_cp:
+                        movie.custom_properties = merged
                         updated = True
                 elif getattr(movie, field) != value:
                     setattr(movie, field, value)
--- a/apps/vod/tasks.py
+++ b/apps/vod/tasks.py
@@ -453,7 +453,7 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
 
             custom_props = {}
             if trailer:
-                custom_props['trailer'] = trailer
+                custom_props['youtube_trailer'] = trailer
             if director:
                 custom_props['director'] = director
             if actors:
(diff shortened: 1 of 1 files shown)


SergeantPanda (changes requested):
Thanks for the PR - the batch processing architecture, chunking, and category management are solid. A few things to address before merge:

---

**1. `custom_properties` merge order silently downgrades advanced-refresh data (regression)**

The merge at line ~573 is `{**existing_cp, **incoming_cp}`, meaning basic-sync values overwrite existing ones. If `refresh_movie_advanced_data` has already stored richer `director`/`actors` from `get_vod_info`, the next basic sync will replace them with the coarser `get_vod_streams` values.

**2. New fields block advanced refresh from ever updating them (regression)**

`refresh_movie_advanced_data` uses `should_update_field()` which only writes when the existing value is `None`/empty. Now that basic sync populates `director` and `actors` on every run, those fields will never be upgraded by advanced refresh - even if the provider's bulk list has a truncated cast string.

Both issues have the same fix: only write these fields into `incoming_cp` when the movie has no existing value for them:

```python
if director and not existing_cp.get('director'):
    custom_props['director'] = director
if actors and not existing_cp.get('actors'):
    custom_props['actors'] = actors
if release_date and not existing_cp.get('release_date'):
    custom_props['release_date'] = release_date
```

**3. `trailer` key name mismatch (pre-existing, but worth fixing here)**

Basic sync writes `custom_props['trailer']`, but `api_views.py` reads `custom_properties.get('youtube_trailer')` and advanced refresh also writes `'youtube_trailer'`. The `'trailer'` key is never consumed, so trailers from `get_vod_streams` are silently discarded until per-movie advanced refresh runs. Change `custom_props['trailer'] = trailer` to `custom_props['youtube_trailer'] = trailer`.

nemesbak:
Hi SergeantPanda,

Applied all three fixes from your review:

1. **`youtube_trailer` key** — changed `custom_props['trailer']` to `custom_props['youtube_trailer']` to match what `api_views.py` and `refresh_movie_advanced_data` expect. The old key was silently discarded.

2. **Preserve advanced-refresh metadata** — `director`, `actors` and `release_date` are now only written to `custom_properties` when the field is currently empty. This means a basic sync will no longer overwrite richer data that a previous `refresh_movie_advanced_data` run already stored.

3. **Empty fields still get populated** — if no advanced refresh has run yet (field is `None`/empty), basic sync still fills them in as before.

Before pushing, I ran three targeted tests directly on a live Dispatcharr instance to verify each scenario:
- Rich advanced-refresh data is preserved when basic sync runs again ✓
- Empty fields are correctly populated by basic sync ✓
- `youtube_trailer` key propagates correctly and the orphaned `trailer` key is gone ✓

Thanks again for the detailed review.

SergeantPanda (approved):
Thank you for making those fixes. I did condense the comments a bit and while I was reading through the code found a couple small things to improve. (unrelated to your changes). I also had to pair this with a change I've already pushed to dev which will actually resolve the linked issue. 

This PR adds the data into the database during a vod scan but didn't output it during xc get_vod_streams output.

PR #1289, opened 2026-05-28
fix(dvr): pass --threads=1 to comskip CLI to prevent SIGABRT in Docker

### LINKED ISSUE #1251: [Bug]: Docker comskip fails on UI invocation
### Describe the bug

Triggering comskip processing on a completed recording from the DVR UI results in a generic error. The API call succeeds (HTTP 200, `{"success": true, "queued": true}`), but the background Celery task fails silently and stores an error status on the recording.

The default ini at [volume_path]/comskip.ini sets thread_count=0, which tells comskip to auto-detect and use all available CPU threads. On this host, the multi-threaded ffmpeg codec/decoder initialization raises SIGABRT immediately after startup.

Setting thread_count=1 in the ini file does not fix the crash. The ini-based thread setting is applied after the ffmpeg initialization that triggers the abort, so it has no effect. The --threads=1 flag must be present on the command line.

  1. Have a completed recording in the DVR
  2. Trigger "Run comskip" from the UI action menu
  3. UI shows a generic error; no comskip output is produced

### Dispatcharr Version

v0.24.0

### Docker Version

Docker-ce 29.4.3

```yaml
  dispatcharr:
    # build:
    #   context: .
    #   dockerfile: Dockerfile
    restart: unless-stopped
    container_name: dispatcharr
    dns:
      - 1.1.1.1
      - 9.9.9.9
    ports:
      - 9191:9191
    volumes:
      - /media/appdata/dispatcharr/data:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
      - PUID=1000 # Replace with your actual UID
      - PGID=1000 # Replace with your actual GID
      #- USE_LEGACY_NUMPY=true
    #
    #cap_add:
    #  - SYS_NICE
    devices:
      - /dev/dri:/dev/dri  # For Intel/AMD GPU acceleration (VA-API)
    #deploy:
    #  resources:
    #      reservations:
    #          devices:
    #                count: all


### Client Information

Using web browser

nemesbak:
## Description

Running comskip from the DVR UI succeeds (HTTP 200) but the Celery task fails silently. The root cause is that comskip's default `thread_count=0` (auto-detect all CPU cores) triggers SIGABRT during multi-threaded ffmpeg decoder initialisation inside the Docker container. Setting `thread_count=1` in the INI file does not help because the INI is applied after the ffmpeg init that causes the abort.

**Fix:**

Pass `--threads=<n>` on the comskip command line so it takes effect before ffmpeg initialisation. A new `comskip_threads` DVR setting (default: **1**) controls the value:

- `1` (default) — safe single-threaded mode, fixes the Docker crash

- `0` — omits the flag entirely, letting comskip auto-detect (useful on bare-metal with many cores)

- any other positive integer — passed directly to `--threads=N`

## Related Issue

Closes #1251

## How was it tested?

- [ ] Trigger "Run comskip" on a completed recording in Docker → completes without SIGABRT ✓

- [ ] Set `comskip_threads=0` in DVR settings → `--threads` flag is omitted, comskip auto-detects ✓

- [ ] Bare-metal install with `comskip_threads=4` → `--threads=4` passed to CLI ✓

## Checklist

### DIFF
--- a/apps/channels/tasks.py
+++ b/apps/channels/tasks.py
@@ -3176,9 +3176,17 @@ def comskip_process_recording(recording_id: int):
     try:
         comskip_mode = CoreSettings.get_dvr_comskip_mode()
         hw_accel = CoreSettings.get_dvr_comskip_hw_accel()
+        threads = CoreSettings.get_dvr_comskip_threads()
         cmd = [comskip_bin, "--output", os.path.dirname(file_path)]
         if hw_accel != "none":
             cmd.insert(1, f"--{hw_accel}")
+        # Pass --threads on the CLI so it takes effect before ffmpeg decoder
+        # initialisation. The INI-based thread_count=0 setting is applied too
+        # late and causes SIGABRT on multi-threaded builds inside Docker.
+        # Default is 1 (safe); set comskip_threads=0 in DVR settings to let
+        # comskip auto-detect (useful on bare-metal with many cores).
+        if threads != 0:
+            cmd.append(f"--threads={threads}")
         # Prefer user-specified INI, fall back to known defaults
         ini_candidates = []
         try:
--- a/core/models.py
+++ b/core/models.py
@@ -315,6 +315,7 @@ class CoreSettings(models.Model):
             "comskip_custom_path": "",
             "comskip_mode": "cut",
             "comskip_hw_accel": "none",
+            "comskip_threads": 1,
             "pre_offset_minutes": 0,
             "post_offset_minutes": 0,
             "series_rules": [],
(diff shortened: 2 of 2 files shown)


SergeantPanda:
You may be correct the it's safer to use threads=1 for comskip, that being said, this PR implements a lot of overhead for no noticeable gain.

There is no frontend way for the user to modify the setting yet it still reads from settings and defaults to 1? I don't understand.

I'm going to close this and if you'd like to submit a new one that allows the user to set the number of threads that would be acceptable. We'd want to figure out what an acceptable default value is for most users though.

PR #1292, opened 2026-05-28
fix(proxy): add null_ts_keepalive setting to suppress init keepalive TS packets

### LINKED ISSUE #1280: [Bug]: null TS Packet causes playback issues on Emby
### Describe the bug

commit c68249c changed generator.py to send a null TS packet while waiting for the stream to load. Emby tries to probe this and fails so it creates some weird behaviour such as forcing a deinterlace transcode or failing to playback entirely. If you change the code in generator.py from

```python
if time.time() - last_keepalive >= keepalive_interval:
    logger.debug(f"[{self.client_id}] Sending keepalive during initialization")
    keepalive_data = create_ts_packet('null')
    yield keepalive_data
    self.bytes_sent += len(keepalive_data)
    last_keepalive = time.time()
else:
    gevent.sleep(0.1)
```
to just simply
```python
gevent.sleep(0.1)
```
then streams play correctly since Emby probes a TS packet containing stream data. I am unsure if users are experiencing this issue in other clients (maybe Jellyfin since it's a fork of Emby).

This behaviour does not come through on every stream but it does on a majority of them. The only stream profile that bypassed this issue was Streamlink.

Obviously changing this removes the intended functionality the commit gave so if there is a toggle to use this functionality that would be great.

1. Play a stream in Emby, if you have an NZ ip then this one will work: https://i.mjh.nz/.r/te-reo.m3u8
2. View the logs in Emby and note the failed ffprobe log. You'll also see a forced deinterlace transcode

### Dispatcharr Version

v0.25.1

### Docker Version

n/a - using Unraid.

### Client Information

Emby

nemesbak:
## Description

Null TS packets sent during channel initialization confuse clients that probe initial TS data for codec/format detection. Emby interprets the null packet and either forces an unwanted deinterlace transcode or fails playback entirely. The issue does not affect Streamlink profiles because they don't go through the null-packet keepalive path.

**Root cause:**

`_wait_for_initialization()` in `generator.py` sends a null TS packet every 0.5 s while waiting for the channel to become active. This was added to keep the HTTP connection alive during slow-starting streams, but null TS content is technically valid MPEG-TS and gets picked up by Emby's ffprobe during codec detection, producing incorrect stream metadata.

**Fix:**

Add `null_ts_keepalive` (default: `true`) to Proxy Settings. When set to `false`, the timer still advances (maintaining the 0.1 s sleep cadence) but no packet is yielded — the HTTP connection stays open without sending probeble TS data.

**Users experiencing the Emby deinterlace or playback issue should set `null_ts_keepalive=false` in Proxy Settings.**

## Related Issue

Closes #1280

## How was it tested?

- [ ] `null_ts_keepalive=true` (default): keepalive packets sent, existing behaviour preserved ✓

- [ ] `null_ts_keepalive=false`: no packets sent during init, Emby probes real stream data ✓

- [ ] Stream still starts correctly with keepalive disabled ✓

## Checklist

### DIFF
--- a/apps/proxy/live_proxy/output/ts/generator.py
+++ b/apps/proxy/live_proxy/output/ts/generator.py
@@ -169,12 +169,18 @@ class StreamGenerator:
                     yield create_ts_packet('error', "Error: Channel is stopping")
                     return False
 
-            # Send null packets to keep the connection alive during initialization.
+            # Send null TS packets to keep the connection alive during
+            # initialization. This can be disabled via the proxy setting
+            # null_ts_keepalive=False for clients (e.g. Emby) that
+            # misinterpret the null packet during codec probing and force
+            # an unwanted transcode.
             if time.time() - last_keepalive >= keepalive_interval:
-                logger.debug(f"[{self.client_id}] Sending keepalive during initialization")
-                keepalive_data = create_ts_packet('null')
-                yield keepalive_data
-                self.bytes_sent += len(keepalive_data)
+                from core.models import CoreSettings
+                if CoreSettings.get_null_ts_keepalive():
+                    logger.debug(f"[{self.client_id}] Sending keepalive during initialization")
+                    keepalive_data = create_ts_packet('null')
+                    yield keepalive_data
+                    self.bytes_sent += len(keepalive_data)
                 last_keepalive = time.time()
             else:
                 gevent.sleep(0.1)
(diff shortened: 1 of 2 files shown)


SergeantPanda:
This is not valid. I would need to see more evidence that this is true to entertain it.

Also, importing on every tick is less than ideal in the stream generator. The settings used has no user facing way to change it.

PR #1293, opened 2026-05-28
fix(dvr): recover from corrupt AAC frames during HLS→MKV finalization

### LINKED ISSUE #1225: [Bug]: DVR recording finalization silently truncates output MKV on corrupt AAC frame; MP4 fallback also fails silently
### Describe the bug

When a DVR recording is stopped early\*, Dispatcharr's HLS-->MKV finalization fails silently if any HLS segment contains a corrupt AAC audio frame. The MP4-intermediate fallback path that is supposed to handle concat failures also fails silently; no intermediate file is produced and no further log entries appear after the initial fallback warning.

In my case, a 2h 13m recording was truncated to 54 minutes in the output MKV. The remaining ~79 minutes of segments were intact on disk but not included in the final file.

The workaround that successfully produced a complete file was running ffmpeg manually with:
```
ffmpeg -fflags +discardcorrupt -err_detect ignore_err \
  -f concat -safe 0 -i .dvr_18_hls/concat.txt \
  -c:v copy -c:a aac -b:a 192k \
  -map 0:v:0 -map 0:a:0 output_full.mkv
```

\*Note: I haven't tested whether this also occurs if the recording ends naturally at its scheduled time. But in my case, I stopped the recording early and ran into this issue. 

1. Schedule a DVR recording of a live IPTV channel.
2. (Maybe necessary): Stop the recording early via the Stop button before its scheduled end time.
3. Observe the recording status becomes stopped. ffmpeg attempts HLS-->MKV concat.
4. If any segment contains a corrupt AAC ADTS frame, the concat exits with rc=183. Look for this in container logs:

> WARNING apps.channels.tasks DVR recording N: direct HLS→MKV concat failed (rc=183); attempting MP4-intermediate fallback.
> [aac_adtstoasc @ 0x...] Error parsing ADTS frame header!
> [matroska @ 0x...] Error applying bitstream filters to an output packet for stream #1: Invalid data found when processing input

5. The output MKV exists on disk but is truncated at the point of the corrupt frame. The UI shows no error.
6. No intermediate .mp4 fallback file is produced; the fallback silently fails.

### Dispatcharr Version

v0.24.0, 90fa739

### Docker Version

29.3.0

```
name: dispatcharr

services:
  dispatcharr:
    container_name: dispatcharr
    restart: unless-stopped
    ports:
      - "9191:9191"
    environment:
      - TZ=America/Los_Angeles
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
    volumes:
      - /home/noah/docker/dispatcharr/data:/data
      - /srv/data/media/dispatcharr-recordings:/data/recordings
    devices:
      - /dev/dri:/dev/dri
```

### Client Information

Dispatcharr webui via Firefox on macOS. Also tried to play the recorded, corrupted clip via Safari to no avail.

nemesbak:
## Description

DVR recordings are silently truncated when any HLS segment contains a corrupt AAC ADTS frame. The direct HLS→MKV concat fails (rc=183), and the MP4-intermediate fallback also fails because `aac_adtstoasc` raises *"Error parsing ADTS frame header"*. No error is shown in the UI; the output MKV exists but is missing the segments after the corrupt frame.

**Fix:**

Three-tier approach:

1. **Primary path** — add `-fflags +discardcorrupt -err_detect ignore_err` so ffmpeg skips corrupt packets and continues to the end of the recording rather than aborting at the first bad frame.

2. **MP4-intermediate fallback** — same flags applied, so `aac_adtstoasc` no longer aborts on the corrupt frame.

3. **Audio re-encode fallback** — if both copy paths still fail, re-encode audio with `-c:a aac -b:a 192k` (matching the user's manual workaround). This produces a complete file at the cost of a lossy audio re-encode for the affected recording only.

Each step is logged at warning/error level so the fallback chain is visible in container logs.

## Related Issue

Closes #1225

## How was it tested?

- [ ] Recording with no corrupt frames → primary path succeeds, behaviour unchanged ✓

- [ ] Recording with corrupt AAC frame → primary path succeeds with discardcorrupt ✓

- [ ] Very corrupt stream → MP4 fallback succeeds ✓

- [ ] Worst case → audio re-encode fallback produces complete MKV ✓

## Checklist

### DIFF
--- a/apps/channels/tasks.py
+++ b/apps/channels/tasks.py
@@ -2503,9 +2503,13 @@ def run_recording(recording_id, channel_id, start_time_str, end_time_str):
                     _escaped = seg.replace("'", "'\\''")
                     _cl.write(f"file '{_escaped}'\n")
 
+            # Common flags that tolerate corrupt frames in the input segments.
+            _corrupt_flags = ["-fflags", "+discardcorrupt", "-err_detect", "ignore_err"]
+
             concat_result = subprocess.run(
                 [
                     "ffmpeg", "-y",
+                    *_corrupt_flags,
                     "-f", "concat", "-safe", "0",
                     "-i", concat_list_path,
                     "-c", "copy",
@@ -2522,8 +2526,7 @@ def run_recording(recording_id, channel_id, start_time_str, end_time_str):
             # MP4-intermediate fallback: some MPEG-TS streams (parameter set
             # changes mid-stream, weird PMT updates, audio discontinuities)
             # fail to mux directly into Matroska but go through an MP4
-            # container cleanly, which can then be remuxed losslessly to
-            # MKV.  Try this before declaring the recording lost.
+            # container cleanly, which can then be remuxed losslessly to MKV.
             if not _ok:
                 logger.warning(
                     f"DVR recording {recording_id}: direct HLS\u2192MKV concat failed "
@@ -2546,6 +2549,7 @@ def run_recording(recording_id, channel_id, start_time_str, end_time_str):
                 _mp4_concat = subprocess.run(
                     [
                         "ffmpeg", "-y",
+                        *_corrupt_flags,
                         "-f", "concat", "-safe", "0",
                         "-i", concat_list_path,
                         "-c", "copy",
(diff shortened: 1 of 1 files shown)


SergeantPanda (changes requested):
I like the added flags for tier 1 and 2 with this PR but let's not introduce transcoding here. I suspect 1 and 2 will help quite a bit when there are minor issues.

SergeantPanda:
Closing this, went another direction with robustness.

PR #1294, opened 2026-05-28
fix(channels): resolve source group relation when Channel Group Override is set

### LINKED ISSUE #1263: [Bug]: Compact numbering breaks when Channel Group Override is set on the source group
### Describe the bug

On a ChannelGroupM3UAccount with both Compact Numbering and Channel Group Override enabled, hiding an auto-synced channel does not release its channel_number slot, and unhiding does not assign a new number. Slots stay consumed by hidden channels indefinitely, producing spurious RANGE_EXHAUSTED failures on subsequent syncs.

Root cause: get_group_relation_for_channel and _repack_inner in apps/channels/compact_numbering.py look up the source ChannelGroupM3UAccount by channel.channel_group_id. When override is active, that field holds the override target's id, not the source group's id, so the lookup returns nothing and every compact path (post_save hide/unhide, bulk-edit hide transition, end-of-sync repack_group) short-circuits.

Expected: hide should set channel_number to NULL and unhide should reassign from the configured range regardless of whether Channel Group Override is set. Slot accounting should resolve the source group via the auto_created_by relationship, not via the channel's current channel_group_id.

1. On a ChannelGroupM3UAccount, enable Auto Channel Sync, set a Start/End range smaller than the source group's stream count, enable Compact Numbering, and set Override Channel Group to a different channel group.
2. Trigger M3U Refresh. Channels are created up to the range cap; extras fail with RANGE_EXHAUSTED.
3. Bulk-edit a subset of the created channels to hidden_from_output=true.
4. Observe that the hidden channels still have channel_number populated (not NULL).
5. Trigger M3U Refresh again. The previously RANGE_EXHAUSTED streams remain uncreated because the used_numbers seed still includes the hidden channels' numbers and repack_group filters by the now-mismatched channel_group_id.

Setting Override Channel Group back to "Source group (default)" and repeating the cycle confirms the release and repack paths work correctly without the override.

### Dispatcharr Version

v0.25.1

### Docker Version

v29.4.0

Bug is agnostic to deployment

### Client Information

Dispatcharr Web UI

nemesbak:
## Description

When **Compact Numbering** and **Channel Group Override** are both enabled on a `ChannelGroupM3UAccount`, hiding a channel does not release its slot and unhiding does not reassign one. Subsequent syncs keep hitting `RANGE_EXHAUSTED` even for ranges that should have free slots.

**Root cause:**

All compact-numbering paths (`get_group_relation_for_channel`, `_repack_inner`, and the bulk-unhide loop in `assign_numbers_for_channels`) looked up the `ChannelGroupM3UAccount` using `channel.channel_group_id`.

When a Channel Group Override is active, Auto Channel Sync stores created channels under the **override target** group's id, not the **source** group's id that is recorded on the `ChannelGroupM3UAccount`. The lookup always missed, every compact path short-circuited, and slot accounting was completely broken.

**Fix:**

Three targeted changes in `compact_numbering.py`:

1. **`get_group_relation_for_channel`**: when the direct `(m3u_account_id, channel_group_id)` lookup fails, fall back to scanning the account's relations and matching on `custom_properties["group_override"]`.

2. **`_repack_inner`**: build `group_ids` as both the source group id and the override target id (from `custom_properties["group_override"]`), then filter `channel_group_id__in=group_ids` so the repack sees all channels regardless of override state.

3. **`assign_numbers_for_channels` bulk path**: after the batched direct lookup, call `get_group_relation_for_channel()` for any `(account_id, group_id)` pairs that remain unresolved to handle the override case without adding an N+1 query for the common case.

## Related Issue

Closes #1263

## How was it tested?

- [ ] Compact numbering + no override → hide/unhide releases and assigns slots ✓ (regression guard)

- [ ] Compact numbering + override active → hiding a channel sets `channel_number = NULL` ✓

- [ ] Compact numbering + override active → unhiding reassigns a number from the range ✓

- [ ] Sync after hide/unhide → `RANGE_EXHAUSTED` no longer reported for freed slots ✓

## Checklist

### DIFF
--- a/apps/channels/compact_numbering.py
+++ b/apps/channels/compact_numbering.py
@@ -44,20 +44,37 @@ def is_compact_group(group_relation):
 def get_group_relation_for_channel(channel):
     """Resolve the ChannelGroupM3UAccount that owns this auto-created
     channel. Returns None for manual channels, or when the relation has
-    been deleted."""
-    if (
-        not channel.auto_created
-        or not channel.auto_created_by_id
-        or not channel.channel_group_id
-    ):
+    been deleted.
+
+    When a Channel Group Override is active, channel.channel_group_id
+    holds the *target* (override) group's id, not the source group's id
+    stored on the ChannelGroupM3UAccount.  In that case the direct lookup
+    fails, so we fall back to scanning relations for the same M3U account
+    and matching on the group_override custom-property.
+    """
+    if not channel.auto_created or not channel.auto_created_by_id:
         return None
-    try:
-        return ChannelGroupM3UAccount.objects.get(
-            m3u_account_id=channel.auto_created_by_id,
-            channel_group_id=channel.channel_group_id,
-        )
-    except ChannelGroupM3UAccount.DoesNotExist:
-        return None
+
+    if channel.channel_group_id:
+        # Fast path: no override active — source group == current group.
+        try:
+            return ChannelGroupM3UAccount.objects.get(
+                m3u_account_id=channel.auto_created_by_id,
+                channel_group_id=channel.channel_group_id,
+            )
+        except ChannelGroupM3UAccount.DoesNotExist:
+            pass
+
+        # Override path: channel.channel_group_id is the override target.
+        # Find the relation whose custom_properties["group_override"] matches.
+        for rel in ChannelGroupM3UAccount.objects.filter(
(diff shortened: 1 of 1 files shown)


SergeantPanda:
Thank you for this submission. The user that coded the current auto-sync has released a fix for it in favor of this PR.

PR #1295, opened 2026-05-28
fix(proxy): match IPTV player HTTP headers on provider stream connections

nemesbak:
## Description

When Dispatcharr proxies a stream to the client, it connects to the provider's stream URL using Python's `requests` library. The library automatically adds `Accept-Encoding: gzip, deflate, br, zstd` and `Connection: keep-alive` headers that a real IPTV player (libmpv, etc.) never sends for video stream connections. These extra headers, combined with the missing `Accept-Language` header that real players do send, form a fingerprint that Cloudflare and provider-side detection can use to identify the client as an HTTP library rather than a native player.

**Fix:**

- `HTTPStreamReader._read_stream()`: set `Accept`, `Accept-Language`, and suppress `Accept-Encoding` / `Connection` via `None` values (requests honours `None` as "remove this header").

- `InputManager._create_session()`: same treatment on the session-level headers.

**Headers comparison:**

| Header | MaxPlayer (captured live) | Python requests (before) | After fix |

|---|---|---|---|

| `User-Agent` | `MaxPlayer` / `MaxPlayer (win32 - x64) (libmpv)` | configured value | configured value ✓ |

| `Accept` | `*/*` | `*/*` | `*/*` ✓ |

| `Accept-Language` | `en_US` | *(absent)* | `en_US` ✓ |

| `Accept-Encoding` | *(absent)* | `gzip, deflate, br, zstd` | *(removed)* ✓ |

| `Connection` | *(absent)* | `keep-alive` | *(removed)* ✓ |

Headers were captured from live MaxPlayer iOS and Windows sessions via Caddy access logs.

**Note on TLS fingerprinting:**

The stream connections go to `103.163.132.7:80` (plain HTTP, not HTTPS), so JA3/TLS fingerprinting does not apply to stream proxy traffic. This fix addresses the HTTP-layer header fingerprint only.

## Checklist

### DIFF
--- a/apps/proxy/live_proxy/input/http_streamer.py
+++ b/apps/proxy/live_proxy/input/http_streamer.py
@@ -48,8 +48,17 @@ class HTTPStreamReader:
     def _read_stream(self):
         """Thread worker that reads HTTP stream and writes to pipe"""
         try:
-            # Build headers
-            headers = {}
+            # Build headers that closely match what a real IPTV player sends.
+            # Explicitly clear Accept-Encoding and Connection so requests does
+            # not add them — video streams are never compressed and the extra
+            # headers can reveal the client is an HTTP library rather than a
+            # native player.
+            headers = {
+                'Accept': '*/*',
+                'Accept-Language': 'en_US',
+                'Accept-Encoding': None,   # suppress requests default
+                'Connection': None,        # suppress requests default
+            }
             if self.user_agent:
                 headers['User-Agent'] = self.user_agent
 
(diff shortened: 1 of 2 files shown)


SergeantPanda:
There *may* be an argument here for changing some of these headers but this is not correct. There are glaring issues with what was decided and would make 'fingerprinting' Dispatcharr (which I don't even think has ever come up as an issue) even easier (for example, en_us is invalid, underscores are NOT used it should be a hyphen.
keep-alive is important
I could see potentially removing accept-encoding as it really wouldn't serve much of a purpose that I know of for streaming video. (would need to look into it more)

PR #1308, opened 2026-05-30
feat: blur public IP display, add IP lookup toggle with env var override, make lookup non-blocking

### LINKED ISSUE #1302: [Feature]: Disable Public IP

Not everyone has a public IP address.
Fills up the log file with failed connection attempts if I block it in my firewall.

The ability to set an environment variable in docker compose to turn this off.
Or the ability to set the public ip address yourself if you know it so these constant checks are not necessary.

### Additional context

Nothing Extra

sethwv:
## Description

The IP lookup in the `environment` endpoint previously made up to three sequential HTTP calls (ipify, ipapi.co, ip-api.com fallback) with 5s timeouts each, blocking the frontend settings load for up to 10+ seconds on a cold start. The raw public IP was also always fully visible in the sidebar.

Changes:
- The public IP field in the sidebar is now blurred by default and reveals on hover (only the text blurs, border and background are unaffected)
- A new toggle in Settings > System Settings lets users disable IP and geolocation fetching entirely
- A new `DISPATCHARR_ENABLE_IP_LOOKUP` environment variable provides a container-level override; when set to `false`, the setting is hidden from the UI entirely and cannot be changed
- When IP lookup is disabled, all three external calls are skipped and the sidebar field is hidden
- The lookup now runs in a background thread and results are cached for 1 hour. The endpoint returns immediately on every request from cache on warm loads, or with `ip_lookup_pending: true` on first cold request. The frontend retries once after 7 seconds so the IP appears shortly after load without blocking anything

## Related Issue

Closes #1302

## How was it tested?

- Confirmed settings page loads immediately on fresh start instead of waiting for IP lookups
- Confirmed public IP and flag appear in the sidebar a few seconds after load on a cold cache
- Confirmed subsequent page loads return the IP immediately from cache
- Confirmed public IP text blurs on idle and reveals on hover with border and background unaffected
- Confirmed toggling off IP lookup in System Settings hides the sidebar field and skips all external requests
- Confirmed `DISPATCHARR_ENABLE_IP_LOOKUP=false` hides the System Settings toggle and prevents all IP fetching

## Checklist


SergeantPanda (approved):
Thank you for the submission Seth, I made a few changes that I think will improve it a little bit more.

I switched to unblur on click instead of hover (to support mobile devices)
Added validation from ipify to ensure it's an IP before passing it on.
Switched from pulling after 7 seconds to the backend sending a websocket update so the frontend gets the IP ASAP.
Added a skeleton while waiting on the backend so the placeholder is always shown
Cleaned up some unused imports while I was in there (unrelated to you)

Overall this will be a solid improvement I think. Thank you!

sethwv:
@SergeantPanda awesome, thanks!

PR #1313, opened 2026-06-02
fix(channels): repair Channel Group Override interactions (compact numbering + clear-override)

### LINKED ISSUE #1263: [Bug]: Compact numbering breaks when Channel Group Override is set on the source group

### LINKED ISSUE #1276: [Bug]: For auto synced channels, the clear override buttons disappear from the edit channel screen when the data in the overridden field is identical to the default data in the auto synced field.

CodeBormen:
## Description

Two bug fixes in the auto-sync overhaul (FR #1196) area, both surfacing when Channel Group Override interacts with other channel features.

**#1263 - Compact numbering breaks when a Channel Group Override is set.** With an override on the source ChannelGroupM3UAccount, auto-sync stores created channels under the override target group's id, not the source group's id recorded on the relation. The compact paths that resolve the relation from `channel.channel_group_id` all missed, so: hiding a channel did not release its slot, unhiding did not assign one, and repack saw zero channels - producing spurious RANGE_EXHAUSTED on later syncs. Fixes in `apps/channels/compact_numbering.py`:

- `get_group_relation_for_channel`: on a direct-lookup miss, fall back to matching `custom_properties.group_override`. The fallback runs only on a miss, so the common no-override path keeps its single query.
- `_repack_inner`: match channels under both the source and override-target group ids; guards the `int()` coercion against a non-numeric stored value.
- `assign_compact_numbers_for_channels`: resolve override-target pairs with one extra query over the unresolved accounts (not one per pair), keeping the common path's single narrow query.

**#1276 - The clear-override control disappears when an override's value equals the provider value.** For auto-synced channels the reset affordance was computed by comparing the form value to the provider value. When an override coincides with the provider value (e.g. an override set while the provider differed, then a later sync brings the raw provider value to match it), the field read as not-overridden and the control vanished even though the override row still existed and the pencil still showed - leaving the override unclearable. `isFormFieldOverridden` is now existence-aware: a field counts as overridden when a persisted override row holds a value for it.

Reimplemented independently using the root-cause analysis in #1294 by @nemesbak as reference, with type-safety hardening, an N+1-safe bulk path, and test coverage.

## Related Issue

Closes #1263
Closes #1276

## How was it tested

- Reproduced both bugs on unpatched `dev` first: 3 failing backend assertions for #1263 (hide does not release, unhide does not assign, repack finds 0 channels) and 1 failing Vitest assertion for #1276. Confirmed each fix turns them green.
- Backend unit tests added: 5 Django tests for #1263 (hide/unhide/repack under override, a no-override fast-path regression guard, and a query-count scaling guard proving the override-aware repack stays query-neutral) and 5 Vitest cases for #1276 (existence-aware detection across manual channel, value divergence, value match without override, persisted-override-equals-provider, and explicitly-cleared field).
- Full backend suites (`apps.m3u`, `apps.channels`): the only 2 failures (migration backfill called with schema_editor=None; the summary endpoint returning JsonResponse while the test asserts .data) reproduce identically on a freshly built pristine `origin/dev`, so they are pre-existing and unrelated to this change.
- Frontend: full Vitest suite 4222/4222; ESLint and Prettier pass cleanly.
- Playwright E2E: spec 10 (channel overrides + compact numbering) and spec 03 (m3u) pass. Full Chromium suite run; every failure was reproduced identically on a pristine `origin/dev` build or passes in isolation (login throttle, DVR timing, EPG-XML and M3U-cache races), all in subsystems this change does not touch.
- Manual end-to-end against a live container. #1263: drove sync_auto_channels with a group override + compact range, confirmed hide releases the slot (channel_number to None), unhide reassigns from the range, and the slot accounting is identical to the no-override path. #1276: created the persisted-override-equals-provider state and verified in the edit form that Clear All Overrides and the per-field reset stay available and clear correctly.

## Checklist

### DIFF
--- a/apps/channels/compact_numbering.py
+++ b/apps/channels/compact_numbering.py
@@ -44,20 +44,38 @@ def is_compact_group(group_relation):
 def get_group_relation_for_channel(channel):
     """Resolve the ChannelGroupM3UAccount that owns this auto-created
     channel. Returns None for manual channels, or when the relation has
-    been deleted."""
-    if (
-        not channel.auto_created
-        or not channel.auto_created_by_id
-        or not channel.channel_group_id
-    ):
+    been deleted.
+
+    With a Channel Group Override active, sync stores the channel under
+    the override target group's id, not the source group's id recorded on
+    the relation. The direct lookup then misses, so fall back to scanning
+    the account's relations for one whose group_override points at the
+    channel's current group. The fallback runs only on a direct miss, so
+    the common no-override path keeps its single SELECT.
+    """
+    if not channel.auto_created or not channel.auto_created_by_id:
         return None
+    if not channel.channel_group_id:
+        return None
+
     try:
         return ChannelGroupM3UAccount.objects.get(
             m3u_account_id=channel.auto_created_by_id,
             channel_group_id=channel.channel_group_id,
         )
     except ChannelGroupM3UAccount.DoesNotExist:
-        return None
+        pass
+
+    # group_override may be stored as int or str; compare as strings so
+    # the match is type-agnostic.
+    target = str(channel.channel_group_id)
+    for rel in ChannelGroupM3UAccount.objects.filter(
+        m3u_account_id=channel.auto_created_by_id
+    ):
+        cp = _custom_properties_as_dict(rel.custom_properties)
+        if str(cp.get("group_override", "")) == target:
+            return rel
+    return None
 
 
 def build_reserved_set(exclude_channel_ids=None, range_start=None, range_end=None):
@@ -150,6 +168,29 @@ def assign_compact_numbers_for_channels(channel_ids):
     )
     for rel in relations_qs:
         relations_by_pair[(rel.m3u_account_id, rel.channel_group_id)] = rel
+
+    # Override fallback: pairs the direct lookup missed carry an override-
+    # target channel_group_id. Resolve them with one extra query over the
+    # unresolved accounts (not one per pair), so the common path keeps its
+    # single narrow query.
+    unresolved = [k for k in pair_keys if k not in relations_by_pair]
+    if unresolved:
+        override_relations = {}
+        for rel in ChannelGroupM3UAccount.objects.filter(
+            m3u_account_id__in={k[0] for k in unresolved}
+        ):
+            cp = _custom_properties_as_dict(rel.custom_properties)
+            target = cp.get("group_override")
+            if not target:
+                continue
+            try:
+                override_relations[(rel.m3u_account_id, int(target))] = rel
+            except (TypeError, ValueError):
+                continue
+        for key in unresolved:
+            rel = override_relations.get(key)
+            if rel is not None:
+                relations_by_pair[key] = rel
 
     by_relation = {}
     for key, group_channels in by_pair.items():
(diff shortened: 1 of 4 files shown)


SergeantPanda:
This looks good. Thank you!

PR #1315, opened 2026-06-02
fix(vod_proxy): fall back to stream_id when content UUID is dead (#961)

R3XCHRIS:
## Description

Read-side graceful degradation for the VOD UUID-orphaning failure reported in #961. When `_get_content_and_relation` (in `apps/proxy/vod_proxy/views.py`) hits a dead `Movie.uuid` / `Episode.uuid`, this patch now falls back to resolving the content via the stable `stream_id` that's already in the same request — so any `.strm` URL of the form `/proxy/vod/movie/<UUID>?stream_id=<S>` keeps working past the next orphaning event, without the external player needing to re-read the file from disk.

Implementation:

1. **Soft UUID lookup** — `Movie.objects.filter(uuid=content_id).first()` instead of `get_object_or_404`.
2. **Stream_id fallback when the UUID misses and `preferred_stream_id` is set.** Strictest-match first: when `preferred_m3u_account_id` is also present, an account-scoped relation query runs first; only on miss does it fall through to the unrestricted ordered query (`-m3u_account__priority, id` — mirrors the existing relation-selection ordering elsewhere in the function).
3. **`[STREAMID-FALLBACK]` logged at WARNING** so the underlying import-side churn (root cause in #961 / #973) stays visible while it remains unfixed.
4. **`Http404` raised only when BOTH identifiers fail** — message names both for diagnosability.

Identical pattern applied to the Movie and Episode branches. Series-UUID lookups are intentionally left UUID-only — players cache episode/movie URLs, not series URLs; the pattern can be added as a follow-up if needed.

### Relationship to #973

Closed [#973](https://github.com/Dispatcharr/Dispatcharr/pull/973) (also targeting #961, closed for inactivity 2026-04-13) was the **write-side root fix** at the import path. This PR is the complementary **read-side fix** at the proxy lookup:

| | #973 (closed) | This PR |
|---|---|---|
| Layer | `apps/vod/tasks.py` import batches | `apps/proxy/vod_proxy/views.py` lookup |
| Strategy | Prevent the data corruption | Tolerate dead UUIDs at request time |
| Covers UUID-bearing proxy URLs | Yes (preventive) | Yes (retroactive — also helps already-broken libraries) |
| Covers XC-compat URLs (`/movie/<u>/<p>/<id>.mkv`) | Yes | **No** — those don't carry a `stream_id` |
| Migration / surface area | Larger; touches import pipeline | None; 1 file, 1 function, +85 lines |

If both land, #961 is fully addressed. If only this PR lands, the underlying churn continues but UUID-bearing URLs survive it. Happy to revive #973 in a follow-up if the maintainer wants both — flagged for steer.

## Related Issue

Related to #961 (does **not** close it — fix is partial; XC-compat URL form is not covered, only UUID+stream_id URLs). See also closed #973.

## How was it tested?

**Unit tests — 7 new mock-based regression tests** in `apps/proxy/vod_proxy/tests/test_streamid_fallback.py`. Ran them against `ghcr.io/dispatcharr/dispatcharr:dev` in a transient one-shot container with the patched `views.py` mounted in:

```
Ran 7 tests in 0.019s
OK
```

Tests cover:

- `test_uuid_hit_no_fallback_attempted` (×2: movie + episode) — happy path unchanged; the `M3U*Relation` table is **never queried** when the UUID resolves (asserted via `RelMock.objects.filter.assert_not_called()`).
- `test_uuid_miss_resolves_via_stream_id` (×2: movie + episode) — UUID miss + `stream_id` resolves and emits `[STREAMID-FALLBACK]` at WARNING.
- `test_uuid_miss_prefers_requested_account_first` — strictest-match guarantee: the account-scoped relation query runs first; the unrestricted ordered fallback is never consulted when the strict match succeeds.
- `test_uuid_miss_with_no_stream_id_raises_404` — fallback is never attempted when there's nothing to fall back to.
- `test_uuid_miss_with_no_matching_relation_raises_404` — both lookups fail cleanly.

**End-to-end on a live install** (pre-PR, against my own running container):

Failing request before the patch:

```
[VOD-REQUEST] /proxy/vod/movie/7a8fcee7-…?stream_id=1530724
ERROR Error getting content object: No Movie matches the given query.
GET 500 /proxy/vod/movie/7a8fcee7-…?stream_id=1530724
```

After the patch (same request):

```
WARNING [STREAMID-FALLBACK] Movie UUID 7a8fcee7-… not found; resolved via stream_id 1530724 -> movie uuid d89d65a9-… (provider: AM-Onyx)
INFO [CONTENT-FOUND] Movie: Over Your Dead Body (2026) (ID: 51309)
INFO [STREAM-SELECTED] Using specific stream: 1530724 from provider: AM-Onyx
# HTTP 206, bytes flowing
```

Verified end-to-end through real Emby playback on three previously-failing titles. Also verified that requests where the UUID is valid still take the existing happy path unchanged (no regression).

## Checklist

- [x] Backend: migrations are included if any models were changed *(no model changes — no migration required)*
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema *(no new endpoints — internal helper only)*
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) *(no frontend changes)*

### DIFF
--- a/apps/proxy/vod_proxy/tests/test_streamid_fallback.py
+++ b/apps/proxy/vod_proxy/tests/test_streamid_fallback.py
@@ -1,0 +1,225 @@
+"""
+Tests for `_get_content_and_relation`'s graceful stream_id fallback when the
+VOD content UUID has been orphaned by an import-time refresh.
+
+Context (see #961 / closed #973): `process_movie_batch` and `process_series_batch`
+can create duplicate `vod_movie` / `vod_episode` records during a refresh and
+repoint existing `M3U*Relation` rows at the new records. The old UUIDs that
+external players (Emby / Jellyfin / ChannelsDVR) cached in `.strm` URLs are
+left orphaned, and the proxy then 404s — even though the same request carries
+a stable `stream_id` that uniquely identifies a live relation.
+
+These tests cover the read-side fallback that resolves content via that
+stream_id when the UUID lookup misses, leaving the existing UUID-first path
+unchanged for the happy case. Both branches (movie + episode) exercise:
+
+  * UUID hit (no fallback fires) — happy path unchanged
+  * UUID miss + stream_id present → resolved via stream_id, [STREAMID-FALLBACK]
+    logged at WARNING
+  * UUID miss + stream_id present + preferred_m3u_account_id → strictest-first
+    account match preferred
+  * UUID miss + stream_id present but no relation matches → Http404 with both
+    identifiers in the message
+  * UUID miss + no stream_id → Http404 (no fallback attempt)
+"""
+
+import logging
+from unittest.mock import MagicMock, patch
+from django.test import SimpleTestCase
+from django.http import Http404
+
+
+# ---------- Movie branch --------------------------------------------------
+
+class TestStreamIdFallbackMovie(SimpleTestCase):
+    """Movie UUID dead -> fall back to M3UMovieRelation.stream_id."""
+
+    def _call(self, **kwargs):
+        # Imported inside each test so the module-level Movie / Episode /
+        # M3U*Relation references can be patched per test without leaking.
+        from apps.proxy.vod_proxy.views import _get_content_and_relation
+        return _get_content_and_relation(
+            kwargs.pop('content_type', 'movie'),
+            kwargs.pop('content_id', 'dead-uuid'),
+            preferred_m3u_account_id=kwargs.pop('preferred_m3u_account_id', None),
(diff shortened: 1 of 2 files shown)


SergeantPanda (approved):
This looks good thank you.

Regarding the PR #961, if you'd like to pick it back up and finish it I would be OK with that. Just please read through what has already been discussed and implement it in a way that will not cause unneccesary processing or regressions for the normal scenario (imdb,tmdb id's existing)

PR #1323, opened 2026-06-05
fix(channels): make compact numbering repack idempotent

### LINKED ISSUE #1321: [Bug]: Auto-synced channel numbers reshuffle on every sync (compact numbering is not idempotent)
### Describe the bug

On a ChannelGroupM3UAccount with Compact Numbering and Auto Channel Sync enabled, the numbers assigned to auto-created channels change on subsequent syncs even when nothing changed at the provider (no streams added or removed). Channels swap numbers within their configured [start, end] range from one refresh to the next.

Expected: channel numbering should be stable/idempotent across syncs. A channel's number should only change when the stream set changes (adds/deletes) or the user changes the sort order or range. With no provider changes, repeated syncs should produce identical numbers.

Root cause (for reference): in apps/channels/compact_numbering.py, _repack_inner selects the group's channels with no ORDER BY, and for the default "provider" sort order it does not sort them before packing. The pack assigns sequential numbers in the database's physical row order, which is not stable - it shifts after the renumber's own UPDATEs to the indexed channel_number column and after (auto)vacuum. So each sync can repack the same channels into different numbers.

1. Create or select an M3U account with a group containing several streams.
2. In M3U Group settings, enable Auto Channel Sync and Compact Numbering for the group, leave the sort order at the default (provider), and set a Start (e.g. 100).
3. Refresh the M3U. Note the channel numbers assigned to the group's channels (e.g. 100, 101, 102, ...).
4. Trigger one or more additional M3U refreshes with no changes at the provider (no streams added or removed).
5. Observe that the channel-number-to-channel mapping within the group has changed - channels have swapped numbers within the range - despite no streams being added or removed.

Deterministic reproduction: after the first refresh, change the channel table's physical row order (e.g. CLUSTER/VACUUM FULL on the channel table, or any operation that moves rows), then trigger another sync or hit the per-group re-pack endpoint - the same channels get reassigned different numbers.

### Dispatcharr Version

v0.25.1

### Docker Version

v29.5.2

Issue is not deployment-based

### Client Information

Issue is server-side

CodeBormen:
## Description

Auto-synced channel numbers reshuffled within their configured range on every sync, even when the provider returned no changes. Compact numbering was not idempotent.

`_repack_inner` (the compact pack) read its auto-created channels with no `ORDER BY`, so the pack followed PostgreSQL's physical row order. That order is unspecified and drifts after the `UPDATE`s each repack issues and after autovacuum, so the default "provider" sort packed the same channels into different numbers on each run. The result was daily channel-number churn that broke client favorites and channel memory.

The fix makes the pack deterministic:

- The `_repack_inner` channel query now uses `.order_by("id")`. id order is creation order, which tracks the provider's stream order (what the default "provider" sort is meant to preserve).
- The explicit `name` / `tvg_id` / `updated_at` sorts now carry `c.id` as a secondary key, so channels with equal sort values (e.g. a blank `tvg_id`) keep a stable relative order instead of churning.

No behavior change for users beyond stability: numbers land in the same range and order they did before, they just stop moving between syncs.

## Related Issue

Closes #1321

## How was it tested?

Reproduced on unpatched code first, then verified the fix.

A new regression test (`CompactNumberingIdempotencyTests` in `apps/m3u/tests/test_sync_correctness.py`) packs a set of auto-created channels, forces the underlying rows into a physical order opposite to their id order, then repacks again and asserts the channel-to-number mapping is identical across both passes. It reproduces the churn deterministically: it fails on the unpatched code (numbers flip once physical order diverges from id order) and passes with `.order_by("id")` in place.

The existing compact-numbering and Channel Group Override coverage still passes, including pack, repack, hide and unhide slot release, and override-pinned channels left untouched. The same behaviors were also verified end to end through the channel override UI and API.
## Checklist

### DIFF
--- a/apps/channels/compact_numbering.py
+++ b/apps/channels/compact_numbering.py
@@ -324,12 +324,17 @@ def _repack_inner(group_relation):
     # into the SAME target group, their channels are indistinguishable
     # here (channels carry no source-group back-reference), so each repack
     # renumbers the shared target's channels into its own range.
+    # order_by("id") makes the pack deterministic. Without it the query
+    # returns rows in unspecified physical order, which shifts after the
+    # renumber's own UPDATEs and autovacuum, so the default "provider" sort
+    # below would repack channels into different numbers on every sync.
+    # id order is creation order, which tracks the provider stream order.
     channels = list(
         Channel.objects.filter(
             auto_created=True,
             auto_created_by_id=account_id,
             channel_group_id__in=group_ids,
-        ).select_related("override")
+        ).select_related("override").order_by("id")
     )
 
     visible = []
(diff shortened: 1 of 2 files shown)


SergeantPanda (commented):
This looks pretty good. Thank you for a small diff!

Curious what is the impact for users that use channel_sort_reverse?

CodeBormen:
I just tested channel_sort_reverse and it is an exact mirror, which is what I would think is expected from a UX perspective.

And you mean you don't love my huge diffs 😢😂

SergeantPanda:
> I just tested channel_sort_reverse and it is an exact mirror, which is what I would think is expected from a UX perspective.
> 
> And you mean you don't love my huge diffs 😢😂

I've had a LOT of large diffs lately, brain has been quite fried trying to read code I didn't write 😂.

Merged, thank you!

PR #1325, opened 2026-06-06
fix(channels): add polling interval to refresh stream stats in expanded channel rows

### LINKED ISSUE #1324: [Bug]: Stream stats not refreshing in expanded channel rows (regression from PR #1202)
### Describe the bug

In the Channels page, when a channel row is expanded to show its configured streams, the stat badges (resolution, FPS, codec, audio, output bitrate) do not update/appear to update and/or show these stats. For streams that have been played previously, stats show on expand sometimes but then never refresh. For streams that have never been played or are newly configured to a channel, no stats are displayed at all until the user either drags a stream to a different position or navigates away and returns. At times, this doesn't appear to work either, which was a previous "work around" I was using prior.

Expected behavior:
Stream stat badges should update periodically while a channel row is expanded, without requiring any user interaction.

Area/file:
Frontend — ChannelTableStreams.jsx

Labels:
bug, area: frontend, regression

1. Go to the Channels page
2. Expand a channel row that has an active stream
3. Wait — stat badges do not update (or even display sometimes, random).
4. Drag a stream to a different position — stats *sometimes* update (confirms the data is available but not refreshing automatically)

### Dispatcharr Version

Dev build - post PR #1202

### Docker Version

Docker version 29.4.1, build 055a478

services:
  dispatcharr:
    build:
      context: .
      dockerfile: docker/Dockerfile
    container_name: dispatcharr_dev
    restart: unless-stopped
    ports:
      - 9191:9191
      - 8001:8001
    volumes:
      - dispatcharr_data:/data
    environment:
      - POSTGRES_HOST=db
      - POSTGRES_PORT=5432
      - POSTGRES_DB=dispatcharr
      - POSTGRES_USER=dispatch
      - POSTGRES_PASSWORD=secret
      - DJANGO_SECRET_KEY=dev-secret-key-not-for-production
      - DISPATCHARR_LOG_LEVEL=info
    depends_on:
      db:
        condition: service_healthy

  db:
    image: postgres:17
    container_name: dispatcharr_dev_db
    restart: unless-stopped
    environment:
      - POSTGRES_DB=dispatcharr
      - POSTGRES_USER=dispatch
      - POSTGRES_PASSWORD=secret
    volumes:
      - postgres_data:/var/lib/postgresql/data
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U dispatch -d dispatcharr"]
      interval: 5s
      timeout: 5s
      retries: 5

volumes:
  dispatcharr_data:
  postgres_data:

### Client Information

Browser — Chrome / Firefox (web UI)

Shokkstokk:
## Description

Stream stat badges (resolution, FPS, codec, audio, output bitrate) were not updating in expanded channel rows on the Channels page. For streams that had been played previously, stats would display on expand but never refresh. For streams that had never been played or were newly added to a channel, no stats were shown at all.

`ChannelTableStreams` called `refreshStats()` once on mount and never again. The drag-reorder action previously caused an incidental full re-render that happened to refresh stats as a side effect. PR #1202 (commit `17fbf1d4`) stabilized the TanStack column definitions and introduced `useRef`-based advanced stats tracking as a render performance optimization but removed that incidental trigger in the process.

The fix adds a `setInterval` that polls `API.getChannelStreamStats` with `since=null` every 15 seconds while the row is expanded, patching the store via `patchChannelStreamStats` on each tick. The interval is torn down in the `useEffect` cleanup when the row collapses so there are no lingering API calls for rows not in view. The initial on-expand fetch is preserved and still uses the delta cursor for efficiency.

The 15 second poll interval is intentionally shorter than the backend's 30 second `_bitrate_db_save_interval`. The throttle on output bitrate writes is specific to that field only — codec, resolution, FPS, and audio can update at any time (e.g. on stream failover) and should be picked up pretty much immediately. The poll is a lightweight read query and only runs while a row is actively expanded.

All render performance improvements from PR #1202 are left intact.

## Related Issue

Closes #1324 

## How was it tested?

Tested on dev build with an active stream. Expanded a channel row and confirmed stat badges update within 15 seconds without any user interaction. Confirmed the interval tears down cleanly on row collapse via browser network tab (no further requests after collapse). Confirmed drag-reorder still functions correctly.

## Checklist

### DIFF
--- a/frontend/src/components/tables/ChannelTableStreams.jsx
+++ b/frontend/src/components/tables/ChannelTableStreams.jsx
@@ -557,9 +557,18 @@ const ChannelStreams = ({ channel }) => {
     [patchChannelStreamStats]
   );
 
-  // Refresh once when the row is expanded.
+  // Refresh once on expand, then poll every 15 seconds while the row remains open.
+  // Pass since=null on interval ticks so we always fetch the full current stats
+  // rather than relying on a delta cursor that could miss updates.
   useEffect(() => {
     refreshStats();
+    const interval = setInterval(() => {
+      API.getChannelStreamStats(channelRef.current?.id, null).then((updates) => {
+        if (!updates || updates.length === 0) return;
+        patchChannelStreamStats(channelRef.current?.id, updates);
+      });
+    }, 15000);
+    return () => clearInterval(interval);
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 


SergeantPanda:
The stats are currently pulled in when the web video player closes. I think simply expanding it to when a stream end instead would fix this. I assume what's happening is you are clicking the preview button and either not letting it play for 5 seconds and then closing the video player or you're previewing the stream without closing the windows in between at all (which should be improved)

Shokkstokk:
Closing as discussed off-thread with Issue #1324

PR #1328, opened 2026-06-09
Fix/1273 provider numbering

### LINKED ISSUE #1273: [Bug]: Auto Channel Sync Broken (channel number assignment)
### Describe the bug

When adding an m3u account, if you do not enable sync on groups as you add the account, then enabling sync later will override the current channel lineup and create new channels with new channel numbers, even if everything matches. 

1. I add my m3u account.
2. I refresh to update all the streams.
3. On the channels page, i select all streams and create channels using my provider channel number. This works perfectly and creates a list of channels with categories and numbers identical to the playlist i have provided in the m3u.
4. I then go to m3u settings, and group settings, and i enable auto channel sync for all groups which are ppv, because for these i want dispatcharr to update the channel name with the next event when it is provided by my playlist provider. in the auto sync options i choose to use the provider number.
5. Before the latest dispatcharr update, this would then result in the channel name being updated if there was a change in the m3u file, but the channel number etc would be maintained. 
6. what happens now, is for example if the m3u playlist for this group had 100 channels numbered 250 - 350, those channels would remain and not update, and and dispatcharr will create 100 new channels at the end of the playlist. 
7. I wondered whether there had been some kind of change to those channels in the m3u file which meant dispatcharr identified them as new channels, but if i manually delete all the channels then add them again from 'streams' on the channels page, they are added with the correct numbers that were provided in the m3u file.
8. If have noticed that if you start from fresh and enable the group channel sync for the ppv groups as you add the m3u account and before you refresh and load the data then it will assign the correct channel numbers and it will maintain those channel numbers when you refresh.

### Dispatcharr Version

0.25.1

### Docker Version

not sure

### Client Information

CodeBormen:
## Description

The auto-sync overhaul added a per-group `[start, end]` numbering range and refactored channel numbering into a shared picker, but each numbering mode's UI exposes only a subset of the persisted fields while the backend read all of them, and switching modes never resets the others. As a result, stale or auto-computed values silently changed numbering:

- In **Provider** mode the visible "Start #" writes `channel_numbering_fallback`, but the backend floored provider numbers at `auto_sync_channel_start` (auto-computed and not editable in this mode), so valid provider numbers were discarded and the channels packed from a high start.
- In **Next Available** mode the UI exposes no Start/End, yet the backend capped assignment at a stale `auto_sync_channel_end` left over from a prior mode.
- The range-enforcement delete (which removes auto-created channels outside `[start, end]`) ran in every mode, deleting provider-numbered and next-available channels based on values those modes do not expose.
- Provider mode lacked the Start>End guard that Fixed mode has, so an inverted fallback range silently failed every numberless stream.

This change makes the backend read only the fields each mode owns:

- **Provider** numbers are authoritative and used as-is; `channel_numbering_fallback` (the visible Start) and End bound only the fallback for streams with no provider number. `auto_sync_channel_start` is no longer applied in Provider mode.
- **Next Available** ignores End.
- Range enforcement runs in **Fixed** mode only, the one mode with a deliberate, user-set `[start, end]` range.
- Provider mode gains the Start>End guard.

Fixed mode behavior is unchanged.

## Related Issue

Closes #1273

## How was it tested?

Added backend unit tests that reproduce each interaction on the unpatched code and then assert the corrected behavior: a provider number below, within, or above the group's Start/End is used as-is; a numberless stream falls back from the visible Start rather than the hidden auto-computed start; Next Available is not capped by a stale End; provider and next-available channels are not deleted for falling outside a range; and a duplicate or already-taken provider number falls back to a free number instead of being dropped or overwriting an existing channel. Added a frontend unit test for the Provider Start>End guard. Confirmed the existing Fixed-mode range-enforcement and exhaustion tests still pass, since range enforcement is now scoped to Fixed mode.

## Checklist

### DIFF
--- a/apps/m3u/tasks.py
+++ b/apps/m3u/tasks.py
@@ -1740,6 +1740,14 @@ def _pick_target_number(
     return _next_available_number(used_numbers, fixed_cursor, end=end_number)
 
 
+def _range_exhausted_error(mode, start_number, end_number, fallback_start):
+    """User-facing range text for RANGE_EXHAUSTED failures."""
+    range_start = (
+        int(fallback_start) if mode == "provider" else int(start_number)
+    )
+    return f"Channel number range {range_start}-{int(end_number)} is full"
+
+
 def _custom_properties_as_dict(value):
     """
     Normalize a JSONField-backed custom_properties value into a dict.
@@ -2454,9 +2462,11 @@ def sync_auto_channels(account_id, scan_start_time=None):
                                     "stream_id": stream.id,
                                     "group": channel_group.name,
                                     "reason": "RANGE_EXHAUSTED",
-                                    "error": (
-                                        f"Channel number range "
-                                        f"{int(start_number)}-{int(end_number)} is full"
+                                    "error": _range_exhausted_error(
+                                        channel_numbering_mode,
+                                        start_number,
+                                        end_number,
+                                        channel_numbering_fallback,
                                     ),
                                 })
                             processed_stream_ids.add(stream.id)
--- a/apps/m3u/tasks.py
+++ b/apps/m3u/tasks.py
@@ -1716,42 +1716,27 @@ def _pick_target_number(
     fixed_cursor,
     fallback_start,
     end_number=None,
-    range_start=None,
 ):
     """
-    Return the channel number a given stream should claim under the group's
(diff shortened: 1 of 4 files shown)


SergeantPanda (approved):
Thank you for this! I made one minor change to just logging for provider numbering. Overall good.

PR #1333, opened 2026-06-10
Settings Menu Redesign & Sidebar Refresh

sethwv:
## Description

Replaces the accordion-based Settings page with a slide-over navigation pattern integrated into the existing sidebar.

- New `SlidingPanels` component and `usePanelNav` hook handle the animated two-panel transition (primary nav slides left, settings nav slides in from the right)
- New `settingsNav.js` config is the single source of truth for settings groups and sections
- Clicking Settings in the main nav opens the settings sub-panel without navigating; the content pane only updates when a specific section is selected
- Clicking Back closes the sub-panel without leaving the selected settings section
- Main nav groups now render as flat sections with uppercase headings instead of collapsible accordions
- Sidebar collapsed/expanded state persisted to `localStorage`

## How was it tested?

- Navigated between all settings sections, confirmed content pane updates on section click only
- Clicked Back from a selected section, confirmed content stayed on that section until navigating elsewhere
- Toggled sidebar collapse, refreshed the page, confirmed state was restored
- Verified admin and non-admin nav items both render correctly
- Checked collapsed icon-only mode for both the primary and settings panels

## Checklist


SergeantPanda (changes requested):
I like the way this transitions from the main sidebar to the settings side bar.

I think we'd need to put a width limit on the settings as it looks a bit wild :joy:

Also I'm not sure how to properly handle the sidebar reordering capabilites if we changed the sidebar grouping. Currently you could move DVR and put it in between connections and logs.

I realize this is a POC though. I think it could be a winner with some tweaks. 

Honestly my biggest concern is on a 1080P monitor, this design leaves no room leftover for more links. Although I suppose we could reduce the spacing a bit if we needed to.

sethwv:
@SergeantPanda 
> Also I'm not sure how to properly handle the sidebar reordering capabilites if we changed the sidebar grouping. Currently you could move DVR and put it in between connections and logs.

The headings are still treated as groups, so you can't actually re-order any menu item to be inside of another group, just between them.

> I think we'd need to put a width limit on the settings as it looks a bit wild

I've modified this to be more in line with the width of the users page's table, and added the border and background of the panes back in.

> Honestly my biggest concern is on a 1080P monitor, this design leaves no room leftover for more links. Although I suppose we could reduce the spacing a bit if we needed to.

I've bumped around the margins, padding, and changed the icons from sm -> xs. That leaves a bit more space

SergeantPanda:
For the reordering, maybe just the fact that it gets put ABOVE the separator instead of below it is why it looks weird:
image

I think I like the look of the settings better now.

Now I think the links are too small :joy: something in between those might be needed or just go back to the larger size.

If you are sick of playing with it, I can mess with it too (I know I'm picky)

sethwv:
@SergeantPanda Do you still want to entertain this one? I'm happy to scrap/kick back to draft if it's a bit extreme or not wanted anymore.

SergeantPanda:
Yeah I'm still wanting to entertain it. Sorry I've been trying to fix this epg memory leak in the little time I've had the last week. I think I made a breakthrough last night so hopefully I can get back to this stuff.

SergeantPanda:
I'll play with it soon. I am reviewing the HLS pr right now.

frontend/src/pages/Connect.jsx
```diff
@@ -208,3 +245,280 @@ function IntegrationRow({ integration, editConnection, deleteConnection }) {
     </Card>
   );
 }
+
+function ConnectLogsSection({ integrations }) {
```
SergeantPanda: I feel like this should probably be moved out of here and into it's own component, probably along side `SystemEvents.jsx` under components.

PR #1344, opened 2026-06-12
feat(proxy): native HLS output for live channels

### LINKED ISSUE #1003: [Feature]: Native HLS Output (m3u8 + segments) Support

Yes.

Currently Dispatcharr outputs streams in MPEG-TS format.
When serving multiple concurrent users, per-client TS streaming may increase server load and upstream connections.

There is no native HLS (m3u8 + segments) output option, which limits scalability and caching efficiency.

I would like Dispatcharr to support native HLS output (m3u8 + segments),
either by merging PR #210 or implementing a similar mechanism.

The ideal behavior would be:

- Accept HLS or other input streams
- Generate shared HLS output segments
- Avoid per-client MPEG-TS streaming
- Improve scalability for concurrent viewers

### Additional context

While researching this, I found PR #210 which appears to implement HLS output functionality.

This feature would be very useful for deployments where multiple users access the same channel concurrently.

Is there any plan to continue development or merge this PR?

jonzey231:
## Description

Adds HLS as a Default Output Format option in System > Settings > Stream Settings. It will show up in the drop down below MPEG-TS and fMP4.

This also makes it so apps requesting a channel with `?output_format=hls` (or the Xtream `.m3u8` extension) returns a real HLS media playlist so HLS-native clients like Safari, AVPlayer on iOS/tvOS, QuickTime on macOS, VLC, etc can play live channels without the client having to remux anything itself. I recently added a remux workaround in my personal project and while it works well, this output works a lot better.

This plugs into the output-format registry the proxy refactor introduced rather than adding a parallel path. `_parse_output_key`'s docstring already uses `'hls:p3'` as its example and `redis_keys.py` notes that "adding a new output format only requires a new manager," so this follows that lead: one new manager class, the existing `output:{fmt}` keyspace, and two URL routes.

## Related Issue

I believe this will close #1003. More than happy to take feedback and make any changes you think would make this work better.

## How it works

- **`HLSOutputManager`** (`apps/proxy/live_proxy/output/hls/manager.py`) registers beside `FMP4RemuxManager` in `_OUTPUT_FORMAT_MANAGERS`. It runs one daemon thread per (channel, format key), no subprocess: the source ring is already 188-byte aligned TS, so producing TS-segment HLS is splitting on keyframe boundaries, not remuxing.
- **`segmenter.py`** is dependency-free (stdlib only, unit-tested in isolation): it tracks PAT/PMT, finds the video PID, cuts a segment at the first video keyframe at or after the target duration (adaptation-field `random_access_indicator`, with an H.264/H.265 NAL-scan fallback for providers that do not set RAI), measures durations from PES PTS deltas, and prefixes every segment with the latest PAT+PMT so segments decode independently.
- Segments are stored one per Redis chunk through the existing format-parameterized output buffer; the chunk index doubles as the HLS media sequence number. A rolling playlist descriptor rides a new `output:{fmt}:playlist` key in the same namespace, so the existing stop/cleanup sweeps cover it.
- **Endpoints**: the existing `/proxy/ts/stream/<uuid>?output_format=hls` request runs the full init/auth/client-registration path, then 302s to a client-scoped playlist at `/proxy/hls/<channel_uuid>/<client_id>/index.m3u8`, with segments served as sibling `<seq>.ts` paths. I first had these under `/proxy/ts/hls/...`, but per your Discord comment about the endpoints not matching, they now live under the existing `/proxy/hls/` prefix; the legacy `hls_proxy` include keeps URL priority (its routes all start with a literal prefix, so the channel-scoped paths cannot shadow them), and segment names match its `<n>.ts` vocabulary. Happy to rename further if you want a different shape. Every playlist/segment request touches the client record, so polling players stay alive and stopped players are reaped by the existing ghost-client heartbeat, which feeds the existing zero-clients shutdown chain. No new teardown code.
- `generate_m3u` already forwards `output_format` into generated URLs, so `GET /output/m3u?output_format=hls` works once the alias exists; the only output-side change is advertising `m3u8` in the XC `allowed_output_formats`.

Scope kept deliberately small for v1: TS segments only (no fMP4/CMAF, no HEVC special-casing), no LL-HLS, no multivariant playlist, no transcoding changes, two `proxy_settings` knobs (`HLS_SEGMENT_DURATION`, `HLS_WINDOW_SIZE`) with safe defaults and no new frontend settings UI beyond the format dropdowns. Roughly 800 new lines in `output/hls/`, under 30 lines touched in existing files.

**Design note / your call:** the same `HLSOutputManager` interface can wrap an ffmpeg-based segmenter (`-f segment` / `-f hls` to a tmpfs harvested back into Redis) instead of the Python splitter; swapping is a one-file change behind the same registry entry, keyspace, URL surface, and teardown. I went with the Python splitter because it spawns no processes and stays entirely in the existing Redis ring, but if you would rather lean on ffmpeg I will build that instead.

## How was it tested?

- Segmenter unit tests (`test_segmenter.py`, 12 cases): PAT/PMT/PTS parsing, RAI and NAL-scan keyframe detection, target-duration cuts on keyframe boundaries, PAT+PMT prefix on every segment, discontinuity propagation, resync after garbage, PTS-wrap tolerance, and playlist rendering (target-duration ceiling, media-sequence monotonicity, discontinuity placement).
- Tested on Unraid with Dispatcharr v0.26.0. Requested a live H.264 + AC-3 channel with `?output_format=hls`, confirmed the 302, played the resulting playlist in Safari and validated it with `ffprobe` (`format_name=hls`).
- Also exercised the XC `.m3u8` path and `GET /output/m3u?output_format=hls`.
- I also tested this on AerioTV using Apple's AVPlayer and it worked incredibly well.

## Checklist

- [ ] I understand line by line every change in this PR and can explain it if asked
- [x] Backend: migrations are included if any models were changed (none changed)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (same `@api_view` treatment as the existing streaming endpoints in this file)


SergeantPanda:
I haven't forgotten about this one yet!

Sorry, focus for the next major release is going to be catchup and then I will look at this one.

SergeantPanda:
There are a couple of (minor) merge conflicts with this one, would you mind resolving them? I'm hoping to start looking in to this one soon. Might be for the next release but I'm not sure yet.

jonzey231:
Conflicts resolved. Rebased onto the latest dev; the collisions were with the stream_ts/stream_xc refactor and the removal of the legacy hls_proxy mount. The HLS branch in stream_ts now follows the resolved_output_format path, checks the ensure_output_format result, and cleans up the pre-registered client on failure, matching the fMP4 branch. The playlist and segment endpoints picked up the same close_old_connections handling the other proxy views got, and the /proxy/hls/ routes now stand alone in apps/proxy/urls.py.

Since you're about to review, I also folded in a set of correctness fixes that came out of extended testing against real clients (AVFoundation on Apple TV and iOS, ExoPlayer/Media3 on Android TV, and hls.js) after I submitted:

- EXT-X-TARGETDURATION is computed once per session and frozen. RFC 8216 6.2.1 says it must not change across reloads, and deriving it per render from the sliding window made it flap on GOP jitter, which can wedge AVPlayer mid-stream. A segmenter ceiling force-cuts any segment that would exceed the advertised value so the frozen number stays truthful.
- EXTINF on a discontinuity cut now reports the measured span of the closed segment instead of substituting the nominal target (RFC 8216 4.3.2.1).
- The playlist pins a deterministic live-edge join point with a byte-stable EXT-X-START tag once the window is deep enough; clients that set their own offset still override it.
- The segment chunk TTL in Redis is sized to the live window plus roll-off availability (RFC 8216 6.2.2), and the default window is now 10 segments, so a client that briefly stalls can still fetch the segment it is on instead of hitting a 404.
- The output owner lock and state key are now heartbeated. They were set once with a 1 hour TTL, so a stream longer than an hour would silently lose mutual exclusion and a second worker could start a duplicate segmenter. On takeover the window and frozen target are seeded from the existing descriptor so MEDIA-SEQUENCE never regresses.

All of it is in one commit on top of the original five so it is easy to see what changed since the initial submission. Segmenter unit tests cover the new invariants (13 passing). Happy to split anything out or adjust endpoints and naming however you prefer.

SergeantPanda (commented):
Thanks for this!

The overall shape looks good: plugging into the output-format registry, Redis-backed segments, client-scoped `/proxy/hls/...` playlist + segment routes, and reusing the existing client touch / ghost-reaper lifecycle. That part I’d like to keep.

Where I want to change direction is the cutter itself. Rather than maintaining a Python packet-copy segmenter (PAT/PMT parsing, RAI/NAL keyframe detection, PTS deltas, discontinuity flags, etc.), I’d prefer we remux/segment with ffmpeg the same way we do for fMP4.

For fMP4 we already pipe the TS ring through ffmpeg and push the result into Redis. HLS can follow that model: TS in, `-c copy` out as TS segments, then store each finished segment in Redis via the existing output buffer (`put_fragment`) and keep rendering the rolling playlist ourselves. Segment TTL can use the same configured Redis chunk TTL as the normal buffer, with a floor so anything still listed in the live window (plus a bit of roll-off) can’t expire out from under the playlist.

Reasons I’m leaning this way:
- Live IPTV sources are messy. ffmpeg already handles keyframes, PTS wraps, multi-packet PMT, and a lot of the discontinuity cases we’d otherwise keep discovering in the field.
- We avoid owning a growing demuxer/segmenter in Python.
- The manager, keyspace, URLs, and teardown can stay basically as you’ve designed them. It’s mostly swapping the segment production path.

I’m not asking to throw away the integration work. Keep `HLSOutputManager`, the playlist descriptor, the endpoints, and the frontend/XC wiring. Just drive segment creation with ffmpeg instead of `segmenter.py`.

If you want a concrete target: mirror the fMP4 manager’s stdin/stdout process lifecycle, write segments into Redis, and keep playlist generation in our code so we still control `EXT-X-TARGETDURATION`, media sequence, and discontinuity tags (including when we skip ahead in the ring or switch providers).

Overall I like how clean you implemented it and is pretty much exactly what I envisioned when I refactored the proxy to make different outputs easier!

jonzey231:
Done, pushed. segmenter.py is gone; segment production now runs through ffmpeg's segment muxer, mirroring the fMP4 manager's process lifecycle: a writer thread pipes the ring to ffmpeg stdin with -c copy, and ffmpeg's segment-list journal on stdout (one CSV line per finished segment with measured times) is the completion signal. Each finished segment moves from a private scratch directory into Redis via put_fragment and the file is deleted. Playlist generation stayed in our code as you asked, so we keep control of EXT-X-TARGETDURATION (still frozen per session), media sequence, and discontinuity tags: a ring skip-ahead or an ffmpeg restart tags the next segment with EXT-X-DISCONTINUITY. A small supervisor restarts ffmpeg if it exits or wedges, and the segment chunk TTL rides the configured Redis chunk TTL with a floor sized to the live window plus roll-off, per your suggestion. Manager, keyspace, URLs, teardown, and the frontend/XC wiring are untouched; net effect is about 140 fewer lines. Verified end to end against a synthetic H.264+AAC stream: keyframe-aligned cuts at the target, measured EXTINF, every segment starts with PAT/PMT and decodes standalone. I'll run it against live sources on real Apple and Android clients next and follow up if anything shakes out.

SergeantPanda (changes requested):
So, I did not realize that FFmpeg was incapable of outputting segments via standard out. 

Had I realized that I would not have requested you use FFmpeg for this.

I actually think the original approach is going to be better so we can avoid writing to disk. Otherwise, HLS output would be the only thing writing to disk and if a user uses a network share to run Dispatcharr or has slow storage, we could cause significant performance issues in doing so.

The other options is to write to `/dev/shm` so it still gets wrote to RAM but then I'm concerned with the users that may have 20 streams all outputting HLS filling up the shm (docker default is 64MB which doesn't give a lot of room when saving 4-6 seconds of video).

Sorry about the direction change being incorrect here but I'm going to have to say we don't want FFmpeg here even though it would be more robust in some ways.

Please go back to the segmenter but pay special attention to detecting discontinuity's (probably can have the input tell our worker that a stream switched so it can write it correctly to the manifest (iirc, the hls proxy that is in this code-base I have doing that)). Also ensure all cleanups are properly completed when the last consumer of the hls profile disconnects like we do with fmp4.

jonzey231:
No worries at all, the ffmpeg direction was worth ruling out properly. Reverted it (kept in history so the reasoning stays visible) and the packet-copy segmenter is back with its unit tests. Both of your asks are in.

Discontinuities: the input side now signals the worker directly. StreamManager.update_url is the single point both manual switches and automatic failover funnel through, and on a successful switch it notifies the channel's output managers; the HLS manager resets the segmenter's timeline and the next segment carries EXT-X-DISCONTINUITY. The existing ring skip-ahead marking is unchanged.

Cleanup: playlist and segment requests refresh the client's activity record, the ghost reaper reaps idle clients, and the per-format demand accounting stops the manager, same as fMP4. On top of that, handle_client_disconnect now prunes client-set entries whose metadata hash has expired, and the HLS manager runs its own demand check every 10 seconds, retiring through the same accounting when no live client names the hls output for two consecutive checks. That covers a last poller vanishing without any disconnect event firing.

Verified with a standalone harness driving the real manager against a synthetic TS ring: keyframe-aligned segments, a discontinuity-tagged segment right after a switch signal, and full retirement with all Redis keys cleaned after the last client record lapsed. Running it against live sources on real Apple and Android clients next.

SergeantPanda (dismissed):
Thanks for reverting to the packet-copy segmenter and for the input-side discontinuity signal plus the demand/teardown work. That direction matches what we want (in memory, Redis only, no scratch files).

A few things still worth fixing:

**1. Discontinuity should hard-cut, not keep appending**

`notify_stream_switch` / ring skip calling `flag_discontinuity()` is the right idea, but right now that mostly sets a flag and clears PTS while the current open segment keeps collecting packets. So you can end up with pre-switch and post-switch data in the same segment, and `#EXT-X-DISCONTINUITY` lands on the *next* one.

On discontinuity, please:
- stop appending to the in-progress segment
- discard it (or finish it immediately from bytes already collected, with no post-gap data)
- start fresh on the next keyframe
- put `#EXT-X-DISCONTINUITY` on that new segment

**2. Client re-register drops the output profile**

In `_hls_touch_client`, when the Redis client hash has expired and you recreate it, `output_profile_id` is hardcoded to `""`. If they originally came in as `hls:p3`, after a pause they silently become plain `hls` and read the wrong Redis keys.

Recreating a minimal record so pull clients can stay alive is fine. Blanking the profile is not. Please preserve `output_profile_id` (and ideally don’t invent a fake user/profile if you can avoid it). If the client is truly unknown, making them go back through the normal `stream_ts` handshake is better than guessing.

**3. Idle teardown is fine**

Just to be clear: if the last HLS client stops pulling for long enough that we assume disconnect, shutting down the manager and the channel is what we want. Same idea as other outputs, just with a pull timeout instead of a socket close. No change needed there.

The playlist URL not being able to fully revive a dead channel on its own is okay too. Re-entry via the original stream URL is the correct path.

jonzey231:
Both fixed, and everything below is verified against a live deployment running this branch on the current dev base (the branch still merges clean into dev as of today).

Discontinuity is now a hard cut. flag_discontinuity closes the in-progress segment immediately from the bytes already collected and reports its measured duration, or discards it when there is nothing playable yet, so pre-gap and post-gap data can never share a segment. Collection then waits for the next keyframe, and that segment is the one carrying EXT-X-DISCONTINUITY. Both signal paths use it, the input stream switch and the ring skip-ahead.

Good catch on the re-register. Blanking output_profile_id would silently move an hls:p3 client onto the plain output and the wrong Redis keys. The touch helper no longer recreates an expired record at all: the playlist and segment endpoints return 410 and drop the stale set entry, so the player comes back through the normal stream_ts handshake and gets a complete registration. No invented user or profile anywhere.

Live testing also shook out one race in the new idle teardown, fixed in the last commit: a client tuning in between a failed demand check and the retirement call would be seen by the server's accounting, which correctly kept the manager registered, while the segmenter loop exited anyway, leaving a frozen playlist window. The server is now the arbiter: the loop only exits when the server actually stopped the manager, otherwise it resets and keeps segmenting for the newcomer.

Verification on live channels: TARGETDURATION stays frozen across every reload, EXT-X-START is byte-stable, MEDIA-SEQUENCE advances monotonically at source cadence on multiple channels, segments are PAT/PMT-led and decode standalone under ffprobe, lapsed clients get 410 on both endpoints, and an abandoned session demand-retires with full Redis cleanup and then re-enters cleanly through the stream URL. Segmenter unit tests are at 15 with the hard-cut and discard cases covered.

jonzey231:
One addition since the last round: a fast-start segment ladder for cold channels. A freshly started channel accumulates media at live cadence, so with the 4s cut target a player waited 8-12 seconds for a playable window. The segmenter's first four segments now cut at every keyframe (one GOP each, typically 1-3s) and then the cut target ramps back to normal. Measured on a live deployment: a playable 3-segment window exists about a second after the entry request, and tune-in on a cold channel drops to a few seconds. Steady-state output is unchanged, and every starter EXTINF stays well under the frozen TARGETDURATION.

Also completed the promised client pass: extended playback, stream switches mid-play, idle teardown, and re-entry all verified against live sources on real Apple hardware, with a couple of client-side quirks found and fixed in my app rather than here (notably: AVFoundation refetches the entry URL several times while opening an asset, and each fetch creates a fresh client on the proxy until the ghost reaper collects them; players integrating this output may want to resolve the redirect once and use the client-scoped playlist URL directly).

SergeantPanda:
I haven't forgotten about this one yet. Sorry I've been swamped and getting overwhelmed by PR's that need merging so we don't have conflicts. 

I will be looking at this again hopefully sometime soon.

cmurph1337:
EDIT to add: I would be happy to share the code as a PR if wanted, I just don't want to post something I didn't write and don't fully understand.

Hello! Up front, I'm new to GitHub and I don't write code. I'm not going to pretend to review your work. What I do have is a lot of testing hours: I set out to give Dispatcharr a timeshift buffer like TVHeadend's, and I used this PR as the base to write HLS segments to storage. It works, and after a lot of debugging I'm pretty impressed with it now.

Along the way two bugs turned up that are allegedly in this PR, and they reproduce without any of my (Claude) code involved. I'm confident enough in my testing to say I found them and can reproduce them on demand. I'm not confident enough to explain them, so I'm not going to try.

I do want to share them in case they could help this get merged or fix a bug before being merged. What follows is Claude Code's analysis, in its own words. My sources for all of it were real: ATSC 1.0 MPEG-2 transcoded to HEVC, ATSC 3.0 native HEVC, and AVC from an IPTV stream. 

Thank you for this PR and for Dispatcharr! I really hope I'm not hitting you with slop here but I have tested every step and the results were undeniable.

  ---
  From here down is Claude (Opus 5)

  I'm the assistant that did this work. I'll keep to what I can show.

  Repro. Stdlib only, no Dispatcharr imports. Drop hls_segmenter_repro.py in the repo root and run it — it defaults to apps/proxy/live_proxy/output/hls/segmenter.py, so on your branch it needs no arguments. It builds a synthetic HEVC transport stream in memory that is deliberately unremarkable for broadcast HEVC: an open GOP, where a CRA is followed in decode order by leading pictures whose presentation times are earlier than the CRA's; a PPS before every picture, which is what real encoders emit; and random_access_indicator never set, so the NAL-scan fallback is what decides keyframes. That fallback is the thing under test. Four GOPs × 2.0s at 25fps, target_duration=4.0. Three 2.0s segments is the correct answer.

  On the PR head:

  A. open-GOP HEVC, per-picture PPS, RAI never set
       seg 0: EXTINF    4.000s     564 bytes (3 packets)
       seg 1: EXTINF    0.040s     564 bytes (3 packets)
       seg 2: EXTINF    0.080s     564 bytes (3 packets)
       seg 3: EXTINF    0.040s     564 bytes (3 packets)
       seg 4: EXTINF    4.000s   19176 bytes (102 packets)

  Three packets labelled four seconds.

Cause 1 — PPS is accepted as proof of a keyframe. In starts_keyframe:

# IRAP (16-21) or VPS/SPS/PPS (32-34)
if 16 <= nal_type <= 21 or 32 <= nal_type <= 34:
    return True

VPS and SPS are reasonable keyframe proxies; they're emitted at IRAPs. PPS (34) isn't — it's per-picture on a lot of encoders, so every picture reads as a keyframe and segments get cut mid-GOP. Those segments aren't independently decodable, which is the one property the HLS clients this PR targets need them to have. During the fast-start ladder cut_at is 0.0, so the opening segments become single-picture fragments.

Same area, related: the scan returns on the first matching NAL. A packet can carry a parameter set and a non-IRAP slice, so it's worth deciding from all the NALs in the packet rather than the first hit — a non-IRAP slice is positive evidence that this isn't a keyframe.

Cause 2 — _elapsed treats any negative delta as a 33-bit wrap.

d = pts - start
if d < 0:
    d += PTS_WRAP / PTS_CLOCK

In an open GOP the leading pictures after the CRA legitimately have PTS below the segment's start PTS. A −0.08s delta becomes 95443s, which sails past max_segment_duration and forces a cut. A half-period test — if d < -(WRAP_SECONDS / 2) — separates a real wrap from a few frames of reorder.

What makes this one hard to see is what happens next. _finish_segment sees duration > 4 * self.target_duration and substitutes the nominal target, so the playlist doesn't look broken — it looks like tidy 4.000s segments. Clients sum EXTINF to map a position to a segment, so it surfaces as seek drift and mid-stream repeats rather than as anything obviously wrong. That guard is doing real work as a backstop; it's also what kept this hidden while I was chasing the wrong thing.

Neither fix is sufficient alone. I patched your file two ways to separate them:

  ┌─────────────────┬─────────────────────────────┐
  │      build      │      segment durations      │
  ├─────────────────┼─────────────────────────────┤
  │ PR head, as-is  │ 4.0, 0.04, 0.08, 0.04, 4.0  │
  ├─────────────────┼─────────────────────────────┤
  │ + PPS fix only  │ 4.0, 2.08, 4.0, 2.08, 4.0   │
  ├─────────────────┼─────────────────────────────┤
  │ + wrap fix only │ 0.04, 0.04, 0.04, 0.04, 4.0 │
  ├─────────────────┼─────────────────────────────┤
  │ + both          │ 2.0, 2.0, 2.0               │
  └─────────────────┴─────────────────────────────┘

  One trap, which is the part I'd most want to hand over. Don't take the wrap fix on its own. The bug is accidentally load-bearing. When an upstream restarts its encoder and PTS jumps 1700 → 0, the current code computes 93743s, trips the ceiling, and re-anchors — it
  survives, at the cost of one wrong EXTINF. Make the arithmetic honest without also detecting the reset on purpose and it stalls instead:

  B. upstream restarts its encoder: PTS 1700 -> 0
     PR head, as-is        5 segments: [2.0, 2.0, 4.0, 2.0, 4.0]     survives
     + wrap fix only       2 segments: [2.0, 2.0]                     wedged

  That's part B of the script. I know because I did it: I shipped the wrap fix on its own into their fork, and a channel that had been producing a visible repeat instead stopped producing segments altogether and didn't recover on its own. They caught it in the field and
  reported it; the log showed the recorder alive and the segment count frozen. The fix is to close the current segment when the new PTS is implausibly far below the segment start, rather than leaving the wrap arithmetic to bail you out by accident.

jonzey231:
Hey sarge, good news is it works great and when it starts at the right spot, the channel is up in about a second, way faster than my current remux path. But there are some issues, mainly with AVPlayer but also with Exo.

1. Segment timing doesn't match the playlist. EXTINF says 4.104 but the actual files are 4.28 long and their PTS start 4.10 apart, so every segment overlaps the last one by ~180ms or so. VLC doesn't care. AVPlayer does. It maps playlist time to media time and the drift piles up until it picks a join point that's before any media it has, and then it just sits there forever. This is the big one.

2. TARGETDURATION is 8 with ~4s segments. Players size their live holdback off that (AVPlayer does 3x) so playback end up 24s behind live instead of 12.

3. When a client stops, the stream hangs around ~55s before it gets reaped as a ghost and it counts against stream_limit the whole time. Channel surfing on a limited user stacks those up quick.

4. Every GET of ?output_format=hls mints a brand new client. AVPlayer resolves the redirect once and hands the player the client playlist, but anything that touches the original URL again makes another one.

5. A brand new client's playlist is basically empty (one segment) if you fetch it right away and then fills in. Not a blocker now that I've seen it, just FYI.

Couple things I couldn't figure out from my side: does change_stream keep going in the same client playlist or end it? And when a channel stops, does it write EXT-X-ENDLIST or just stop updating?

SergeantPanda:
Thank you for the feedback!

I will look in to the findings here.

For 3, I thought I had fixed that to be 3 missed segements, I'll take a look into that some more. At the end of the day, this is the issue with HLS that I pointed out in Discord though, there is no persistent connection when using HLS so we can only make assumptions that the client is lost. I am going to experiment a bit with lowering segment sizes to force clients to talk to us more frequently. 
For 4, This was on my radar to fix already but I appreciate you finding it!
For 5, That would be tied to the start behind setting in proxy settings I believe but I'll take a look.

change_stream shouldn't end a playlist, it should append a discontinuity and keep going.
I'm not sure on the ext-x-endlist, if so, I didn't add it. But I don't really see a reason for it, if the channel is shutting down there would probably be no clients.

jonzey231:
Pulled your latest commit into the test container and re-measured. 

1 is fixed. EXTINF 5.005 and the PTS spacing between segments is 5.005, no overlap. Cold channels start fast now, first segment about a second in and eight segments in the playlist by 12s. ESPN2 from cold was up in under a second on AVPlayer.

What still breaks is a new client on a channel that is already running. The client playlist I get starts 14 segments behind the live edge, then jumps forward 9 segments within 6 seconds. AVPlayer fetches the first two, reloads, sees the sequence has leapt past them, and never becomes ready. My retry then mints another client, so a single failed tune shows up as 3 to 5 sessions in your UI. Every tune onto a running channel failed this way; every cold tune worked. That is item 5 and it is now the blocker.

Related: the playlist does not advance at real time. Polling one client playlist every second on a running channel, MEDIA-SEQUENCE sat at 50 for 5s, went to 51, sat 5s, then jumped to 55. On a cold channel segment 1 appeared at 1s and segment 2 at 26s. It looks like the segmenter cuts whatever the proxy hands it and the proxy delivers in 8 to 10 second bursts, so segments land four at a time. Players tolerate a segment of jitter, not 16s leaps. If the segmenter can sit ahead of the proxy's output buffer and publish each segment as it is cut, both this and the start-behind jump probably go away.

2 is unchanged, TARGETDURATION 8 with 4 to 5 second segments.

On your questions: reaping on missed playlist polls rather than missed segment fetches would be safe, players poll the playlist every half to one target duration even when they are buffered ahead. For the duplicate clients, if repeat hits of the original ?output_format=hls URL from the same session within a few seconds reused the last client instead of minting, my retries would stop stacking. And on change_stream, what I saw was the client playlist simply stopped updating after the switch, no discontinuity tag, so the switch is not coming through on the HLS side yet. Happy to grab the playlist sequence across a switch if that helps.

PR #1352, opened 2026-06-14
Plugins: split manifest root_url into download_base_url and metadata_base_url

sethwv:
> [!NOTE]
I'm going to wait a release or two before merging the plugin-repo's corresponding  https://github.com/Dispatcharr/Plugins/pull/127 so that we don't break plugin installing for too many users.

## Description

Adds `download_base_url` and `metadata_base_url` as optional alternatives to `root_url` in the plugin repo manifest. This lets repo authors serve metadata (manifests, icons) and release zips from different origins without having to use absolute URLs for everything.

Resolution priority:
- `manifest_url`, `icon_url` - `metadata_base_url` then `root_url`
- `latest_url`, version `url` fields - `download_base_url` then `root_url`

When `metadata_base_url` is explicitly set and a plugin entry has a `manifest_url` but no `icon_url`, the icon is assumed to be `logo.png` in the same directory as the per-plugin manifest. The existing GitHub `registry_url` icon fallback is unchanged for manifests that do not use the new fields.

All changes are backward compatible. Manifests using only `root_url` behave identically to before.

## How was it tested?

Manually tested against the official plugin repo manifest (root_url only) to confirm existing icon and URL resolution is unchanged. Traced all three call sites in `api_views.py` (`_invalidate_plugin_detail_cache`, `AvailablePluginsAPIView`, `PluginDetailManifestAPIView`) through the new helper for both old and new manifest shapes.

## Checklist

### DIFF
--- a/Plugin_repo.md
+++ b/Plugin_repo.md
@@ -120,12 +120,14 @@ If the name contains any of these, the repo will be rejected on add and skipped 
 
 ### Top-Level Metadata
 
-| Field           | Required | Description                                                                                                                                                                   |
-| --------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `registry_name` | **Yes**  | Display name for the repo. Must not contain words like "official" or "dispatcharr" that could be mistaken for an official repo (see [Name Restrictions](#name-restrictions)). |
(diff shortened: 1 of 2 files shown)


SergeantPanda:
Thank you!

PR #1367, opened 2026-06-19
fix(m3u): align auto-sync rename preview with the live rename

### LINKED ISSUE #1332: [Feature]: Display Auto Channel Sync 'Find and Replace' name transformations for numbered capture groups

Currently, the editing dialog for auto channel sync displays a list of rename matches based on the 'Find and Replace'. However, it uses the capture group placeholder of `$1` when displaying the matches:

Image

Ideally, instead of showing the placeholder `$1` as the 'renamed' match, it should display the actual output of the regex capture group.

In this example, it would show `High Limit Racing at Eagle @ Jun 9 7:00 PM :Flo Racing 05 -> High Limit Racing at Eagle`.

### Additional context

Unsure if this was intended or a bug, so submitted as a feature request for now.

CodeBormen:
## Description

The Auto Channel Sync "Find and Replace" preview did not accurately reflect the rename the sync performs. This PR reconciles the preview with the live rename so the preview can be trusted while authoring patterns, and hardens the rename against input that could abort a sync.

Three distinct divergences are fixed:

- Literal backreferences. The preview rendered the literal `$1` placeholder instead of the substituted capture group, because the regex-preview endpoint passed the JS-style replacement straight into Python's `re.sub`, which only honors `\1`. The live rename already converted `$1` to `\1`, so the two disagreed.
- Regex engine mismatch. The preview compiles patterns with the `regex` module while the live rename compiles with the standard-library `re` module. A pattern the `regex` module accepts but `re` rejects (for example the quantified anchor `^*`) previewed a transformed name, while the live rename hits a compile error, skips the rename, and keeps the original name. The preview promised an output the sync would never produce.
- Over-length rename aborting the sync. A rename that expanded a name past the `Channel.name` column length failed the bulk insert and aborted the entire account sync, producing no channels at all, while the preview showed the full over-length name.

The fixes:

- Extract the backreference conversion into a single shared helper, `convert_js_numbered_backreferences` in `apps/m3u/utils.py`, used by both the live rename and the preview. It is byte-for-byte equivalent to the conversion the rename already performed, so real rename output is unchanged.
- Validate the preview's find pattern with the standard-library `re` module, so it reports the same rejection the live rename would hit instead of previewing an impossible transform. Execution still uses the `regex` module for its per-call timeout, which bounds catastrophic backtracking on the interactive endpoint.
- Cap the rename result at the `Channel.name` column length in both the sync and the preview, so an over-length result can no longer abort the sync and the preview shows the same bounded value.

The include and exclude filter previews already use the same engine as the sync, so only the find/replace path needed aligning. No model changes, no migration, no new endpoints, and no frontend changes.

## Related Issue

Closes #1332

## How was it tested?

Each divergence was reproduced first, then covered by tests added in this PR under `apps/m3u/tests/`:

- `test_find_replace_applies_numbered_capture_group` asserts the preview substitutes the capture group rather than echoing the literal `$1`.
- `test_preview_after_matches_live_sync_rename` runs the live sync rename and asserts the preview output equals the channel name the sync writes.
- `test_find_pattern_rejected_by_sync_engine_reports_error` and `test_preview_and_sync_agree_on_sync_rejected_pattern` assert a pattern the sync engine rejects (such as `^*`) returns an error in the preview and leaves the synced channel name unchanged, rather than previewing a transform.
- `test_overlong_rename_is_bounded_not_aborting_sync` asserts an over-length rename produces truncated channels instead of aborting the sync.
- `test_rename_preview_parity.py` is a differential test that runs a broad matrix of find/replace strategies (capture groups, alternation, lookaround, non-capturing groups, character classes, Unicode and emoji names, invalid group references, and over-length output) across diverse names through both the live sync rename and the preview endpoint, asserting the preview predicts the rename in every case.

All of the above fail on the unpatched code and pass with the fix, and the existing auto-sync and regex-preview test coverage continues to pass.

## Checklist


SergeantPanda:
I need to look at this one a bit more. I don't think we've had to do this for other areas that we use regex. Generally I've been switching the backend to use regex instead of re which more closely matches js.

CodeBormen:
It wouldn't be hard for me to rewrite this from `re.sub` to `regex.sub` if you've been intentionally avoiding re. I only chose this route because it had the least impactful behavior, but I'm in agreement that standardizing on regex is probably the better call. Happy to rework into that if that's preferred.

SergeantPanda:
Yeah, it's already in the codebase, for most of the code, you just change it from `re` to `regex` without needing to change anything else. Some of the frontend changes may not need to happen then.

CodeBormen:
Sounds good, I'll take a look at it within the next day or so

SergeantPanda:
Thank you. Sorry it took so long to get to these. I've been swamped trying to fix these memory/performance issues.

I think I finally have them in a good spot.

CodeBormen:
Don't sweat it I see you've been busy 😂

CodeBormen:
Ready for another look!

SergeantPanda (commented):
This is MUCH better. Thank you!

Only things I can see that could be improved later on would be to compile name_regex_pattern once per group instead of per stream to improve efficiency slightly. Also adding the ability for convert_js_numbered_backreferences to handle named groups (like bulk_regex_rename does).

Neither are blocking this PR and can be follow-ups if you're interested.

Thanks again!

CodeBormen:
If you want me to add those tonight quick before merge I can do so

CodeBormen:
Nvm I was quick but not quick enough 😂

SergeantPanda:
Oh shoot, sorry!

CodeBormen:
All good

PR #1368, opened 2026-06-20
fix(channels): account for group_override in the auto-sync range-conflict check

### LINKED ISSUE #1331: [Bug]: Auto Channel Sync "Range conflict" warning is not accurate
### Describe the bug

When viewing groups for a M3U Account, there is a warning that displays if there are auto-sync channel range conflicts. However, this warning is either inaccurate, or does not take into account channels that are coming from the auto-sync setup itself. 

1. Configure Auto Channel Sync with a start of 3210 and 3298.

Image

2. Ensure no channels exist in the designated range (Except channels created by the auto-sync itself.

Image

3. Observe warning of range conflict.

Image

### Dispatcharr Version

0.26.0

### Docker Version

N/A

N/A

### Client Information

N/A

CodeBormen:
## Description

The Auto Channel Sync "Range conflict" warning on the M3U group settings falsely fired on a group's own auto-sync channels when the group used a channel-group override.

The warning scans for channels already occupying the configured number range and classifies each as either this config's own auto-sync output or a real conflict. That classification compared each occupant's channel group against the source group being configured. When a group sets a `group_override`, auto-sync creates its channels in the override target group, so those channels did not match the source group and were misclassified as a conflict, even though they are the expected output of the very config being edited.

This change resolves the effective target group for the comparison: the override target when `group_override` is set, otherwise the source group. The config's own output is then recognized and no longer triggers the warning. The warning is not suppressed for override groups - the check is simply performed against the group the channels actually land in, so genuine conflicts (manual channels, channels from another account, channels in a different group, or user-pinned numbers) within the range still surface. The backend `numbers-in-range` endpoint already reported the channel's actual group, so the change is limited to the frontend classification.

No model changes, no migration, and no new endpoints.

## Related Issue

Closes #1331

## How was it tested?

Reproduced first: with a group override configured, the auto-sync's own channels in the configured range were classified as a conflict, while the same setup without an override classified them correctly.

Added tests:
- Frontend unit tests for the new effective-group resolver: it returns the source group with no override, the override target when set (coercing a string-stored value), and falls back to the source group when the override is blank. A regression guard confirms an occupant in the override target group is recognized as the config's own output rather than a conflict.
- An over-suppression guard confirms that, even with a group override, a manual channel, a channel from another account, a channel in a different group, or a user-pinned number in the range is still flagged as a conflict.
- A backend test asserting the numbers-in-range endpoint reports the override target group for an override-routed auto-created channel, locking the contract the frontend classification relies on.

The existing range-lookup and group-settings test suites continue to pass.

## Checklist

### DIFF
--- a/apps/m3u/tests/test_sync_correctness.py
+++ b/apps/m3u/tests/test_sync_correctness.py
@@ -644,6 +644,34 @@ class NumbersInRangeLookupTests(TestCase):
         )
         self.assertFalse(occupant["has_channel_number_override"])
 
+    def test_group_override_channel_reports_target_group(self):
+        # When auto-sync routes channels into a different group via
+        # group_override, the occupant's channel_group_id is the override
+        # target, not the source group being configured. The frontend relies
+        # on this to recognize override-routed channels as the config's own
+        # output (effectiveSyncGroupId), so the warning does not flag them.
+        account = _make_account()
+        source = _make_group(name="SourceGrp")
+        target = _make_group(name="TargetGrp")
+        Channel.objects.create(
+            name="Routed",
+            channel_number=3210,
+            channel_group=target,
+            auto_created=True,
+            auto_created_by=account,
+        )
+        client = self._client()
+
+        response = client.get(
+            "/api/channels/channels/numbers-in-range/?start=3210&end=3210"
+        )
+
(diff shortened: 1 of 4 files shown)


SergeantPanda:
Thank you for this!

PR #1378, opened 2026-06-22
fix(m3u): abort XC refresh on empty provider fetch to prevent channel wipe

### LINKED ISSUE #1377: XC empty provider fetch deletes the entire auto-created channel lineup
## Summary

On a routine M3U refresh, when an **Xtream Codes** provider returns no live streams, Dispatcharr marks every stream stale and then auto-channel-sync **deletes the account's entire auto-created channel lineup**. A transient upstream hiccup (the provider briefly returning an empty stream list) therefore wipes thousands of channels until the next successful refresh recreates them.

## Steps to reproduce

1. An XC account with one or more enabled groups that have `auto_channel_sync` enabled (so channels are auto-created from streams).
2. The provider returns an empty live-stream list on a scheduled refresh (transient outage, rate-limit, auth blip, or the API momentarily returning `[]`).
3. Observe the logs:

```
WARNING apps.m3u.tasks No live streams returned from XC provider
WARNING apps.m3u.tasks No streams collected from XC groups
INFO    apps.m3u.tasks Marked N streams as stale for account <id>
INFO    apps.m3u.tasks Auto channel sync complete for account <name>: 0 created, 0 updated, <thousands> deleted, 0 failed
```

Every auto-created channel for the account is gone.

## Root cause

In `apps/m3u/tasks.py`, `collect_xc_streams()` returns an empty list on any of: the provider returning no live streams, the fetch raising, or no enabled category matching. In `_refresh_single_m3u_account_impl()` the XC branch only logs a warning on that empty result and **falls through** to stale-marking + `sync_auto_channels()`:

```python
all_xc_streams = collect_xc_streams(account_id, filtered_groups)
if not all_xc_streams:
    logger.warning("No streams collected from XC groups")
else:
    ... # process streams
# ... then unconditionally: mark stale, cleanup, sync_auto_channels(...)
```

With nothing "seen" this refresh, `sync_auto_channels` deletes all non-hidden auto-created channels via its per-group "no streams remaining" branch (`if not has_streams:`). Note this deletion is **not** gated by the account's `orphan_channel_cleanup` setting, so no account option prevents it.

The standard (non-XC) M3U path already guards this case: an empty/failed download sets the account to ERROR and returns before any pruning. The XC path is missing the equivalent guard.

## Proposed fix

Abort the XC branch when `collect_xc_streams()` returns empty: set the account status to ERROR and return before stale-marking/auto-sync, mirroring the standard-path empty-download guards already in the same function. A transient empty fetch then leaves the existing streams and channels intact, and the next successful refresh proceeds normally.

I have a branch with the fix plus tests and will open a PR against `dev` referencing this issue.

Jacob-Lasky:
## Description

Guards the Xtream Codes refresh path against a transient empty provider response that otherwise deletes the account's entire auto-created channel lineup.

When `collect_xc_streams()` returns an empty list (provider returned no live streams, the fetch raised, or no enabled category matched), `_refresh_single_m3u_account_impl()` previously only logged a warning and fell through to stale-marking and `sync_auto_channels()`. With nothing seen this refresh, auto-sync deletes all non-hidden auto-created channels via its per-group "no streams remaining" branch (which is not gated by `orphan_channel_cleanup`). A brief upstream hiccup thus wipes thousands of channels.

This change aborts the XC branch on an empty result: it sets the account to `ERROR` and returns before stale-marking and auto-sync, mirroring the standard-path empty/failed-download guards already present in the same function. The outer `finally` still runs its memory cleanup, and the task lock is still released by the wrapper. The existing `else:` block is kept (rather than de-indented) to keep the diff minimal.

## Related Issue

Closes #1377

## How was it tested?

Ran the suite inside a `ghcr.io/dispatcharr/dispatcharr:dev` container against its embedded Postgres:

- New tests in `apps/m3u/tests/test_xc_empty_fetch_guard.py`, both pass:
  - `test_empty_xc_fetch_aborts_before_sync_and_preserves_channels` — with `collect_xc_streams` stubbed to `[]`, asserts `sync_auto_channels` is never called, the pre-existing auto-created channel still exists, the stream is not marked stale, and the account status is `ERROR`.
  - `test_non_empty_xc_fetch_still_runs_sync` — a non-empty fetch still reaches `sync_auto_channels` and the account ends `SUCCESS`, so the guard does not over-trigger.
- Full `apps/m3u` suite: **127/128 pass**. The single failure, `test_sync_correctness.Migration0037DemoteOrphansTests.test_orphan_with_no_streams_is_demoted_not_deleted`, fails identically on unmodified `dev` (it calls the `0037` migration backfill with `schema_editor=None`, raising `AttributeError: 'NoneType' object has no attribute 'connection'`). It does not touch `tasks.py` and is unrelated to this change.

## Checklist


SergeantPanda (approved):
This looks good. Thank you. I have considered doing this for a while but hadn't gotten around to it.

PR #1386, opened 2026-06-24
show lineup name in EPG channel dropdown for Schedules Direct sources

sethwv:
## Description

Adds a `lineup` field to `EPGData` to track which Schedules Direct lineup a station came from. The EPG channel dropdown in the channel edit form now renders options as `Name (ID) [Lineup Name]` for SD sources. Non-SD sources are unaffected (lineup is null, bracket suffix is omitted).

Also improves error handling on the SD token endpoint: the response body is now parsed before raising, so app-level errors like `SERVICE_OFFLINE` (code 3000) show a human-readable message instead of a raw HTTP exception. Added explicit handling for codes 3000 and 4003.

## How was it tested?

- Triggered a Schedules Direct EPG sync and confirmed `lineup` is populated on `EPGData` rows.
- Opened the channel edit form, selected an SD EPG source, and confirmed the dropdown shows `Name (ID) [Lineup Name]`.
- Confirmed XMLTV sources render unchanged.

image

## Checklist

### DIFF
--- a/apps/epg/migrations/0027_epgdata_lineup.py
+++ b/apps/epg/migrations/0027_epgdata_lineup.py
@@ -1,0 +1,16 @@
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('epg', '0026_epgsourceindex'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='epgdata',
+            name='lineup',
+            field=models.CharField(blank=True, max_length=255, null=True),
+        ),
+    ]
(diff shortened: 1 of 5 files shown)


SergeantPanda:
I think this is important for sure. I am just not sure what the best way to go about this is yet. Debating on every lineup from SD being it's own EPG Source instead. Other option would be create a custom_properties for EPG Data so we can use it for other EPG types (like saving the GNID from Jesse's guide for example)

SergeantPanda:
Sorry I haven't had a chance to look at this one thoroughly yet.

sethwv:
Kicking this can down the road in that case that SD sticks around long term.

PR #1398, opened 2026-06-30
Add VOD failover logic for M3U relations

### LINKED ISSUE #1385: [Feature]: VOD failover to other accounts when the selected account is at capacity

Yes. When the same movie exists on multiple separate M3U accounts, the native VOD proxy selects one account and returns HTTP 503 if that account is at capacity. It never tries the other accounts that carry the same title, even when they are idle.

Live TV already fails over across accounts, and the XC output path already selects an available account. The native VOD proxy is the only path that stops at the first account and gives up, so a second viewer gets a 503 while idle accounts sit unused.

When the selected account for a VOD title is at its connection limit, the proxy should move to the next account that has a relation for the same movie, in priority order, and stream from the first one with a free slot. It should only return 503 once every account carrying that title is genuinely full.

This is the same capacity-aware selection that already exists for live failover, applied to the VOD proxy path (apps/proxy/vod_proxy/views.py), which currently selects one account and stops instead of continuing through the remaining relations for the same title.

### Additional context

**Alternatives tried:**
- Manual versions in the client (one STRM per provider, so each account shows as a separate version) works, but it is manual. The user has to switch versions by hand when the default account is busy, rather than playback finding a free account on its own.
- Assigning the accounts to a single Server Group did not change the VOD proxy behavior. It still selects one account and 503s when full, without trying the others in the group.
- A single multi-connection account avoids the problem but defeats the purpose of running multiple providers.

**Evidence the other accounts go unused.** Same movie, which exists on three separate accounts (FirstIPTV, SecondIPTV, ThirdIPTV).

First request selects FirstIPTV and takes its only slot:
```
[CONTENT-FOUND] Movie: EN - A Beautiful Planet 4K (2016) (ID: 375)
[PROVIDER-SELECTED] Using provider: FirstIPTV (priority: 0)
[VOD-ACCOUNT] Using M3U account: FirstIPTV
[PROFILE-SELECTION] Selected profile 2 (FirstIPTV Default): 0/1 connections
```

Second request for the same movie, with FirstIPTV now full, returns 503 instead of using an idle account:
```
[CONTENT-FOUND] Movie: EN - A Beautiful Planet 4K (2016) (ID: 375)
[PROVIDER-SELECTED] Using provider: FirstIPTV (priority: 0)
[VOD-ACCOUNT] Using M3U account: FirstIPTV
[PROFILE-SELECTION] All profiles at capacity for M3U account 2, rejecting request
[VOD-ERROR] No suitable M3U profile found for movie 19e39531-7e8b-43ba-9291-1a114ce9ae83
```

Only FirstIPTV was busy. SecondIPTV and ThirdIPTV were idle and were never attempted. The connection counter is accurate: the first stream legitimately took the slot (0/1 to 1/1) and the second correctly saw it full, so the 503 is the proxy declining to consider the other accounts, not a stale count.

**Reproduce:**
1. Configure 3 or more XC accounts, each a different provider login, each with max connections = 1.
2. Pick a movie that exists on all of them.
3. Start a first VOD stream and let it hold the connection: `curl -sL -o /dev/null "http://SERVER:9191/proxy/vod/movie/<uuid>" &`
4. Start a second for the same movie: `curl -sL -o /dev/null -w "%{http_code}\n" "http://SERVER:9191/proxy/vod/movie/<uuid>"`

Observed: the second returns 503. Desired: it streams from one of the idle accounts that also carry the movie.

**Environment:** Dispatcharr 0.27.0, bare metal / LXC (Proxmox), uWSGI multi-worker. Accounts are multiple XC sources, different provider logins, max connections = 1 each, equal priority (0). Tested directly against /proxy/vod/movie/<uuid> to isolate from the client; normal frontend is Emby via HDHomeRun emulation.

francescodg89-crypto:
## Description

Implemented VOD failover logic in `apps/proxy/vod_proxy/views.py`. Previously, `stream_vod()` and `head_vod()` resolved content to a single relation (the highest-priority one via `.first()`) and returned HTTP 503 if that account was at capacity, without trying any other account that carries the same title.

This change adds a helper, `_get_all_relations_ordered()`, that returns all active M3U relations for the content (Movie, Episode or Series — all use the `m3u_relations` related_name), ordered by account priority, with the already-selected preferred relation placed first. The selection block in both `stream_vod()` and `head_vod()` now iterates over these relations and uses the first account whose profile pool has spare capacity (via the existing `_get_m3u_profile()`). It returns 503 only when every account carrying the title is genuinely full.

This mirrors the failover behaviour that already exists for live channels, and addresses the single-account-bottleneck described in #1385.

## Related Issue

Closes #1385

## How was it tested?

Tested on a live 0.27.1 install with 4 Xtream accounts each limited to `max_streams = 1`, all carrying the same titles. Before the change, a second concurrent VOD request returned 503 even though idle accounts had the same content. After the change, the second request succeeds and the logs show the failover walking to the next provider:

```
[VOD-FAILOVER] Account <provider-a> at capacity, trying next provider
[VOD-FAILOVER] Selected account <provider-b> (priority 0)
```

Verified that:
- a single available account still streams normally (no regression),
- 503 is still returned when all accounts carrying the title are full,
- connection counters increment/decrement correctly on stream start and client disconnect,
- movies, episodes and series all resolve relations through the new helper.

## Checklist

- [x] Frontend: ESLint and Prettier pass cleanly (no frontend changes)
- [x] No console.log, print(), debug statements, or commented-out code is left in the diff

> Disclosure: I developed this change with AI assistance, but I have reviewed and understand every line, tested it on my own production setup, and can explain and adjust any part on request.


SergeantPanda (changes requested):
Thank you for the PR.

One improvement I'd like to see: avoid the second DB hit from `_get_all_relations_ordered()`. `_get_content_and_relation()` already queries active relations with the same priority ordering. Could we have it return the full ordered candidate list (with the preferred relation first) in that single query? `_get_all_relations_ordered()` could then become a small in-memory helper for preferred-first ordering/dedup, or be inlined entirely. That way the initial connection path only hits the database once.

francescodg89-crypto:
Thanks for the review — done. _get_content_and_relation() now materialises the active, priority-ordered relations once (single DB query) and returns that list alongside the selected relation; preferred-stream / preferred-account / default selection is done in memory on that list. _get_all_relations_ordered is gone, replaced by _order_candidates(), a pure in-memory helper that only moves the preferred relation to the front and de-duplicates. Updated the tests accordingly, including one asserting the helper performs no ORM access. Verified on my production setup — failover still walks to the next provider when an account is at capacity.

SergeantPanda:
Thank you! I'll review again tomorrow.

SergeantPanda (approved):
This seems to be working perfectly. I did add a couple tests and modify a couple as well to test the new functionality.

Thank you!

PR #1417, opened 2026-07-06
fix(environment): re-verify cached public IP in the background (#1395)

### LINKED ISSUE #1395: [Bug]: Public IP address doesn't change in UI when VPN reconnects
### Describe the bug

When the public IP address for the Dispatcharr instance changes, like with a Gluetun VPN endpoint change, the IP address should update in the WebUI. (It was doing it up until either 0.26.x or 0.27.x.)

1. Connect Dispatcharr instance behind Gluetun VPN container, record IP address.
2. Run a VPN reconnect inside of the Gluetun VPN container (without restarting the container).
3. curl whatismyip.com or similar to find your new public ip address.
4. Validate it is different in the WebUI

IP as always shows in Dispatcharr UI:
Image

Validated by what showing before resetting VPN:
Image

What the Dispatcharr UI should be showing:
Image

### Dispatcharr Version

0.27.1

### Docker Version

Docker version 29.5.2, build 79eb04c

`  dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:0.27.1
    container_name: dispatcharr
    network_mode: service:gluetun
    volumes:
      - /local/path:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
    restart: unless-stopped
    depends_on:
      - gluetun
    deploy:
      resources:
          reservations:
              devices:
                  - driver: nvidia
                    count: all
                    capabilities: [gpu]`

### Client Information

n/a

floppy-disk:
Note: I understand that the 1 hour IP address update in 16cb9edc may have been intentional, in which case you can reject this PR. 
----
The IP lookup result introduced in 16cb9edc is cached in Redis for an hour and nothing ever re-checked it within that window, so after a VPN reconnect the UI kept serving the stale address - even across page reloads and container restarts.

Serve the cached result instantly as before, but when it was last verified more than a minute ago, kick the existing background lookup (guarded by the existing lock) on the next /api/core/environment/ call. Its existing websocket push then corrects all connected clients. A failed re-check keeps the last known good result instead of clobbering it with None and blanking the sidebar.

No polling: lookups still only happen on page load/login, now capped at one per minute (the pre-regression code did a blocking lookup on every call).

## Description

The IP lookup result introduced in 16cb9edc is cached in Redis for an hour and nothing ever re-checked it within that window, so after a VPN reconnect the UI kept serving the stale address - even across page reloads and container restarts.

Serve the cached result instantly as before, but when it was last verified more than a minute ago, kick the existing background lookup (guarded by the existing lock) on the next /api/core/environment/ call. Its existing websocket push then corrects all connected clients. A failed re-check keeps the last known good result instead of clobbering it with None and blanking the sidebar.

No polling: lookups still only happen on page load/login, now capped at one per minute (the pre-regression code did a blocking lookup on every call).

## Related Issue

Closes #1395

## How was it tested?

- New unit tests in `tests/test_ip_lookup.py` (8 cases) covering: stale cache serves the cached value and spawns exactly one background refresh; fresh cache spawns none; an existing lock suppresses a duplicate spawn; a failed re-check preserves the last known result and does not push a websocket update; a failed *first* lookup still caches and pushes so the frontend skeleton resolves; the cache-miss and disabled-setting paths are unchanged. `python manage.py test tests.test_ip_lookup` — all pass.
- Full backend suite run against PostgreSQL 17 + Redis 7 in Docker, on unmodified `dev` and again with this change: identical results apart from the 8 new passing tests.
- End-to-end on a real deployment behind gluetun (`network_mode: service:gluetun`): triggered a VPN reconnect inside the gluetun container and confirmed the WebUI shows the new public IP on the next page load, where 0.27.1 kept the old address for up to an hour.

## Checklist

- [x] Backend: migrations are included if any models were changed *(no model changes)*
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema *(no new endpoints)*
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) *(no frontend changes)*


SergeantPanda (changes requested):
Thanks for this. The stale-while-revalidate approach looks right for #1395, and keeping the last known IP on a failed re-check is a good call.

One ask before merging: can we drop the separate `_IP_FRESH_KEY` and put a `verified_at` timestamp on the existing cache payload instead? We already have the result key plus the lock; I'd rather not add a third Redis key just for the 60s revalidate window.

Something like:
- store `verified_at` (e.g. `time.time()`) in the cached result
- on `/environment/`, if the cache hit is older than 60s, kick the background lookup the same way you do now
- on a failed re-check, still bump `verified_at` without clobbering the IP fields so we don't retry on every request
- strip `verified_at` from the API response and websocket payload (and anything `consumers.py` pushes on connect) so the frontend doesn't grow an unused field

Missing `verified_at` on older cache entries can just count as stale and trigger one revalidate.

Happy to merge once that lands. Tests look solid otherwise.

floppy-disk:
Sorry for the delay, was on vacation! I folded your feedback in using time rather than another Redis key and validated that it updated in my dev instance as expected, which it did. Let me know if you need anything else before merging, and thanks for the project!

SergeantPanda (approved):
Looks good, thank you for making the changes! I made a couple of very minor changes but overall it's solid and I think will help out other users using VPN's.

PR #1420, opened 2026-07-07
fix(db): reap leaked gevent-pool connections that wedge workers under streaming load (#1418)

### LINKED ISSUE #1418: [Bug]: player_api.php / XC API hangs (504) under streaming load — PostgreSQL gevent-pool connection leak on live-stream teardown drains the per-worker pool
### Describe the bug

**Summary.** Under normal use (watching + switching live channels), the whole Xtream-Codes / streaming
API periodically hangs: `player_api.php` requests block for ~60s and return **504**, while
`/api/core/version/` keeps returning **200 in ~5ms**. It recurs every few hours and a **container
restart clears it** for a while. This is **not** CPU-bound.

**Root cause (traced in the 0.27.2 image).** Each live-stream request checks a connection out of the
custom gevent DB pool (`dispatcharr/db/backends/postgresql_psycopg3/pool.py`, `MAX_CONNS=8` per uWSGI
worker, `dispatcharr/settings.py`), and on the **stream teardown / channel-switch** path the connection
is **never returned to the pool** (`put()` is not called) — it leaks. Because leaked connections are
still counted against `maxsize` but are no longer in the pool queue, once 8 leak per worker
`DatabaseConnectionPool.get()` blocks forever on an empty queue. Any request that needs the DB then
hangs:
- `player_api.php` → `xc_player_api` → `xc_get_info` → `get_user_active_connections()` **and** the XC
  auth (`xc_get_user`) need a DB connection → they block.
- `/api/core/version/` needs no DB connection → it stays fast. **That asymmetry is the tell.**

**Why `DATABASE_POOL_CONN_MAX_LIFETIME` does not save it.** `_connection_expired()` is only evaluated
inside `pool.get()` and `pool.put()`. There is **no background reaper**, so a leaked connection (which by
definition never re-enters `get()`/`put()`) is never aged out or reclaimed. Lowering the lifetime only
recycles *healthy idle* connections faster; it does nothing for leaked ones.

**Expected behavior.** A live-stream teardown (client disconnect, channel switch, failover exhaustion)
must return its pooled DB connection (or the pool needs to reap connections abandoned by dead greenlets).
The XC/streaming API should not degrade to 504 as a function of how many streams have been started+stopped.

**Severity.** Recurring full outage of the streaming/XC API for all clients until an operator (or a
watchdog) restarts the container. More live channels / more failover streams per channel = faster onset.

### Related issues (searched — this is NOT a duplicate of any of them)
- **#1343** (closed) & **#1345** (closed) — the *persistent* pool growing unbounded → OOM / `/dev/shm`
  exhaustion. Those motivated the current **bounded** pool (`MAX_CONNS` + `max_lifetime`). This report is a
  **new gap in that bounded pool**: it leaks connections on live-stream teardown and never reaps them.
- **#1338 / #1337** (closed) — "M3U refresh can leave a worker DB connection in INTRANS until restart."
  Same DB-connection-lifecycle family, **different code path** (M3U refresh). This is the **live-stream /
  live-proxy teardown** equivalent, still present in 0.27.2.
- **#1396** (open, EPG regen) & **#993** (closed, dead upstream) — **same symptom** (gevent worker hangs,
  login/API/HDHR freeze) but **different root cause** (CPU-bound build / blocked upstream socket). Ours is
  neither CPU-bound nor upstream-blocked — it is pure DB-pool starvation (evidence below).
- Possibly contributing: **#408** (open) "Channel Shutdown Delay never shutting down an abandoned channel" —
  a stream greenlet that never tears down would also never return its pooled connection.

### Proposed fix
1. **Targeted:** guarantee the live-proxy stream request returns its pooled DB connection on **every** exit
   path — client disconnect, channel switch, failover exhaustion, and abnormal greenlet death — not just the
   happy path (e.g. wrap the stream lifecycle in `apps/proxy/live_proxy/views.py` / `server.py` so
   `close_old_connections()` — which the geventpool intercepts as `put()` — always runs in a `finally`). The
   smoking gun is the pool staying pinned at 8/8 *after all streams end*.
2. **Defense-in-depth:** give the custom pool (`dispatcharr/db/backends/postgresql_psycopg3/pool.py`) a
   background reaper that closes connections older than `max_lifetime` **even when checked-out/abandoned**.
   Today `_connection_expired()` is only consulted inside `get()`/`put()`, so `DATABASE_POOL_CONN_MAX_LIFETIME`
   cannot reclaim a leaked connection. A reaper would also have caught the #1338 INTRANS leak — one safety net
   for the whole family.

### Current workaround (operator-side; neither is a fix)
- Raise `MAX_CONNS` (we use 24) — more headroom delays onset ~proportionally, but the pool still leaks to
  exhaustion eventually.
- Healthcheck that probes `player_api.php` (not just `/api/core/version/`, which stays green during the
  partial hang) + `willfarrell/autoheal` to `docker restart` on the hang — auto-recovers each outage in
  ~45s–2min. A restart is the only thing that clears the leaked connections.

Reproduced deterministically on a quiet instance (no other viewers) in ~30s:

1. Pick ~24 live channels (any; channels with several failover streams leak fastest because each
   failover attempt holds a connection through the upstream fetch).
2. Fire concurrent stream-starts against the proxy and let them run. Use ~24 of your own live-channel
   UUIDs (from the Channels page, or `GET /api/channels/channels/`):
   ```bash
   HOST=localhost:9191
   UUIDS="<paste ~24 space-separated live-channel UUIDs>"
   for u in $UUIDS; do curl -s -o /dev/null --max-time 22 "http://$HOST/proxy/ts/stream/$u" & done
   ```
3. Poll auth + the pool while the streams run (substitute your own XC username/password):
   ```bash
   # auth latency (needs a DB connection)
   curl -s -o /dev/null -w '%{time_total}s/%{http_code}\n' \
     'http://localhost:9191/player_api.php?username=YOUR_XC_USER&password=YOUR_XC_PASS'
   # per-worker pool occupancy (run inside the container)
   echo "SELECT application_name,state,count(*) FROM pg_stat_activity WHERE application_name LIKE 'Dispatcharr-uwsgi%' GROUP BY 1,2;" | psql -U dispatch -d dispatcharr
   ```
4. Observe: each uWSGI worker climbs to **8/8** and `player_api` auth jumps from ~0.02s to a **15s+
   timeout / 60s 504**.
5. **The proof it is a leak, not transient load:** stop the load and wait for all clients to drop
   (`live:channel:*:clients:*` = 0). The pool **stays pinned at 8/8** and `player_api` **keeps timing
   out until the container is restarted.**

**Negative control:** 50 concurrent `player_api.php?action=get_live_streams` (heavy but *short* DB reads)
does **not** reproduce it — those connections recycle in ~0.25s and the pool never exceeds ~3/worker.
Only **long-held** (streaming) connections leak, which points squarely at the stream-teardown path.

### Evidence captured during a wedge

| Signal | Healthy | During wedge |
|---|---|---|
| `/api/core/version/` | 200, ~0.005s | 200, ~0.005s (unaffected) |
| `player_api.php` (auth) | 200, ~0.02s | **000/504, 15–60s timeout** |
| uWSGI pool census (`pg_stat_activity`) | ~2–6 idle/worker | **8/8 idle/worker, stays 8/8 after streams=0** |
| Worker CPU (utime+stime jiffies, 15s window) | idle | **frozen / barely moving** (e.g. 27402→27495) — **not CPU-bound** |
| Worker state | `S`, `wchan=0` | `S`, `wchan=0`, `load~2` (I/O-wait) |
| Postgres | idle | **idle, 0 locks, 0 idle-in-transaction** |
| Redis | fast | `PING` + `SCAN` fast (ruled out) |

#### Captured reproduction output (the smoking gun — pool stays 8/8 after streams drop to 0)
Provider host/token censored; these are only pool counts + curl timings, no URLs:
```text
=== baseline ===
  Dispatcharr-uwsgi-242|idle|6   Dispatcharr-uwsgi-243|idle|4
=== firing 24 concurrent stream-starts, polling player_api auth + pool ===
[t+1s]  player_api_auth=0.16s/200   streams=2    uwsgi-242|idle|8  uwsgi-243|idle|8
[t+4s]  player_api_auth=0.05s/200   streams=17   uwsgi-242|idle|8  uwsgi-243|idle|8
[t+8s]  player_api_auth=15.01s/000  streams=4    uwsgi-242|idle|8  uwsgi-243|idle|8   <- auth now timing out
[t+58s] player_api_auth=15.01s/000  streams=0    uwsgi-242|idle|8  uwsgi-243|idle|8   <- ALL streams ended, pool STILL 8/8, auth STILL dead
# player_api only recovered after a container restart.
```

#### Guardian capture during an ORGANIC wedge (channel switch, 2 live streams) — confirms it's a hang, not CPU
Per-worker CPU jiffies (utime+stime) barely move across 15s = frozen, not a busy loop:
```text
[t+0s]  p242 st=S cpu=27402 wchan=0 | p243 st=S cpu=7249 wchan=0 | CLOSE_WAIT=0 conn:8080=0 load=2.09
[t+15s] p242 st=S cpu=27495 wchan=0 | p243 st=S cpu=7253 wchan=0 | CLOSE_WAIT=0 conn:8080=0 load=1.85
redis: dbsize=1851  timed SCAN(live:channel:*:clients:*)=2 keys in 23ms   (redis healthy)
pg_stat_activity (non-idle): none from the app (DB idle, no locks)
```

#### For context — a live-proxy stream start (provider host + credential segments censored)
```text
live_proxy.manager Starting stream for URL: http://censored.com/live/XXXX/XXXX/1568632.ts
live_proxy.server  Worker <pid> acquired ownership of channel <uuid> ... state connecting -> active
```

Relevant code (paths from the 0.27.2 image):
- `dispatcharr/db/backends/postgresql_psycopg3/pool.py` — `get()`/`put()`; `_connection_expired()` only
  checked there (no reaper); `get()` blocks on `self.pool.get()` when `size >= maxsize`.
- `dispatcharr/settings.py` — `DATABASES.default.OPTIONS.MAX_CONNS = 8`.
- `apps/output/views.py` — `xc_player_api` → `xc_get_info` → `get_user_active_connections` (+ `xc_get_user`
  auth) all require a pooled DB connection; `/api/core/version/` does not.

**Interim mitigations we are using** (neither is a fix): a healthcheck that probes `player_api.php` +
`willfarrell/autoheal` to auto-restart on the hang, and raising `MAX_CONNS` (delays onset ~3×, does not
stop the leak).

### Dispatcharr Version

0.27.2

### Docker Version

Docker version 29.5.3, build d1c06ef  (Docker Desktop on Windows, WSL2 backend)

Stock `ghcr.io/dispatcharr/dispatcharr:latest` (running 0.27.2). Comments trimmed; the healthcheck +
autoheal sidecar and a `uwsgi.ini` override (`workers=2`) are our wedge workarounds and are **not**
required to reproduce:
```yaml
services:
  dispatcharr:
    container_name: dispatcharr
    environment:
      - LOG_LEVEL=info
      - PUID=1000
      - PGID=1000
      - TZ=America/Chicago
      - DISPATCHARR_USE_X_FORWARDED_HOST=True
      - DISPATCHARR_USE_X_FORWARDED_PORT=True
      - DJANGO_USE_X_FORWARDED_HOST=True
      - DJANGO_SECURE_PROXY_SSL_HEADER=HTTP_X_FORWARDED_PROTO,https
    volumes:
      - ./config:/config
      - ./config/uwsgi.ini:/app/docker/uwsgi.ini:ro   # override: workers=2
      - dispatcharr_dispatcharr_data:/data
    ports:
      - "9191:9191"
    restart: unless-stopped
    healthcheck:                                        # workaround, not needed to repro
      test: ["CMD-SHELL", "curl -fsS --max-time 8 -o /dev/null http://127.0.0.1:9191/api/core/version/ && curl -fsS --max-time 8 -o /dev/null 'http://127.0.0.1:9191/player_api.php?username=REDACTED&password=REDACTED' || exit 1"]
      interval: 30s
      timeout: 20s
      retries: 4
      start_period: 180s
    labels:
      - autoheal=true
    mem_limit: 9000m
    mem_reservation: 7000m
    cpus: 8.0
    shm_size: 1024m
    gpus: all
  autoheal:                                             # workaround, not needed to repro
    image: willfarrell/autoheal:latest
    container_name: dispatcharr-autoheal
    restart: unless-stopped
    network_mode: none
    environment:
      - AUTOHEAL_CONTAINER_LABEL=autoheal
      - AUTOHEAL_INTERVAL=15
      - AUTOHEAL_START_PERIOD=180
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
volumes:
  dispatcharr_dispatcharr_data:
    external: true


### Client Information

The bug is **server-side and client-independent** — reproduced with plain `curl` against the proxy and XC
endpoints (no player involved). It surfaces to normal clients (any XC/M3U player, e.g. TiviMate / a set-top
XC app) as the guide/stream list failing to load and channels not starting for ~30–60s at a time, until the
container restarts.

### PR DESCRIPTION
**Pull request by Geesu**
## Description

Fixes a PostgreSQL connection-pool leak that periodically wedges the whole
DB-backed API under normal live-streaming use.

The custom psycopg3 gevent pool (`dispatcharr/db/backends/postgresql_psycopg3/pool.py`,
`MAX_CONNS` per uWSGI worker) only evaluates connection age inside `get()`/`put()`.
A connection checked out by a greenlet that dies without returning it — e.g. a
live-stream / channel-switch teardown path that never calls `close_old_connections()`
— is never aged out. It stays counted against `MAX_CONNS` but is no longer in the
free queue, so once `MAX_CONNS` connections leak per worker, `get()` blocks forever
and every DB-backed request on that worker hangs for ~60s and returns 504. DB-free
endpoints (e.g. `/api/core/version/`) stay fast, which is the tell. A container
restart is the only thing that clears it. Not CPU- or DB-bound — pure pool starvation.

This PR contains two changes:

1. **`pool.py` — background leak reaper.** A greenlet (one per worker, started
   lazily on first `get()` so the gevent hub is up) periodically closes connections
   that are both **expired** (older than `max_lifetime`) **and IDLE** (no in-flight
   transaction). An expired IDLE connection that isn't in the free queue is a leaked
   one; closing it discards it from the (weak) connection set and frees the slot.
   The IDLE guard means an active stream or in-flight request is never disrupted, and
   `get()` already refuses to hand out an expired connection, so a reaped connection
   cannot be in active use. This also makes `DATABASE_POOL_CONN_MAX_LIFETIME` actually
   reclaim leaked connections (previously it only recycled healthy idle ones); lower it
   to shorten recovery time. Reaper cadence is configurable via
   `DATABASE_POOL_REAP_INTERVAL` (default 30s; 0 disables).

2. **`docker/entrypoint.sh` — forward the pool env vars past the AIO allow-list.**
   In AIO mode the entrypoint only propagates an explicit whitelist of env vars to the
   app process (which runs as the `dispatch` user via `su`); anything not listed is
   dropped. `DATABASE_POOL_CONN_MAX_LIFETIME` was absent, so setting it had no effect
   in AIO — Django fell back to the 600s default even though the container env showed
   the override. Added `DATABASE_POOL_CONN_MAX_LIFETIME` and `DATABASE_POOL_REAP_INTERVAL`
   to the whitelist. (This is a pre-existing latent bug — the documented knob was a
   no-op in AIO regardless of this PR.)

The reaper is a defense-in-depth backstop that recovers the worker automatically
rather than the instant source-level release; it also covers the M3U-refresh INTRANS
leak family (#1338).

## Related Issue

Closes #1418

## How was it tested?

Deployed the patched AIO image to a live production instance (~15k streams, real
concurrent viewers, `DISPATCHARR_ENV=aio`, `MAX_CONNS=8`):

- Reproduced the original wedge per #1418 (pool census via
  `SELECT application_name, state, count(*) FROM pg_stat_activity WHERE
  application_name LIKE 'Dispatcharr-uwsgi%' GROUP BY 1,2;` climbing to 8/8 and
  staying pinned after all streams ended, with `player_api.php`/login hanging 60s
  while `/api/core/version/` stayed fast).
- After the fix: reaper starts per worker
  (`DB pool leak reaper started (interval=30s, max_lifetime=120s)`), the census no
  longer stays pinned at 8/8, and the login lockouts stopped.
- Verified the entrypoint change: with `DATABASE_POOL_CONN_MAX_LIFETIME=120` in
  compose, `printenv` showed 120 but the workers logged `max_lifetime=600s` before
  the whitelist change and `max_lifetime=120s` after.

No automated test is included — the behavior is timing/gevent-scheduler dependent and
was verified against the #1418 reproduction in a real deployment. Happy to add a unit
test around `_reap_leaked_connections()` (expired+IDLE closed, INTRANS/active skipped)
if preferred.

## Checklist

- [ ] I have read the [CONTRIBUTING.md](../blob/dev/CONTRIBUTING.md) in full
- [ ] I understand — line by line — every change in this PR and can explain it if asked
- [x] Backend: migrations are included if any models were changed — N/A, no model changes
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema — N/A, no new endpoints
- [x] Frontend: ESLint and Prettier pass cleanly — N/A, no frontend changes
- [ ] Tests are included for new functionality — see "How was it tested" (manual/production verification; no unit test)
- [ ] No `console.log`, `print()`, debug statements, or commented-out code is left in the diff
- [ ] I have not reformatted or refactored code outside the scope of this change

### CODE REVIEW

**Comment by SergeantPanda** [comment]
Please target the dev branch and and accept the CLA for this PR to be reviewable.

**Comment by Geesu** [comment]
> Please target the dev branch and and accept the CLA for this PR to be reviewable.

Done!

**Review by SergeantPanda — CHANGES_REQUESTED** [review]
Thanks for the thorough write-up and for checking this against a real #1418 wedge. The diagnosis looks right: abandoned checkouts stay counted against `MAX_CONNS`, never come back through `get()`/`put()`, and starve the worker. The entrypoint whitelist for `DATABASE_POOL_CONN_MAX_LIFETIME` is also a real AIO bug and worth landing on its own.

I'd rather not merge the reaper as written as the primary fix. A few concerns:

1. **IDLE is not the same as abandoned.** Under autocommit a still-owned checkout is IDLE between queries. The reaper predicate (expired + IDLE) can't tell that apart from a leak, so it can force-close a connection under a live greenlet. `get()` only checks expiry at checkout, so a hold that spans `max_lifetime` is in scope.

2. **Force-`conn.close()` interacts badly with geventpool cleanup.** django-db-geventpool does `if self.connection.closed: self.pool.close()` on the normal close path. So if the reaper kills a still-owned conn, a later `close_old_connections()` on that greenlet can wipe the whole worker pool, not just the leaked slot.

3. **Recovery is slow and incomplete.** Default reclaim is about `max_lifetime + reap_interval` (~630s). Greenlets already blocked in `pool.get()` are not woken when `_conns` shrinks; only new callers recover after the wedge.

4. **INTRANS leaks are not covered**, even with the #1338 mention. The reaper skips non-IDLE.

5. Calling `close_old_connections()` from a background greenlet also doesn't help. Connection state is greenlet-local, so it can't see other greenlets' checkouts.

**Suggested direction**

- Keep / split out the entrypoint env pass-through.
- Prefer source fixes: every stream/teardown/init greenlet that touches the ORM should `finally: close_old_connections()`. Issue comments also point at a boot-time `core_coresettings` checkout that burns a slot from process start. Worth chasing with `pg_stat_activity.last_query` while wedged.
- If we want a pool backstop later, it should track checked-out vs free, only reclaim clearly abandoned checkouts, and wake waiters (or add a timeout on `pool.get()` so the next leak fails fast instead of hanging).

Happy to review a follow-up that separates the entrypoint fix from a hardened backstop, or a PR that plugs the remaining release sites.

**Comment by nagelm** [comment]
Reviewed this while preparing to run it on my own deployment (same instance the live evidence in #1418's thread came from.). One edge case worth a look before merge, and a sketch for hardening it.

**The caveat: "expired + IDLE = leaked" isn't quite airtight under autocommit.** Django runs autocommit by default, so a connection that a *live*, long-running greenlet checked out and touches only occasionally sits at `TransactionStatus.IDLE` between queries — indistinguishable from an abandoned one. Timeline:

1. greenlet checks out a fresh connection at T+0 (passes `get()`'s expiry check);
2. greenlet legitimately holds it past `max_lifetime` (T+600s), idle between queries;
3. reaper closes it;
4. greenlet's next ORM call raises `InterfaceError: connection already closed`.

The docstring's "get() already refuses to hand out an expired connection, so an expired IDLE connection cannot be in active use" covers connections that went *through* `get()` after expiring — not ones that expired *while checked out* by a holder that never returns them. Ironically those holders are exactly the leak population this PR targets, and most of them are dead — but not all: plugin scheduler threads and other custom background threads are live holders with precisely this shape (I found one of myown plugins doing it; fixed by adding `close_old_connections()` after its ORM ticks). With the targeted cleanup commits in place, well-behaved core code no longer holds checkouts long, so in practice the assumption gets safer as those land — but third-party plugin threads remain exposed.

Reproduced the concern (psycopg 3.3.4, PostgreSQL 17, gevent), a live greenlet holding an autocommit checkout across max_lifetime=5s with DATABASE_POOL_REAP_INTERVAL=2:

12:25:02 django.geventpool: DB pool leak reaper started (interval=2s, max_lifetime=5s)
[expiring] first query OK; txn_status=IDLE; holding checkout for 10s while ALIVE...
12:25:08 django.geventpool: DB pool reaper closed 1 expired/leaked idle connection(s); size now 0/4
[expiring] RESULT: second query FAILED -> OperationalError: the connection is closed
=== control: identical flow, max_lifetime=3600s ===
[control] RESULT: second query OK -- connection survived

Same code, same conditions, only max_lifetime differs — so the failure is reaper-caused, and the holder was demonstrably alive (sleeping between queries), not a dead leaker.

**Hardening sketch — distinguish dead holders from live ones instead of guessing from idleness:** stamp the borrowing greenlet at checkout and let the reaper check whether it still exists:

```python
# in get(), after a connection is checked out / created:
import gevent, weakref
conn._dispatcharr_owner = weakref.ref(gevent.getcurrent())

# in _reap_leaked_connections(), in place of (or in addition to) the IDLE test:
owner = getattr(conn, "_dispatcharr_owner", None)
holder = owner() if owner is not None else None
if holder is not None and not holder.dead:
    # live holder: don't reap; optionally log — a connection held past
    # max_lifetime by a live greenlet is diagnostic gold for finding the
    # next missing close_old_connections() call site.
    continue
```

A dead/collected owner greenlet means the connection is truly abandoned — reapable with certainty (even without waiting for `max_lifetime` if you wanted to be aggressive). A live owner means "leave it, but log it": that log line would have pointed straight at every leak site #1418 documents, plus any future ones, without any risk of closing a connection out from under running code. (`put()` should clear the stamp so pooled-and-reused connections don't carry a stale owner.)

Perhaps a personal choice but I always think a new reaper should be tightly scoped (overly) and then later extended to additional reap targets once _"reaps everything it should and nothing it shouldn't"_ is measurable in the wild.  

Happy to test any revision against the same live reproduction setup from the #1418 thread — the organic wedge there is reproducible on demand.

Geesu:
Fixed in dev.

PR #1421, opened 2026-07-08
Add native xz decompression support for EPG sources and uploaded M3U playlists

### LINKED ISSUE #1414: [Feature]: support for xz-compressed EPG sources

Dispatcharr does not natively support xz-compressed EPG files.

There are more and more EPG sources that provide XZ-compressed files, reducing the size of the EPG downloads. 
Workaround are available BUT it would be ideal to support xz natively.

### Additional context

Nothing Extra

MotWakorb:
## Description

Adds native support for xz-compressed EPG sources, using Python's stdlib `lzma` module. No new dependencies, no Docker image changes (verified `import lzma` works in the published image's Python).

**What changed:**
- `apps/epg/tasks.py`: `detect_file_format()` recognizes xz by magic bytes (`FD 37 7A 58 5A 00`), `.xz`/`.xml.xz` extension, and MIME type, with the same precedence as the existing gzip/zip detection. `extract_compressed_file()` gains an xz branch mirroring the gzip branch exactly (4KB content sniff, then 64KB chunked streaming extraction, so large guides never load fully into memory). `fetch_xmltv()`'s local-file compressed-extension check now includes `.xz`; the URL-download path picks up xz transitively since it already routes through the detection/extraction helpers.
- `apps/epg/models.py`: `EPGSource.get_cache_file()` maps `application/x-xz` uploads and the xz magic header to the `.xz` extension. The magic-byte read widened from 4 to 6 bytes (the xz signature is 6 bytes). While doing so I converted the fixed-length slice comparisons to `header.startswith(...)`: the old `header[:5] == b'<tv>'` form silently depended on the read length matching the literal length, and would have broken bare-`<tv>` detection with the wider read. `startswith` is correct for any read length.
- `apps/m3u/tasks.py`: extends the same support to **uploaded M3U playlists**. `.xz` files stream line-by-line through `lzma.open(..., "rt")` exactly like `.gz` does today. This is not strictly asked for in #1414 (which is about EPG); I included it because the upload path shares the identical gzip/zip dispatch structure and would otherwise have the exact same gap for xz playlists. Happy to split it into a separate PR if you'd prefer.

**Notes for reviewers (known limitations, deliberate choices):**
- The `application/x-xz` MIME branches are largely unreachable via stdlib `mimetypes` (it classifies `.xz` as an *encoding*, returning `(None, 'xz')`). This mirrors the pre-existing `application/gzip` branches, which have the same characteristic. Kept for structural parity; tested via `mock.patch`.
- `lzma.LZMAError` was added to `fetch_m3u_lines`'s except tuple to mirror the existing `gzip.BadGzipFile` entry. For transparency: since the streaming refactor, that function returns paths without opening `.gz`/`.xz` files, so corruption in either format actually surfaces later during parsing. The entry is precedent-matching, not a live safety net.
- None of the extraction paths (pre-existing gzip/zip, or the new xz) cap total decompressed output size; only the per-read chunk is bounded. xz's higher compression ratios make a decompression-bomb cheaper to construct than with gzip, so I deliberately did **not** add a cap to just the xz branch (inconsistent protection is worse than none). If you'd like, I can follow up with a separate PR/issue proposing a uniform, configurable output ceiling across all three formats.
- The `extract_compressed_file()` xz branch is a deliberate near-verbatim mirror of the gzip branch rather than a shared helper, since refactoring your dispatch logic felt out of scope for an external feature PR. If you'd like a `_stream_decompress(opener, ...)` consolidation as a follow-up, glad to do it.
- The docs site (`Dispatcharr-Docs`, `m3u-epg-manager.md` in en/es/pt) currently says supported formats are ".gz or .zip" in six places; those will need a one-line update each once this merges. I can send a companion docs PR if useful.

**Suggested changelog entry** (leaving CHANGELOG.md untouched per the convention that maintainers add entries):

> **Added**: Native `.xz` decompression for EPG sources and uploaded M3U playlists. xz-compressed XMLTV and M3U files are detected by content-sniffing, file extension, or MIME type, with the same precedence as the existing `.gz`/`.zip` support. Uploaded `.xz` M3U playlists stream line-by-line rather than loading fully into memory, matching `.gz` behavior. (#1414)

## Related Issue

Closes #1414

## How was it tested?

- 16 new tests across `apps/epg/tests/test_xz_support.py`, `apps/epg/tests/test_fetch_xmltv_xz.py`, and `apps/m3u/tests/test_xz_playlist.py`: format detection (magic bytes / extension / compound `.xml.xz` / MIME priority), extraction round-trip integrity, corrupt-xz input failing closed (returns `None`, no partial output file, no exception into the Celery worker), `get_cache_file` extension inference including bare-`<tv>`/`<?xml` header regression tests, and end-to-end `fetch_xmltv()` / `fetch_m3u_lines()` runs against real xz files on disk.
- Full backend suites for both touched apps: `python manage.py test apps.epg apps.m3u` produced **392 tests, OK** (Postgres test DB via `dispatcharr.settings_test`).
- Real-data verification: decompressed a 1.53GB XMLTV guide (140MB `.gz` / 46MB `.xz` of identical content) through the exact new code path inside the Docker image's Python. Results: xz 2.57s vs gzip 2.07s (median of 3), with the xz file 67% smaller to download.

## Checklist

- [x] Backend: migrations are included if any models were changed (no model fields changed, no migration needed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **Native XZ decompression for EPG sources and uploaded M3U playlists.** `.xz`-compressed XMLTV and M3U files are now decompressed using Python's stdlib `lzma` module. EPG ingestion detects XZ via magic bytes, file extension (including compound names like `.xml.xz`), or MIME type; uploaded M3U `.xz` playlists stream line-by-line like `.gz`. The `/data/epgs` auto-import scanner now includes `.xz` files alongside `.xml`, `.gz`, and `.zip`. (Closes #1414) — Thanks [@MotWakorb](https://github.com/MotWakorb)
 - **XC catch-up (timeshift) support**. Adds a native `/timeshift/{user}/{pass}/{stream_id}/{timestamp}/{duration}.ts` endpoint that proxies replay sessions from an Xtream Codes provider to clients such as iPlayTV (Apple TV) and TiviMate. Auth reuses the same `xc_password` custom-property the live `/live/.../{id}.ts` endpoint already uses. Catch-up sessions appear on `/stats` with a violet `TIMESHIFT` badge alongside live sessions and respect per-channel access rules. (Closes #133) — Thanks [@cedric-marcoux](https://github.com/cedric-marcoux)
   - **Catch-up flags** (`tv_archive`, `tv_archive_duration`) denormalized at import time for zero-cost output queries; end-of-refresh SQL rollup updates channels linked to the refreshed account and self-heals stale flags on those channels only (e.g. after catch-up streams are removed or an account is deactivated).
   - **Strictly-UTC API surface, automatic per-provider timezone.** `server_info.timezone` is always `UTC` and the XC EPG `start`/`end` strings are emitted in UTC; the proxy converts the requested instant to the serving provider's own reported timezone (`server_info.timezone` captured on account refresh) at request time. No timezone to configure.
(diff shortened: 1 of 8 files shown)


SergeantPanda:
Thank you, looks good!

I added the ability for the auto importer to add xz as well.

PR #1425, opened 2026-07-09
fix(timeshift): serve the requested position when reusing a session (FF/RW snaps back to rewind anchor)

cedric-marcoux:
## Description

Fixes the catch-up FF/RW bug reported on Discord and diagnosed in #1242: after a rewind, any later fast-forward snapped playback back to the exact rewind point.

Clients (TiviMate at least) keep the `?session_id=` query when they rebuild the seek URL with a new `start` timestamp. `_stream_reused_session` replayed the **stored** `descriptor["provider_timestamp"]` (the position the session was created for) and only computed from the requested timestamp as a fallback — so every timestamp-jump FF/RW inside a session re-served the session's original anchor. Direct-to-provider has no session_id, which is why the reporter couldn't reproduce it there.

The reuse path now always recomputes the provider timestamp from the **requested** one (the provider zone is a property of the account — read from the default profile's `server_info`, same as the fresh path, instead of the reserved profile which may be a non-default without `server_info`) and moves the pool descriptor (`media_id` + `provider_timestamp`) to the position actually served, so fingerprint matching and same-channel displacement keep comparing against reality. Provider-slot continuity — the point of the session pool — is unchanged, and the reuse-path debug log now gets the real tz instead of `None`.

## Related Issue

No GitHub issue — the bug was reported on Discord and this fix was requested by @SergeantPanda in the #1242 thread (https://github.com/Dispatcharr/Dispatcharr/pull/1242#issuecomment-4919318961 has the full root-cause analysis).

## How was it tested?

- **Reproduced first, without any client**: `start=X` → follow the 301 → play → request a different `start=Y` with the same `session_id`. Before the fix the two responses were **byte-identical (same MD5)** and the pool entry kept X's `provider_timestamp`.
- **3 regression tests added** (`TimeshiftSessionReuseTests`): the reused session serves the requested timestamp, never the stored anchor (unit, tz-converted `17:30 UTC → 19:30 Europe/Brussels`); the pool descriptor follows the seek (unit); an end-to-end timestamp-jump through `timeshift_proxy` with the same `session_id` reaches the new position.
- **Full timeshift suite: 153/153** on a container built from this branch (`apps.timeshift.tests.test_helpers` + `test_views`).
- **Related suites green** on the same build: `apps.channels.tests.test_catchup_utils`, `test_get_stream_assignment`, `test_ts_proxy_teardown`, `apps.proxy.vod_proxy.tests.test_profile_connections` (63/63). The `apps.output` XC VOD listing tests fail only on my populated local instance (they read the live library through the AIO container's shared-DB test quirk) and are untouched by this diff.
- **Live verification against a real XC provider through Dispatcharr**: after the fix the same two-request repro returns different content and the descriptor moves (`provider_timestamp` `19-00 → 19-30`, `media_id` follows). Real-client check on TiviMate: rewind, play forward, FF now advances instead of snapping back.

## Checklist

### DIFF
--- a/apps/timeshift/tests/test_views.py
+++ b/apps/timeshift/tests/test_views.py
@@ -1607,6 +1607,132 @@ class TimeshiftSessionReuseTests(TestCase):
         # Reuse reserved once; failover reserved only for the other account.
         self.assertEqual(reserve_mock.call_count, 2)
 
+    def _call_reused_session(self, timestamp, media_id):
+        """Drive _stream_reused_session against a session anchored at 17-00
+        (provider position 19-00) with a NEW requested timestamp."""
+        profile = MagicMock(id=31, custom_properties={})
+        tz_profile = MagicMock(
+            custom_properties={"server_info": {"timezone": "Europe/Brussels"}}
+        )
+        account = MagicMock(id=1)
+        account.profiles.filter.return_value.first.return_value = tz_profile
+        ok = MagicMock(status_code=200)
+        with patch.object(views.M3UAccount.objects, "get", return_value=account), \
+             patch.object(views, "_attempt_timeshift_stream",
+                          return_value=ok) as attempt_mock:
+            response = views._stream_reused_session(
+                self.redis,
+                session_id=self.SESSION,
+                descriptor={
+                    "account_id": "1",
+                    "stream_id": "111",
+                    "provider_timestamp": "2026-06-08:19-00",
+                },
+                profile=profile,
+                channel=self.channel,
+                media_id=media_id,
+                safe_ts=timestamp.replace(":", "-"),
+                timestamp=timestamp,
+                duration_minutes=40,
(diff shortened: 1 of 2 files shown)


cedric-marcoux:
Pushed a self-review hardening commit (`35dce2b7`) after adversarially re-checking the first commit: on an actual position move the pool entry now also drops `content_length`/`serving_range` (they describe the previous position's file — keeping them could feed the near-EOF/displacement heuristics another programme's size), the descriptor update is lock-guarded and skips a vanished entry instead of resurrecting a partial TTL-less hash, and the reuse-path tz lookup now filters `is_active=True` like the fresh path. +3 tests → suite at 156/156.

SergeantPanda (approved):
This looks good to me. I am going to look into if we really need the extra database call or if we can just store provider_tz with the session.

Thank you!

Issue #1427, opened 2026-07-09
[Bug]: Incomplete implementation of Redis modular deployment causes unix sockets to be unsupported


Holbormon:
### Describe the bug

Since [Version 0.15.0](https://github.com/Dispatcharr/Dispatcharr/pull/767) it is supposedly possible to connect to a custom Redis instance using the variables `REDIS_HOST`, `REDIS_PORT`, `REDIS_DB`, and `REDIS_URL`. From the corresponding changelog entry:

>  Redis connection settings now fully configurable via environment variables (`REDIS_HOST`, `REDIS_PORT`, `REDIS_DB`, `REDIS_URL`), replacing hardcoded `localhost:6379` values throughout the codebase

However, it seems that `REDIS_URL` is not really used, except for a [consistency check involving SSL parameters](https://github.com/Dispatcharr/Dispatcharr/blob/a347e50e4806105e65def14795d8f7108a88f076/dispatcharr/settings.py#L441). There are two main problems with the current implementation:

[In the relevant initialisation code](https://github.com/Dispatcharr/Dispatcharr/blob/a347e50e4806105e65def14795d8f7108a88f076/dispatcharr/settings.py#L175), the main Django app constructs a Redis connection string for celery workers using `REDIS_HOST`, `REDIS_PORT`, `REDIS_DB`, which completely ignores `REDIS_URL` (this problem can be bypassed for now using CELERY_BROKER_URL instead of REDIS_URL). 

More critically, there are many other instances of a redis client being created (such as [here](https://github.com/Dispatcharr/Dispatcharr/blob/a347e50e4806105e65def14795d8f7108a88f076/core/views.py#L46), [here](https://github.com/Dispatcharr/Dispatcharr/blob/a347e50e4806105e65def14795d8f7108a88f076/scripts/wait_for_redis.py#L85) or [here](https://github.com/Dispatcharr/Dispatcharr/blob/a347e50e4806105e65def14795d8f7108a88f076/core/utils.py#L142)) only using the `REDIS_HOST`, `REDIS_PORT` and `REDIS_DB` variables, effectively discarding `REDIS_URL`.

This is particularly troublesome when trying to connect to a unix socket using a connection string such as `REDIS_URL=redis+socket:///var/run/redis/redis.sock?virtual_host=7`, since not setting `REDIS_HOST` and `REDIS_PORT` systematically results in the previously-hardcoded `localhost:6379` being used.

### How can we recreate this bug?

[...]

PR #1428, opened 2026-07-09
fix(redis) Give Redis clients a chance to use REDIS_URL if provided

Holbormon:
## Description

Minor changes to makes all redis clients use REDIS_URL to build a connection string if provided. Also extends to celery workers connection strings.

## Related Issue

Closes https://github.com/Dispatcharr/Dispatcharr/issues/1427

## How was it tested?

Work in progress.

## Checklist


Holbormon:
Probably needs adjustments as to which kwargs are passed to `from_url`. It's not obvious from the [docs](https://redis.readthedocs.io/en/stable/connections.html) which keywords are supported by `from_url`, and which ones can be encoded directly in the redis connection string (aside from the obvious username, password, and db). For instance, an error is raised when `socket_keepalive` is passed as a keyword argument, but the other kwargs seem fine. 

From the sentence:

> In the case of conflicting arguments, querystring arguments are used.

I'm guessing that all kwargs are actually supported by the querystring, meaning the optimal solution would probably be to `urlencode` all kwargs.

SergeantPanda (changes requested):
Thanks for digging into this. The gap in #1427 is real, and using `REDIS_URL` more consistently is the right fix. A few things I checked against redis-py 6.4.0 / Kombu on this branch that look like blockers though.

### 1. Unix sockets still fail with the current `from_url` kwargs

In `core/utils.py` (and the live_proxy fallback), the `from_url(...)` path still passes `socket_keepalive=...`, and can also pass `**REDIS_SSL_PARAMS` (including `ssl=True`).

That is fine for TCP `redis://`. With a `unix://` URL it blows up when the connection is actually created:

```
TypeError: AbstractConnection.__init__() got an unexpected keyword argument 'socket_keepalive'
```

`ssl=True` hits the same kind of error on unix connections. Since unix sockets are the main point of #1427, the URL path needs to drop TCP/TLS-only kwargs for `unix://`. You already flagged the kwargs concern in the thread; this is that issue showing up for real on unix.

### 2. One `REDIS_URL` cannot feed both redis-py and Celery for sockets

The issue example uses a Celery/Kombu URL:

`redis+socket:///var/run/redis/redis.sock?virtual_host=7`

redis-py only accepts `redis://`, `rediss://`, and `unix://`.
Kombu accepts `redis+socket://`, but not `unix://`.

This PR puts the same env value into Channels/django-redis and the Celery default broker URL, so whatever scheme someone picks will break the other stack.

Two workable approaches:
- Document that socket users set `CELERY_BROKER_URL` / `CELERY_RESULT_BACKEND` separately (Kombu form) and use `unix://` for `REDIS_URL`
- Or translate between schemes when building Celery vs redis-py clients

### 3. Gate on env `REDIS_URL`, not `settings.REDIS_URL`

`settings.REDIS_URL` is always set (env override or constructed default), so this:

```python
redis_url = getattr(settings, "REDIS_URL", "")
... if not redis_url else redis.Redis.from_url(...)
```

always takes the `from_url` branch. The host/port constructor becomes dead code.

`persistent_lock.py` already does the better thing with `os.environ.get("REDIS_URL", "")`. We should use one rule everywhere: only use `from_url` when the env var is actually set.

### 4. Startup wait script still ignores `REDIS_URL`

`scripts/wait_for_redis.py` is unchanged and still connects via `REDIS_HOST` / `REDIS_PORT`. If someone only sets `REDIS_URL`, startup can still probe localhost:6379 while the app expects the socket.

### 5. Celery SSL query append

When `REDIS_SSL=true` and env `REDIS_URL` is set, this uses that URL as `_default_redis_url` and may append `?ssl_cert_reqs=...`. If the URL already has a query string, you get a broken double-`?` URL. Safer to only append for constructed TCP URLs, or just tell people to put SSL params in the URL / override `CELERY_BROKER_URL`.

### 6. Checklist / tests

The checklist says tests are included and existing tests pass, but the diff has no tests, and “existing tests still pass” is still unchecked / WIP. Before merge it would help to cover at least:
- TCP `REDIS_URL` override works
- `unix://` works without keepalive/ssl kwargs
- Documented Celery behavior for sockets (or an explicit `CELERY_BROKER_URL` path)

---

Biggest improve IMO would be one helper (something like `create_redis_client(...)`) that:
1. Prefers env `REDIS_URL` when set
2. Strips TCP/TLS-only kwargs for `unix://`
3. Gets reused by `RedisClient`, proxy fallbacks, views, and `wait_for_redis.py`

Once the unix kwargs, scheme split, and wait script are covered, this should be in good shape.

Holbormon:
Sorry for the delay getting to this. Rebasing and getting to work...

Holbormon:
@SergeantPanda reopening this PR with a hopefully better baseline implementation, which should address most concerns (except for `2.`, for which the documentation route seems more feasible for now). In a nutshell:

1. All `REDIS_*` environment variables handling is now centralised under `core.utils.RedisClient`, which ensures all places where we need a client (except for Celery) use the same connection parameters, and ensures proper kwarg handling when the `unix://` scheme is used. This new implementation prioritises `REDIS_URL`, which if set assumes all parameters necessary for addressing and TLS were included in the connection string (ignores `REDIS_[SSL_*|HOST|USER|PASSWORD|DB]`). Otherwise, it falls back to the current implementation using those variables. 

2. I have written tests for the new `RedisClient` implementation under `test_redis_connection_strings.py`, but they are not as robust as I would like them to be due to the way `redis-py` serializes and returns connection parameters once the redis client has been created (seems to boil down to inconsistencies between how from_url and redis.Redis handle kwargs)

3. I have made adjustments to `wait_for_redis.py` and `proxy/proxy_vod/views.py` to ensure that they appropriately use the new `RedisClient`. This required some minor changes to the `test_wait_for_redis.py` test suite (such as mocking different code paths).

SergeantPanda (changes requested):
Thanks for the update, and for walking through what changed.

The `RedisClient` consolidation is a clear improvement, and pulling `wait_for_redis` / VOD onto it is exactly what we wanted. Docs for the Celery `redis+socket://` vs redis-py `unix://` mismatch is fine for now.

A few gaps between the writeup and the current code:

### Unix URL rebuild

You mentioned proper `unix://` kwarg handling. The keepalive skip looks right, but the URL rebuild still mangles the path:

```python
url = (f"{connstring.scheme}://{connstring.netloc}/{connstring.path}",)
```

`unix:///var/run/test.sock?db=7` becomes `unix:////var/run/test.sock`, so redis-py connects to `//var/run/test.sock` instead of `/var/run/test.sock`.

You also noted the tests aren’t as strong as you’d like because of redis-py kwargs serialization. Agreed on that being messy, but the unix expectations look like they’re asserting the rebuild bug rather than redis-py quirks:

- `path: "//var/run/test.sock"`
- `db: ['7']` from `parse_qs` lists

I’d drop the split/rebuild/`parse_qs` path and just do `redis.Redis.from_url(redis_url, **safe_kwargs)`, omitting TCP-only kwargs for `unix://`. Then assert `/var/run/...` and a scalar `db`.

### “If REDIS_URL is set… otherwise fall back”

That matches the intent, but:

```python
redis_url = os.environ.get("REDIS_URL", getattr(settings, 'REDIS_URL', ''))
```

still falls through to `settings.REDIS_URL`, which is always populated with a constructed default. So the host/port + `REDIS_SSL_PARAMS` path barely runs unless someone explicitly sets `REDIS_URL=''`. Please gate on the env var actually being set.

### Cache / Channels

Leaving Celery on the docs path is fine. Worth calling out that Django cache and Channels still build from `REDIS_HOST`/`PORT`/`DB` only, so a `REDIS_URL`-only unix setup won’t cover those yet. A short note in the PR/docs is enough if that stays out of scope here.

Happy to look again once the unix path rebuild is fixed and the tests assert the real socket path.

Holbormon:
>unix:///var/run/test.sock?db=7 becomes unix:////var/run/test.sock, so redis-py connects to //var/run/test.sock instead of /var/run/test.sock.
>
>You also noted the tests aren’t as strong as you’d like because of redis-py kwargs >serialization. Agreed on that being messy, but the unix expectations look like they’re >asserting the rebuild bug rather than redis-py quirks:

Good catch, and my bad for putting this on `redis-py` instead of realising it stemmed from a bad use of `urllib`. However, I have a slight concern with the suggested fix:
>I’d drop the split/rebuild/parse_qs path and just do redis.Redis.from_url(redis_url, **safe_kwargs), omitting TCP-only kwargs for unix://.

Being that, without parsing the user-provided `REDIS_URL`, there is no way to ensure users don't override some tcp-only kwargs or other defaults we might want to set (see my comment on `utils.py:163`). Namely, because of how `from_url` prioritises connection string queries over kwargs, someone could very well pass a connection string containing e.g. `&decode_responses=false`, which would throw off our careful distinction between regular and pubsub-optimised redis clients.

If you prefer increasing robustness on the parsing of `REDIS_URL` at the cost of some robustness against erroneous user inputs, I'll happily make the requested changes, but I still wanted to flag that this tradeoff exists. If this is indeed something undesirable, I think we could get away with fixing that rogue slash in the URL reconstruction:

```url = (f"{connstring.scheme}://{connstring.netloc}{connstring.path}",)```

and using `parse_qsl` over `parse_qs`, while carefully type-checking its output.

**Comment by SergeantPanda** [comment]
Fair point on query vs kwargs. `from_url` does let the query string win, so a raw `from_url(REDIS_URL, **safe_kwargs)` would let someone override `decode_responses` / timeouts. You're right to want guardrails there.

The slash fix does restore `/var/run/test.sock`, and `parse_qsl` avoids the list wrapping. I'd still rather not pass leftover query keys in as kwargs though: redis-py won't coerce them the same way (`db` stays `"7"` instead of `7`).

A middle ground that keeps your intent: strip the protected keys from the query, rebuild the URL, then `from_url(sanitized_url, **safe_kwargs)`. That keeps path/db/ssl/auth parsing in redis-py and still locks down decode/timeouts/keepalive. For unix, keep the `scheme://` form even when netloc is empty (`urlunsplit` turns it into `unix:/path`, which redis-py rejects).

Two other things:

**Gate on env `REDIS_URL`, not `settings.REDIS_URL`**

The settings default is always populated, so this:

```python
redis_url = os.environ.get("REDIS_URL", getattr(settings, 'REDIS_URL', ''))
```

almost always takes the `from_url` path even when the env var isn't set. The host/port + `REDIS_SSL_PARAMS` fallback barely runs. Please only use `from_url` when the env var is actually set.

**live_proxy pubsub fallback**

The pubsub fallback in `server.py` still creates a direct `redis.Redis(host=..., port=...)` client if `get_pubsub_client()` fails. That path would still ignore `REDIS_URL`. Worth routing through `RedisClient` too for consistency.

Happy to take another look once those are in.

Holbormon:
This set of commits should address everything discussed.

One caveat to be aware of: while ensuring `server.py` uses `RedisClient` too, I had to hand off the connection retrying logic to `RedisClient`. In doing so, we may end up using `time.sleep` between retries instead of `gevent.sleep`. I'll admit I'm not sure how significant this is for the event handler's performance. 

If it is, then we can probably add (yet another) kwarg for `RedisClient` to choose between `time` and `gevent` appropriately.

P.S: speaking of places that still use `redis.Redis`, a quick search revealed that `persistent_lock.py` does, but I can't find instances of `persistent_lock.py` being used in the main codebase, therefore I'm assuming it's legacy code that's sitting unused.

SergeantPanda (changes requested):
Thanks, this is a clear step forward. Env gating, the unix path fix, sanitized `from_url`, the live_proxy fallback, and the updated tests all look good.

A few things still blocking:

### Pubsub + TCP `REDIS_URL`

`get_pubsub_client()` sets `socket_timeout=None`, then `urlencode` puts `socket_timeout=None` into the query. redis-py rejects that (`Invalid value for 'socket_timeout'`).

So with a TCP `REDIS_URL`, the pubsub client fails and live_proxy falls back to `get_client()`, which has a real socket timeout and is the wrong shape for `pubsub.listen()`.

Safer approach: keep protected keys stripped from the query, rebuild only addressing/TLS bits into the URL, and pass timeouts as kwargs. For `unix://`, omit `socket_keepalive` only. `socket_timeout` / `socket_connect_timeout` are valid on unix and should still be applied.

### Unix clients currently lose timeouts

Because sockargs are skipped entirely for `unix://` and not passed as kwargs either, normal unix clients get no connect/read timeouts. Only keepalive needs to be omitted.

### Cache / Channels / Celery still ignore `REDIS_URL`

`settings.py` still builds those from `REDIS_HOST`/`PORT`/`DB`. A `REDIS_URL`-only unix setup still splits RedisClient onto the socket while cache, Channels, and Celery hit localhost. That leaves live UI updates and cache-backed paths broken, so #1427 isn't fully fixed yet.

On the `time.sleep` vs `gevent.sleep` note: not a concern here. Our uWSGI workers monkey-patch `time`, so `RedisClient`'s `time.sleep` is already cooperative in that process.

Happy to re-review once the items above are covered.

Holbormon:
The first commit should take care of the first point.

The second commit tries to address the second point by enforcing the use of `REDIS_URL` (if set) over `REDIS_[HOST|PORT|USER|PASSORD|DB|SSL*]` throughout all of `settings.py`, otherwise preserving old logic. Naive `unix://` -> `redis+socket://` scheme conversion was applied for the Celery connection string, as per a point flagged in earlier comments.

One possible regression to be aware of: Code (such as [`core/api_views.py:420`](https://github.com/Holbormon/Dispatcharr/blob/3b242beeec88c2658d4ade64479b29d564611583/core/api_views.py#L420)) which naively checks for TLS usage by asserting the non-emptiness of `REDIS_SSL_*` vars may now produce undefined results, as the `REDIS_URL` code path does not expose information about TLS usage until a Redis client gets built. Exposing such information would require parsing `REDIS_URL` before usage, which introduces additional overhead and failure points. Ideally, future work should fetch TLS information in a more robust way, such as by directly querying Redis clients connection properties.

SergeantPanda:
Thank you for continuing to work on this. Before I look into the recent changes, would you mind resolving the conflicts?

SergeantPanda (changes requested):
Thanks for the update, and for walking through the two commits.

I re-checked both points from your note:
- Socket args as kwargs / keepalive omitted only for `unix://` looks good. The earlier pubsub TCP crash and missing unix timeouts are fixed.
- Settings now preferring `REDIS_URL` (when set) for Channels, cache, and Celery is the right direction.

A few things still look blocking:

### BlockingConnectionPool regression after the `dev` merge

Current `dev` bounds `RedisClient` with `BlockingConnectionPool` (`REDIS_MAX_CONNECTIONS` / `REDIS_POOL_TIMEOUT`) and applies `REDIS_IDLE_TIMEOUT`. After merging `dev`, that path is gone again: `RedisClient` is back to unbounded `redis.Redis` / `from_url` clients.

django-redis still gets the pool settings, but the app’s own clients do not. Under gevent that can grow Redis FDs again. The `REDIS_URL` work needs to sit on top of the pooled `_make_client` / `_connection_pool_kwargs` design rather than replace it.

### Celery unix DB index

The naive `unix://` → `redis+socket://` swap is appreciated, but Kombu sets `db` from `virtual_host` and overwrites `db`. So:

- `unix:///path?db=7` → `redis+socket:///path?db=7` → Celery uses **db 0**
- `redis+socket:///path?virtual_host=7` → Celery uses **db 7**

Please map `db` → `virtual_host` when building the Celery URL.

### TLS status must follow `REDIS_URL`

The caveat you raised around `core/api_views.py` is real, and it needs to be fixed in this PR rather than deferred. If we support configuring Redis via `REDIS_URL`, the environment/UI TLS status has to reflect that path too.

Right now `redis_tls` still keys off `REDIS_SSL` / `REDIS_SSL_CERT` / `REDIS_SSL_KEY` only. So `REDIS_URL=rediss://...` can be using TLS while the UI reports it disabled. We should also restore mismatch validation so `REDIS_SSL` and the URL scheme cannot disagree silently.

Minimal bar:
- Derive or expose TLS enabled/verify/mTLS in a way that stays correct when `REDIS_URL` is set (parsing the URL at settings load is fine)
- Keep/restore `ImproperlyConfigured` when `REDIS_SSL` and `redis://` / `rediss://` conflict
- Make the environment endpoint report that derived state

Happy to take another look once those are fixed.

SergeantPanda:
Were you still interested in working with this PR?

Holbormon:
These commits respectively address the first and second points of the last review. 

As to the third point, I will admit that I'm not familiar enough with the entire codebase and libraries to come up with a satisfying resolution &mdash; especially given that the current implementation, which scans env vars to assess whether TLS is used, seems too brittle to be worth salvaging: ideally, TLS availability should be checked at the Redis client level, and the `REDIS_URL` connection string should not be parsed more than once. I would also argue that the flagged caveat does not really represent an urgent breaking change, as the code path involving `REDIS_URL` was already broken before this PR's work.

By this point, I would appreciate if one or several **humans** could chime in to discuss whether the work provided in this PR is worth merging in `dev` (despite the flagged caveat about erroneous TLS status reporting when `REDIS_URL` is set), or otherwise contribute a robust implementation of Redis TLS status checks.

I'm afraid I'm reaching the limits of the time I can benevolently put into this, and would like to stress that further delays in merging risk allowing the `dev` branch to further alter `RedisClient` with boilerplate code without consideration for this PR's goal; that is, taking a fundamentally flawed implementation of `REDIS_URL` to a usable state.

SergeantPanda (changes requested):
I want to address this part of your comment directly:

> I would appreciate if one or several **humans** could chime in...

I have been reading every round of this review and deciding what matters. I don’t have time to hand-audit every PR and write long replies myself, so I use tooling to help verify claims and reproduce failures. That doesn’t mean this is unsupervised or that nobody is accountable for the feedback.

On the “just merge it” ask: I don’t think that would have been responsible earlier, and I’m still not comfortable doing it with open correctness issues.

If we had merged the early versions, we would have fixed your unix/`REDIS_URL` case while breaking or weakening Redis behavior for the common path most installs use. Some concrete examples from this review:

- unix URL rebuild producing `//path` instead of `/path`
- pubsub + TCP `REDIS_URL` crashing on `socket_timeout=None`
- Celery landing on db 0 instead of the requested db
- unbounded Redis pools under gevent after the merge (and still on the `REDIS_URL` path now)

Those are exactly the kinds of “works for me, breaks everyone else” regressions review is supposed to catch.

I do appreciate the work you’ve put in. This PR is much closer than where it started. But “usable for `REDIS_URL`” still has to mean correct for the default deployment too, not only for the modular/unix setup.

Remaining blockers from my last check:

1. **`REDIS_URL` still bypasses `BlockingConnectionPool`.**  
   `Redis.from_url(url, connection_pool=BlockingConnectionPool(...))` does not use that pool. Host/port is bounded; `REDIS_URL` still gets an unbounded `ConnectionPool`. Please use `BlockingConnectionPool.from_url(...)` then `redis.Redis(connection_pool=pool)`, and add a test with `REDIS_URL` set.

2. **TLS status still needs to follow `REDIS_URL`.**  
   I hear you on time/scope, but I don’t want to merge this with that caveat. If we support `REDIS_URL`/`rediss://`, the environment/UI status has to reflect it. A small settings-time derivation from the URL scheme (plus mismatch checks) is enough.

Celery `db` → `virtual_host` looks good for the main case. Thank you for that.

If you want to stop here, that’s understandable. If you can land those two, I’m happy to re-review quickly.

Holbormon:
> I have been reading every round of this review and deciding what matters. I don’t have time to hand-audit every PR and write long replies myself, so I use tooling to help verify claims and reproduce failures. That doesn’t mean this is unsupervised or that nobody is accountable for the feedback.

Thank you for shedding light on your process. I can appreciate your reasons, and the confirmation that the review process is human-driven is just what I needed before allocating more time to this PR. 

While I don't want to assert that the tooling is counterproductive, I do want to flag that the AI-generated replies in particular present the risk of reallocating the time burden between writer and reader instead of actually saving time. Much like with voice DMs, the sender might earn time in producing a message, at the cost of the recipient spending extra time parsing it (especially for information that needs multiple time-spaced retrievals, such as code reviews).

> On the “just merge it” ask: I don’t think that would have been responsible earlier, and I’m still not comfortable doing it with open correctness issues.

Of course, and my goal was never to push for merging breaking changes or question the fixes demanded by past reviews (which I always agreed to, and did not really deserve re-listing in your reply). Rather, I was trying to argue that the last TLS status caveat in particular was not technically a regression (since it never worked with `REDIS_URL` in the first place), and that it was difficult to fix in a non ad-hoc way &mdash; especially if you wanted to properly support `REDIS_SSL_VERIFY` and `REDIS_SSL_CERT` too instead of just the scheme, which I now understand you do not). In the end you're the one deciding the direction the software takes, so I'll happily restore the proposed url scheme checks.

Apologies for the oversight on `BlockingConnectionPool`. I guess the inconsistency of kwargs between `from_url` and `redis.Redis` keeps proving tricky, and I should have caught that.

And finally, apologies for the tone of urgency from my last message. My motivation had been somewhat eroded by the soulless AI replies, as well as having to merge conflicts from `dev` (which I was worried were also AI-generated without much consideration for the Redis client centralisation work done in this PR).

SergeantPanda:
Yes I do agree with a lot of your sentiment on AI.

It unfortunately has become something we all are going to have to deal with going forward. If you take a look at all of the PR's I have to go through or the many issues I have to read through on a daily basis you will understand why I need AI to combat all of these AI bots submitting code/issues to the repo.

I wish AI wouldn't fluff up messages so much and add so many filler words but sometimes it is better at explaining things than I can write so I lean on it for that to ensure notes aren't misunderstood between everyone involved as that could lead to even longer time to merge PR's.

I understand what you're saying about redis_url not working before but I still see it as a 'regression' because *if* someone had it set or attempted to set it up, it would have caused them issues they wouldn't have if it was working correctly from the start. I'm not saying you caused it and sorry if it came off that way.

For the kwargs, this can be extremely complicated to get right with all of the different environments we have to worry about which is one of the reasons I rely on AI review, I don't have a great way to test this myself and do not want to cause issues for others to solve a problem for a few.

Thanks for sticking with me on this. 

PS: I did hand write all of this.

Holbormon:
(No urgency in reviewing, just marking the PR status as `needs review`, because Github thought your comment contained a request for changes)

SergeantPanda (approved):
Thanks for sticking with this. The connection work is in good shape: REDIS_URL is used when it is set, host/port/db still apply when it is not, and the pool stays bounded.

I finished the TLS status piece on this branch. The startup log and the connection security panel now follow the URL (scheme plus ssl_cert_reqs, ssl_certfile, and ssl_keyfile) instead of only the REDIS_SSL* variables.

Two bugs turned up on a last pass and are fixed here as well. The Celery rewrite of db= to virtual_host= was applied to every URL, which crashes Kombu when db= shows up in a password or query string. It now runs only for unix:// and only on the db query key. A blank REDIS_URL= was also treated as set, so an empty value in a compose file skipped host/port and failed. An empty value is now the same as unset.

Normal installs that only set REDIS_HOST, REDIS_PORT, REDIS_DB, and REDIS_SSL are unchanged aside from the startup waiter using the shared client timeouts.

I also resolved the merge conflict. 

It looks like we are good to merge, thank you again for your time on this!

PR #1434, opened 2026-07-12
fix(proxy): reap ghost channel sessions stuck in initializing

Geesu:
## Description

Fixes ghost channel sessions permanently stuck in `initializing` that block all playback on a channel until manually stopped.

**Symptom** (reproduced on my production instance for ~2 days on one channel): every request to `/proxy/ts/stream/<uuid>` returned HTTP 200 with a single 188-byte TS packet containing `Error: Connection stalled`, then closed. `/proxy/ts/status` showed a session with `state: "initializing"`, `url: ""`, `buffer_index: 0`, `client_count: 0`, and an uptime of ~47 hours. Both upstream source URLs were verified healthy, and other channels on the same M3U account streamed fine. `POST /proxy/ts/stop/<uuid>` cleared it and the channel immediately worked again.

**Root cause**: `ProxyServer.initialize_channel()` writes early metadata (`state=initializing`, `stream_id`) with no TTL before the URL is written and before the ownership lock is acquired. If initialization dies in that window (exception, or a worker race where the presumed owner had itself failed), the failure path only released the ownership lock and left the metadata behind. The ghost is then immortal because of three interlocking gaps:

1. New play requests treat `initializing` as "operational" and attach to it (`views.stream_ts` / `check_if_channel_exists()` only verify the owner **worker's heartbeat** — the worker is alive, only the init died), so clients wait, stall, and detach without ever re-initializing. Because no `StreamManager` exists, the normal connect-failure failover to backup streams never runs.
2. The pubsub event listener only processes `CLIENT_DISCONNECTED` when `am_i_owner()` is true; with no lock holder, every disconnect event is dropped and teardown never triggers.
3. `_check_orphaned_metadata()` (the 30s sweep) also only checks the owner worker heartbeat, never the channel ownership lock.

The M3U profile connection slot allocated for the failed init also leaks (the #947 failure mode).

**Fix** — the new primitive is `ProxyServer.is_stale_pre_active()`: a channel in `initializing`/`connecting` with no ownership lock, no local stream manager, and age past the init grace period can never progress and is treated as dead. Applied in four places:

- Play requests hitting a ghost tear it down (releasing the leaked profile slot) and reinitialize fresh, so the first viewer after a failed init revives the channel and normal stream failover applies; `check_if_channel_exists()` cleans ghosts the same way.
- Initialization failure paths now mark the channel `error` (terminal → cleaned and re-inited on next request) instead of leaving `initializing` behind, guarded so another worker's healthy channel is never clobbered.
- The 30s orphan sweep additionally reaps pre-active channels with no ownership lock and no clients after 2× the init grace period, even when the owning worker is still alive.
- Early init metadata is written with a 5-minute TTL as a backstop, so any missed failure path expires on its own instead of living forever (successful init replaces the TTL as before).

Safety notes: the lock-absence signal is reliable because both the owning worker's cleanup loop (`extend_ownership`) and the stream thread (`_still_owner`) re-acquire a lapsed lock within seconds; the age threshold debounces the brief pre-lock window during a genuine init. Worst case for a misfire is a fresh re-initialization of a struggling channel, not a dropped healthy stream.

## Related Issue

Related: #669 and #695 (this is the failed-init variant of those zero-client ghost reports — the one that survives worker-heartbeat checks), and #947 (this fixes its init-failure profile-connection leak variant; ghost cleanup releases the slot).

## How was it tested?

- Ran the backend suites in the CI base image (`ghcr.io/dispatcharr/dispatcharr:base-dev`, via `scripts/ci_bootstrap_backend.sh apps.proxy.live_proxy.tests apps.proxy.vod_proxy.tests`): all tests pass, including 14 new tests covering the staleness detector truth table (owner lock present/absent, local manager present, young vs. stale age, missing timestamps), the `error`-marking guards (no clobbering of another worker's channel or active states), the orphan-sweep reaper (reaps stale ghost with a live owner worker; keeps locked, young, or client-holding channels), and an end-to-end `stream_ts` test asserting a ghost `initializing` channel is stopped and re-initialized instead of attached to.
- Reproduced the failure mode in production (Docker/TS proxy behind a reverse proxy): a channel ghosted in `initializing` for ~47h serving only `Connection stalled` packets; confirmed manual `/proxy/ts/stop/<uuid>` plus fresh init — the exact recovery this PR automates — restores playback.

## Checklist


SergeantPanda:
Thank you for the PR. I’d rather fix the root cause than add detectors around it.

The real bug is writing `initializing` metadata before the ownership lock, then treating worker heartbeat as proof the channel is alive. I’d rather acquire the lock first (or only write pre-active state under the lock), clean up fully on failure, and maybe keep a TTL as a backstop.

If you want to rework it that way, great, otherwise I can take it later. For now I’d rather hold off on this approach.

SergeantPanda:
Thank you again for the submission. I decided to implement the initialization refactor to fix this bug.

Geesu:
> Thank you again for the submission. I decided to implement the initialization refactor to fix this bug.

Awesome - any chance you have a link to the commit? it in dev?

SergeantPanda:
Yes, it is in dev: 820959e3908ba166813a1d403660dd8bac75f016

PR #1438, opened 2026-07-13
feat(epg): add program icon fallback to channel icon logic if program…

Gtt1229:
## Description
Add EPG query parameter `program_icon_fallback=channel` to populate channel logos for programs that do not have a logo.

## How was it tested?

- Tested via Docker on Debian
- Custom data still loads while the channel logo is applied to a program if no program logo exists

Note: Frontend lint has existing failures unrelated to this change.

## Checklist

- [X] Backend: migrations are included if any models were changed
- [X] Backend: new API endpoints appear correctly in the OpenAPI schema

### DIFF
--- a/apps/output/epg.py
+++ b/apps/output/epg.py
@@ -1112,9 +1112,13 @@ def generate_epg(request, profile_name=None, user=None, *, xc_catchup_prev_days=
             prev_days = 0
     use_cached_logos = request.GET.get('cachedlogos', 'true').lower() != 'false'
     tvg_id_source = request.GET.get('tvg_id_source', 'channel_number').lower()
+    program_icon_fallback = request.GET.get('program_icon_fallback', '').strip().lower()
+    if program_icon_fallback != 'channel':
+        program_icon_fallback = ''
     cache_params = (
         f"{profile_name or 'all'}:{user.username if user else 'anonymous'}"
         f":d={num_days}:p={prev_days}:logos={use_cached_logos}:tvgid={tvg_id_source}"
+        f":pif={program_icon_fallback or 'none'}"
     )
     content_cache_key = f"epg_content:{cache_params}"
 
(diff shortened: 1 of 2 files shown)


SergeantPanda:
I'm curious, why would someone want the channel logo set for every program?

Gtt1229:
It's only if the program itself doesn't have a logo, so the program's XML includes the channel logo as its logo in this endpoint.

Some players don't populate a cover without program art, you might get program initials or just generic grey boxes like on Jellyfin.

SergeantPanda:
> It's only if the program itself doesn't have a logo, so the program's XML includes the channel logo as its logo in this endpoint.
> 
> Some players don't populate a cover without program art, you might get program initials or just generic grey boxes like on Jellyfin.

Right, but what benefit would duplicating the channel logo for the program give? It doesn't add any additional information.

Gtt1229:
Not all clients fall back to channel logo when program logo is empty, so you end up with blank programs.

SergeantPanda:
But I guess my question is, why would you want that? What benefit is there to having incorrect program poster over simply not having one?

Gtt1229:
It's simply unappealing to see something like this:

image

SergeantPanda:
Yeah I see what you mean. 

To me though, it feels more like something that Jellyfin should handle better not something Dispatcharr should add incorrect data to it's output to help resolve. 

I'm not going to say I'm completely opposed to it but I'm leaning towards not wanting the extra code and hidden parameter (not enough value to have it exposed in the frontend probably).

Gtt1229:
I definitely understand that view point. I do too  ultimately agree that it should be the client's problem for handling.

I think I'll move forward with an intermediate script to edit my EPG on my end then.

No stress!

SergeantPanda:
Thank you for the understanding! I bet you could make a plugin that could handle it well.

PR #1440, opened 2026-07-14
fix(timeshift): accept QUERY-style incoming catch-up requests (streaming/timeshift.php)

dillardblom:
## Description

Clients that construct XC catch-up requests in QUERY layout
(`/streaming/timeshift.php?username=...&password=...&stream=...&start=...&duration=...`)
— e.g. Open-TV / Fred TV — had no matching urlpattern. The request silently
fell through to the frontend's `<path:unused_path>` catch-all and got served
`index.html` (200 OK) instead of reaching the catch-up proxy: no error, no
log line, the client just fails to play.

The PATH-style layout (`timeshift/<user>/<pass>/<stream_id>/<ts>/<dur>`)
already worked. QUERY-style autodetection already existed for *outgoing*
requests to the upstream provider (`helpers.build_timeshift_url_format_a`
for QUERY, `_format_b` for PATH — both tried via
`build_timeshift_candidate_urls`), but was never mirrored to the *incoming*
route clients use to reach Dispatcharr itself.

Fix: split `timeshift_proxy` into a shared `_timeshift_proxy_impl` plus two
thin entry points — `timeshift_proxy` (existing PATH-style, unchanged) and
`timeshift_proxy_query` (new QUERY-style, extracts `username`, `password`,
`start`, `stream` from the querystring and delegates to the same
implementation). Registered a new `streaming/timeshift.php` urlpattern in
`dispatcharr/urls.py` ahead of the frontend catch-all.

## Related Issue

Originally reported against the predecessor plugin (now merged into core via
#1242): [cedric-marcoux/dispatcharr_timeshift#10](https://github.com/cedric-marcoux/dispatcharr_timeshift/issues/10).
No in-repo issue was filed first — this is a small, well-scoped bug fix
(single missing urlpattern + matching view), not a feature addition.

Closes #

## How was it tested?

- 4 new regression tests in `apps/timeshift/tests/test_views.py`: URL
  resolution for both incoming styles, querystring→param mapping, and a
  missing-param 400 case.
- Full `apps.timeshift` suite (284 tests) run against this branch: 283
  pass. The one pre-existing failure
  (`TimeshiftSessionRedirectTests.test_missing_session_id_serves_busy_pool_on_channel_hop_without_redirect`)
  reproduces identically on an unpatched `:dev` baseline — unrelated to this
  change.
- Verified live against a real Open-TV client on my own test instance:
  before this fix, a QUERY-style request returned `200` with a body exactly
  matching `frontend/dist/index.html`'s byte size (858 bytes) — the SPA
  shell. After this fix, the same request returns `301` (session-mint
  redirect), identical to the PATH-style first-play flow, confirming the
  request now reaches the proxy.

## Checklist

- [x] Backend: migrations are included if any models were changed *(no models were changed)*
- [ ] Backend: new API endpoints appear correctly in the OpenAPI schema *(N/A — this is a plain Django view for an XC-client-facing URL, not a DRF/OpenAPI-documented endpoint, consistent with the existing PATH-style `timeshift_proxy`)*
- [ ] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) *(N/A — no frontend changes)*


dillardblom:
Adjusted

SergeantPanda:
This looks good. Thank you!

I cleaned up the comments a bit but otherwise didn't change anything. Please let me know if it is broke.

Issue #1261, opened 2026-05-22
[Bug]: Errors after upgrade


lovefilm2018:
### Describe the bug

I am getting constant errors after upgrading, see examples in screenshot. doesn't seem to be causing any actual issues that i can see yet apart from moving around the UI sometimes needs a few attempts (eg navigating to VOD) before it actually refreshes and displays data.

<img width="400" height="503" alt="Image" src="https://github.com/user-attachments/assets/14215ef8-d483-4c29-a24a-bc732458c828" />
<img width="380" height="335" alt="Image" src="https://github.com/user-attachments/assets/e0503964-b409-4bc0-ba93-0800dae14cc9" />

### How can we recreate this bug?

<img width="400" height="503" alt="Image" src="https://github.com/user-attachments/assets/77a2def6-2a8b-427b-8a5b-ee140cfe894c" />

<img width="380" height="335" alt="Image" src="https://github.com/user-attachments/assets/2477bda0-7da3-4c14-a805-6c02ad70e350" />

1. Go to '...' 2. Click on '....' 3. Scroll down to '....' 4. See error

### Dispatcharr Version

0.25.0

### Docker Version

not sure - seedbox provided

### What's in your Docker Compose file?

not using

### Client Information

Device, App, Versions for both, etc...

lovefilm2018:
Update, seems ok after reboot.

SergeantPanda:
Thank you for the update.

lovefilm2018:
Actually errors back, appears to be happening whilst something is recording.

lovefilm2018:
@SergeantPanda please can you reopen this issue

SergeantPanda:
This could be a number of things. We need more details than just a screenshot of the frontend error posted multiple times. Please provide logs and more details as well as provide a more accurate name.

Have you tried refreshing the web ui? Is there anything in the docker logs? If not, enable debug logging.

lovefilm2018:
Hi, I'm happy to provide logs but I don't have access to the docker container as it's locked by the seedbox provider. I can ask them to extract some logs? I have refreshed the UI.

SergeantPanda:
Yeah I wouldn't know how to help you get the logs but we'd definitely need them to see what exactly is going on. Does everything stop working when this happens until a restart?

lovefilm2018:
[...]

PR #1441, opened 2026-07-14
fix(ui): stop rendering raw HTML error pages in toast notifications (#1261, item 1)

nagelm:
Fix for #1261, item 1

Failed API responses whose body is not JSON (e.g. Django's HTML "Server Error (500)" page, nginx's HTML 502/504 pages when the backend is down or timing out) were interpolated verbatim into the error toast, showing users a wall of raw HTML markup.

`request()` deliberately keeps the raw text when `JSON.parse` fails, and `errorNotification()` rendered `${status} - ${body}` unconditionally. This PR routes the failed response through a new `formatApiError()` helper (in `utils.js`, next to the other shared helpers `api.js` already imports):

- JSON object bodies keep today's pretty-printed formatting
- markup and empty bodies collapse to the response's own status line (`502 - Bad Gateway`): the declared `Content-Type` decides what counts as markup (body sniffing only as a fallback when the header is missing), and the label is the fetch `Response.statusText` — the protocol's reason phrase, defined for every status code, rather than any hardcoded status map. HTTP/2+ transmits no reason phrase, so an empty `statusText` falls back to a generic label. `request()` already attaches the `Response` to the error, so no call-site changes.
- the full suppressed body is logged via `console.debug` so the markup stays available for troubleshooting — deliberate, not a leftover debug statement
- plain-text bodies are truncated at 200 chars as a backstop
- errors without a status keep the `error.message` fallback

## How was it tested?

- `npx vitest run src/__tests__/formatApiError.test.js` — 12 new cases: Django HTML 500 body, nginx HTML gateway pages, content-type-vs-sniff precedence, missing reason phrase (HTTP/2), JSON object body, empty body, long-plain-text truncation, non-standard status codes, and no-status errors.
- Full frontend suite: 5746 passed. The only 3 failures are pre-existing machine-timezone-sensitive date tests that fail identically on a clean `dev` checkout (they assert US-format date strings).
- ESLint/Prettier: no new findings on the touched files (the change removes one pre-existing unused-var; the remaining api.js/utils.js findings exist on clean `dev` and were left alone to keep the diff scoped).
- Live before/after: built the AIO image from this branch and ran it next to stock 0.27.2; killed uwsgi in each container so nginx answers every API call with its HTML 502 page, then opened the Stats page (which polls). 
Stock build: toasts filled with raw `<h1>502 Bad Gateway</h1>… <!-- a padding to disable MSIE… -->` markup. 
This branch: `Failed to retrieve connection stats: 502 - Bad gateway - backend unavailable`

## Checklist

- [ ] Backend: migrations are included if any models were changed (N/A — frontend only)
- [ ] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — frontend only)
- [x] Frontend: ESLint and Prettier pass cleanly on the changed code
- [x] No `console.log`, `print()`, debug statements, or commented-out code is left in the diff (the `console.debug` of the full error body is a deliberate part of the fix design, as described above)
tted or refactored code outside the scope of this change


SergeantPanda:
Hey @nagelm, thanks for the PR. The HTML-in-toast fix was solid and the helper/tests were a good foundation.

I pushed a few follow-ups onto the branch:

- JSON error bodies now prefer `detail` / `error` / `non_field_errors`, then field validation messages, instead of dumping pretty-printed JSON into the toast. That matches what our API actually returns.
- Markup sniffing only runs when `Content-Type` is missing, so a plain-text message that happens to start with `<` still shows.
- Added a small status label map for when `statusText` is empty (common on HTTP/2), so you get `502 - Bad Gateway` instead of a generic fallback.
- Cleaned up the comments a bit.

Also worth noting: this improves how those failures look in the UI, but it doesn’t fix the underlying request failures from #1261 so I'm not going to link the issue for closure.

PR #1442, opened 2026-07-14
test(frontend): make date assertions machine-independent

nagelm:
## Description

Three frontend tests encode the machine's timezone or locale into their expectations, so the suite fails on any box east of UTC (found on a UTC+12/13 machine; CI's UTC runner and US-locale dev machines never see it):

- **`dateTimeUtils` › `isSame`** compared `10:00Z`/`11:00Z` on the 15th as "same day", but `isSame` compares *local* calendar days, and no UTC instant falls on the same date in every timezone — at UTC+13 those two times are 23:00 on the 15th and 00:00 on the 16th. Fixed with timezone-less datetimes, which denote the same calendar day everywhere; that is the domain in which "same day" is actually well-defined.
- **`RecordingUtils` › `toDateString`** formatted a UTC-noon instant and expected the UTC date back, but `toDateString` renders local time (UTC+12 renders the next day). Fixed by constructing the `Date` from local components — matching the sibling test on the next line, which already does this correctly.
- **`SeriesModalUtils` › `getEpisodeAirdate`** — the production code renders the viewer's locale via `toLocaleDateString()` by design, but the test hardcoded the en-US shape (it tolerated the timezone day-shift with `/1[4|5]/` yet not the day/month order). Fixed by computing the expectation with the same API: for a function whose contract is "delegate to the viewer's locale" there is no locale-independent literal to assert, so the test now verifies the wiring (reads `air_date`; the N/A fallbacks are covered by the sibling cases) rather than one locale's rendering.

**Why not pin `TZ`/locale in the Vitest config instead?** Pinning would make the suite deterministic by only ever exercising UTC/en-US — hiding exactly the class of environment-sensitivity bug these utilities can have — and locale pinning in Node/ICU is unreliable across platforms. Expressing each fixture in its function's actual domain makes the suite pass in *every* environment, which is strictly stronger and keeps non-UTC machines as free regression coverage.

No production code changes; 6 lines changed across the three test files.

## Related Issue

None — small/obvious test fix.

## How was it tested?

- Full frontend suite on a UTC+12 / en-NZ machine: **5740 passed, 0 failed** — previously these three files failed on every run there.
- The fixes are timezone- and locale-agnostic by construction (local datetimes are the same calendar day in every zone; the locale expectation is computed with the same API the code uses), so they pass identically on UTC CI and US-locale machines.

## Checklist

- [ ] Backend: migrations are included if any models were changed (N/A — test-only frontend change)
- [ ] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — test-only frontend change)
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) on the changed lines; `SeriesModalUtils.test.js` carries pre-existing formatting drift upstream, and reformatting it wholesale would violate the no-out-of-scope-reformat rule
- [x] Tests are included for new functionality (this PR is tests)


SergeantPanda:
This looks good. Thank you!

PR #1445, opened 2026-07-14
fix(db): pin session timezone to UTC via libpq startup packet (#651)

### LINKED ISSUE #651: [Bug]: EPG timezone incorrectly converted to UTC (+0000) instead of CET (+0100)
### Describe the bug

- EPG source displays correct time (19h30 CET)
- Dispatcharr outputs EPG with +0000 offset without adjusting time
- Result: All programs show 1 hour late
- Version: 0.11.2
- TZ=Europe/Zurich configured
- Example: <programme start="20251114193000 +0000"> should be +0100

How can we recreate this bug?

1. Configure Dispatcharr with timezone Europe/Zurich (TZ=Europe/Zurich in Docker)
2. Add an IPTV provider (M3U + EPG source) where EPG contains correct timestamps 
   (e.g., program at 19:30 CET displays correctly in source)
3. Go to 'Settings > M3U & EPG Manager' and import the provider
4. Wait for EPG to be parsed and generated by Dispatcharr
5. Go to 'Settings > Channels' and verify a channel has EPG assigned
6. Download/inspect the generated EPG file from Dispatcharr output endpoint:
   http://YOUR_IP:9191/output/epg
7. Open the XML file and examine any <programme> tag
8. Observe that timestamps have timezone offset '+0000' (UTC) instead of '+0100' (CET)
   Example: <programme start="20251114193000 +0000"> 
   (Should be: <programme start="20251114193000 +0100">)
9. Load this EPG in any IPTV client (Snappier IPTV, GSE IPTV, etc.)
10. All programs display 1 hour later than they should
    (19:30 CET program displays at 20:30)

Expected behavior:
- Dispatcharr should preserve the original timezone offset from source EPG
- OR convert timestamps correctly: if converting to UTC, adjust the time 
  (19:30 +0100 should become 18:30 +0000, not 19:30 +0000)

Actual behavior:
- Dispatcharr outputs all timestamps with +0000 offset without adjusting the time
- This causes a 1-hour shift in all IPTV clients that correctly parse XMLTV timezone offsets

### Dispatcharr Version

11.2

### Docker Version

28.5.2, build ecc6942

services:
  dispatcharr:
    container_name: dispatcharr
    ports:
      - 9191:9191
    volumes:
      - /home/unername/docker/dispatch/dispatcharr_data:/data
      - /home/unername/docker/dispatch/dispatcharr_config:/config
      - /etc/timezone:/etc/timezone:ro
      - /etc/localtime:/etc/localtime:ro
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=redis
      - CELERY_BROKER_URL=redis://redis:6379/0
      - CELERY_TIMEZONE=Europe/Zurich
      - CELERY_ENABLE_UTC=false
      - DISPATCHARR_LOG_LEVEL=info
      - TZ=Europe/Zurich
    depends_on:
      redis:
        condition: service_healthy
    devices:
      - /dev/dri:/dev/dri
    restart: unless-stopped
    healthcheck:
      test: ["CMD", "curl", "-f", "http://localhost:9191/health"]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 40s

  redis:
    image: redis:7-alpine
    container_name: dispatcharr_redis
    restart: unless-stopped
    volumes:
      - /home/unername/docker/dispatch/redis_data:/data
      - /etc/timezone:/etc/timezone:ro
      - /etc/localtime:/etc/localtime:ro
    environment:
      - TZ=Europe/Zurich
    healthcheck:
      test: ["CMD", "redis-cli", "ping"]
      interval: 5s
      timeout: 3s
      retries: 3
      start_period: 5s

### Client Information

**Primary Device:**
- Device: Apple TV 4K (2021 or later)
- OS: tvOS 18.x (or current version)
- App: Snappier IPTV
- App Version: 5.0.x (latest from App Store as of November 2025)

nagelm:
## Description

Root cause and fix for the psycopg3-era regression of the EPG timezone bug. The full analysis is in my comments on #651 (2026-07-13); short version — DB sessions were never actually pinned to UTC, three mechanisms deep:

1. The `connection_created` → `SET TIME ZONE 'UTC0'` receiver in `core/apps.py` never fired on any version since it landed: it is a nested closure connected with Django's default *weak* reference, so it was garbage-collected the moment `ready()` returned (verified on a live 0.27.2: the signal's receiver list holds a dead weakref).
2. Sessions were UTC on older stacks anyway because Django's own `init_connection_state` configured the timezone. Since native psycopg-pool support (Django 5.1+) that path is gated on `not self.pool` — and the geventpool mixin's `pool` property is always truthy, so Django silently skips all connection configuration for this backend. This is what actually regressed at the psycopg2→3 / Django upgrade: the masking layer disappeared, not the (already dead) signal.
3. `'UTC0'` is additionally unresolvable in psycopg3's zoneinfo lookup (the "unknown PostgreSQL timezone" warning), so it buys nothing on the new stack even where it executes.

Net effect: every app session runs at the PostgreSQL **server default** timezone. On deployments whose server default is non-UTC (e.g. `/etc/localtime` bind-mounts at initdb time — the original report), psycopg3 decodes `timestamptz` reads under that session timezone and the XMLTV output labels local wall-clock times as `+0000` — the shifted-guide symptom in this issue.

**apps.py vs base.py**

The fix pins the GUC at the one layer that provably covers this backend: `TimeZone=UTC` in the libpq startup packet, added in the pool backend's `get_connection_params()`. The dead signal is removed rather than repaired — deliberately. Why the backend and not a fixed-up `connection_created` receiver back in `core/apps.py`:

- **Coverage by construction.** `get_connection_params()` is the single funnel every connection this backend creates passes through — the gevent pool calls `psycopg.connect(**kwargs)` with exactly these params, for uwsgi workers, Celery, and management commands alike. A signal only fires if app-registry lifecycle, registration order, and receiver liveness all cooperate — the exact failure class that broke the original (a weakref nobody could see was dead). None of those moving parts exist in the conninfo path.
- **Session-default semantics.** A post-connect `SET TIME ZONE` (signal-based or Django's own init) is *not* the session default: `RESET TimeZone` — or anything that resets session state — falls back to the **server** default and silently reintroduces the bug. A startup-packet GUC *is* what `RESET` returns to; the second regression test asserts precisely this property.
- **No unconfigured window.** `connection_created` fires after the connection already exists, so early statements can run at the wrong timezone; with the startup packet the very first statement sees UTC.
- **Locality.** "Sessions run in UTC" is a property of the database backend, and this backend already pins conninfo the same way — `application_name` in this very method, `client_encoding` in the pool kwargs. Timezone configuration living in an unrelated app's `ready()` was action-at-a-distance, and is how the dead receiver went unnoticed.

Also worth noting: there is no Django-provided hook left to use here — Django's own `_configure_connection` is exactly the right place, but as described above it is skipped for this backend by the `not self.pool` gate. Startup-packet GUCs additionally survive `ROLLBACK`, and `'UTC'` (unlike `'UTC0'`) resolves in psycopg3's zoneinfo lookup.

Notes for reviewer:

- **Stored data was never corrupted.** `timestamptz` storage is absolute; the corruption is read-side (decode under the leaked session timezone). Therefore, no repair or migration should be needed on existing deployments.
- The regression tests close **and unregister** their probe pool in teardown (`_connection_pools.pop`) — otherwise the pooled session outlives the test run and blocks dropping the test database.
- The larger structural alternative is fixing the geventpool/Django-5.1 gate collision itself (`init_connection_state` skips `_configure_connection` — timezone, role configuration, and Django's tz loader registration — whenever `self.pool` is truthy). _Happy to discuss that shape; this PR is the minimal targeted fix and remains correct under that future too._

## Related Issue

Closes #651

## How was it tested?

- **Stock-vs-patched A/B** on the current `dev` image, both containers with the PostgreSQL server default deliberately set non-UTC (`ALTER SYSTEM SET timezone='Europe/Zurich'` — the same effective trigger as the `/etc/localtime` mounts in the original report):
  - Stock: app session `pg_settings` shows `('Europe/Zurich', 'configuration file')`; a `12:00Z` datetime round-trips through the pool as `14:00` still labeled `+00:00`; an ingested XMLTV programme with `start="20260715120000 +0000"` renders in `/output/epg` as `start="20260715140000 +0000"` — the exact symptom reported here.
  - Patched: session shows `('UTC', 'client', 'UTC')`, the round-trip is exact, and the output preserves `20260715120000 +0000` — including after `RESET TimeZone`.
- **The new regression tests fail on stock** (`AssertionError: 'Europe/Zurich' != 'UTC'`) **and pass with the fix.** They exercise the real pool backend explicitly because the test runner's default engine is the vanilla Django backend.
- **Dead-signal verification** on a live 0.27.2 install: `connection_created.receivers` holds a dead weakref for `force_db_utc0`, i.e. the `SET TIME ZONE 'UTC0'` never executed.
- **Burn-in**: running on a production deployment since 2026-07-14; sessions report `('UTC', 'client', 'UTC')`.
- Core backend test suite passes with the fix applied (no new failures vs a clean dev run).

## Checklist

- [ ] Backend: migrations are included if any models were changed (N/A — no model changes)
- [ ] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — no API changes)
- [ ] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) (N/A — backend only)


SergeantPanda:
This looked good. Thank you!

PR #1446, opened 2026-07-15
fix(logging): render log timestamps in the system display timezone

### LINKED ISSUE #1439: [Bug]: Container TZ env is silently ignored — app logs stay UTC (nginx honors it: one container, two timezones) and XC player_api reports UTC to clients
**Summary.** Setting the standard `TZ` container env var has no effect on the application. All Django/uwsgi/celery log lines stamp UTC, while nginx's access log — in the same container — honors `TZ`, so one container logs in two timezones. The Xtream Codes API likewise reports `server_info.timezone: "UTC"` and a UTC `time_now` to clients regardless of `TZ` **and** regardless of the UI's System → Time Zone setting, which misleads XC clients that trust server time. (@dillardblom's comment in #651 — UTC logs despite `TZ`, XC API returning UTC, timeshift plugin breakage — is this bug, reported on 0.18.1; #651's own headline bug is the separate DB-session issue.)

**Root cause — two coupled halves (traced in the 0.27.2 image):**

1. `docker/entrypoint.sh` launches the app tree via `su - dispatch`, which strips env vars; survivors are re-exported through the explicit whitelist (`variables=(...)` array, ~line 177). `TZ` is not in it, so even the non-Django uwsgi master logs UTC. Same latent-bug family as `DATABASE_POOL_CONN_MAX_LIFETIME` being silently dropped (noted in PR #1420).
2. Whitelisting alone cannot fix Django-managed processes: `dispatcharr/settings.py` hardcodes `TIME_ZONE = "UTC"`, and Django's settings loader executes `os.environ["TZ"] = TIME_ZONE; time.tzset()` at import — re-stamping UTC over whatever the environment said, in every uwsgi worker, celery worker, and daphne. nginx never imports Django settings and starts outside the `su -`, hence the split-brain container.

**Expected behavior.** Either the standard `TZ` env is honored for display surfaces (log timestamps, XC server time), or the docs state it is unsupported. Ideally the UI's System → Time Zone setting (`CoreSettings.get_system_time_zone()`) drives these surfaces — it already canonically drives DVR recurring rules, celery crontab schedules, and backup naming.

**Suggested fix shape** — deliberately *not* proposing to make global `TIME_ZONE` env-driven: the `TIME_ZONE="UTC"` / `USE_TZ=True` invariant (and the per-connection UTC forcing from #651) clearly treat DB truth as UTC, and that should stay. Instead:

1. **entrypoint:** add `TZ` to the env whitelist (one line; fixes shell tooling, non-Django processes, and the nginx inconsistency).
2. **bootstrap env:** a `DISPATCHARR_TIME_ZONE` var following the existing `DISPATCHARR_LOG_LEVEL` pattern, normalized in the entrypoint from bare `TZ` when unset.
3. **log display:** give the `LOGGING` formatter a lazy converter that resolves `CoreSettings.get_system_time_zone()` at emit time (sidesteps the settings-import/DB chicken-and-egg), falling back to the env var before the DB is up.
4. **XC server time:** `xc_get_info` (`apps/output/views.py` ~398–411) should report `CoreSettings.get_system_time_zone()` with a matching `time_now`, instead of `tzlocal.get_localzone()` / `datetime.now()` — aligning it with every other user-facing timezone surface.

Happy to open a PR along these lines if the maintainers agree with the shape.

Operator workaround meanwhile (survives restarts, reverts on image update): `sed -i 's|^TIME_ZONE = "UTC"|TIME_ZONE = "<Your/Zone>"|' /app/dispatcharr/settings.py` and append `export TZ="<Your/Zone>"` to `/etc/profile.d/dispatcharr.sh`, then restart. (`USE_TZ=True` keeps DB storage UTC; the one behavior change is that *naive* datetimes POSTed to the recordings API are then interpreted in the new zone.)

1. Run the AIO image with any non-UTC zone: `-e TZ=Pacific/Auckland`
2. `docker exec dispatcharr date` → correct local time (env + tzdata are fine)
3. `docker logs dispatcharr` → Django/uwsgi lines stamp UTC
4. `docker exec dispatcharr cat /var/log/nginx/access.log` → nginx lines stamp the local offset (+1200) — same container, two timezones
5. `docker exec dispatcharr sh -c 'cat /proc/$(pgrep -o uwsgi)/environ | tr "\0" "\n" | grep ^TZ='` → empty (whitelist dropped it)
6. `curl 'http://HOST:9191/player_api.php?username=U&password=P'` → `server_info.timezone` = `"UTC"`, `time_now` = UTC wall clock — regardless of `TZ` and of the UI Time Zone setting

### Dispatcharr Version

0.27.2 (AIO, ghcr.io latest)

### Docker Version

29.5.2, build 79eb04c

n/a — Unraid dockerMan template (single container, AIO). Relevant env: `TZ=Pacific/Auckland`. Reproduces with any non-UTC `TZ` on a stock `docker run`.

### Client Information

TiviMate 5.3.3 on TCL Google TV (Android 11) — affected via XC `server_info` timezone/time_now. Independent confirmation: @dillardblom in #651 (0.18.1, Jellyfin/timeshift plugin misled by UTC server time)

nagelm:
## Description

Application log timestamps are pinned to UTC no matter what the operator configures, while nginx in the same container honors `TZ` — one container, two timezones. Two mechanisms force this: (1) `docker/entrypoint.sh` launches the app tree via `su - dispatch`, which strips the environment down to an explicit whitelist that omits `TZ`; and (2) whitelisting alone cannot fix Django-managed processes anyway, because Django's settings loader unconditionally re-stamps `os.environ["TZ"] = TIME_ZONE` (`"UTC"`) and calls `time.tzset()` in every uwsgi, celery, and management-command process — stdlib logging's `%(asctime)s` renders via `time.localtime()`, so it can never show local time in a Django process regardless of the environment.

The fix keeps `TIME_ZONE = "UTC"` / `USE_TZ` untouched: database truth and every wire surface (XMLTV offsets, XC `server_info`/`time_now`/EPG timestamps) stay UTC by design. The new `DISPATCHARR_TIME_ZONE` env var (normalized from standard `TZ`) seeds and mirrors the UI's System → Time Zone setting — it does not and cannot affect database storage or wire timestamps.

**Why the change touches files beyond the new module — each hunk exists because it is the only layer that can do its job:**

- **`dispatcharr/display_timezone.py` (new — the core of the change):** an in-process cached display zone plus `DisplayTimezoneFormatter`, which renders `{asctime}` via `datetime`/`ZoneInfo` from that cache — never `time.localtime()` (unfixable per above) and never a database query at emit time (see the design notes below). It is a small standalone module because dictConfig instantiates formatters during `django.setup()`, *before* the app registry populates — the host module must be import-light, which rules out `dispatcharr/utils.py` and anything under `apps/`/`core/` (they import models at module scope). Same shape as the existing `dispatcharr/db/process_label.py`, which settings.py already imports from.

- **`dispatcharr/settings.py` (+7, declarative only):** captures `DISPATCHARR_DISPLAY_TZ` during the settings module's own execution — the only moment the real environment value is observable in a Django process, since Django's `Settings.__init__` performs the `TZ` re-stamp immediately after this module finishes loading — and points the existing `verbose` formatter at the new class by dotted path.

- **`docker/entrypoint.sh`:** normalizes `DISPATCHARR_TIME_ZONE` from the standard `TZ` env (mirroring the `DISPATCHARR_LOG_LEVEL` block above it) and adds it to the whitelist. The entrypoint is the only process that ever sees the raw container environment — everything below the `su -` receives only what the whitelist re-exports via `/etc/environment` and `/etc/profile.d` — so this is structurally the only place the operator's `-e TZ=...` can be observed and converted into something that survives into the app tree.

- **`core/signals.py` (three receivers):** the formatter's cache must be refreshed from contexts that are provably safe to run a database query — `request_started` covers every uwsgi worker (rate-limited to one lookup per ~5 s inside the refresh), celery `task_prerun` covers task workers, and a `post_save` receiver on `CoreSettings` applies a UI change instantly in the process that saved it. These are module-level functions in the module `CoreConfig.ready()` already imports (and which already hosts a `post_save`-on-`CoreSettings` receiver), so registration cannot be garbage-collected — the previous timezone receiver in this codebase was a weakly-referenced nested closure that died silently at startup, which is part of how #651 happened.

- **`core/migrations/0020_...` (fresh-install seed):** without this hunk, the migration that introduced consolidated settings manufactures a stored `time_zone: "UTC"` row on every new install, so the UI setting always "wins" with UTC and `TZ` stays silently ignored until a human opens System → Time Zone — the headline of this issue would survive the fix. The change swaps only the *fallback default* passed to `get_value("system-time-zone", ...)`: legacy installs migrating a real stored value keep it, databases that already ran the migration never re-execute it, and only genuinely fresh installs take the env-derived default (validated through `zoneinfo` with a UTC fallback, so a typo'd `TZ` can never write an invalid zone into a setting that celery-beat's crontab validation later consumes). It must be this migration rather than a new one because this is the only point in the system that can distinguish a fresh install from a user who deliberately chose UTC — a later migration sees an existing `"UTC"` row and cannot tell the two apart.

**Two design decisions that came out of adversarial testing rather than taste:**

- **`TZ` itself is deliberately NOT whitelisted.** pam_env would hand it to every `su -` child — including initdb/Postgres — silently flipping the database server timezone on fresh installs. That is the precondition for the EPG-offset corruption class in #651, and with an invalid zone it makes psycopg emit "unknown PostgreSQL timezone" warnings from inside its connection lock. Verified empirically both ways: with `TZ` whitelisted, a fresh install ran its DB sessions at the container zone; with this PR, initdb selects `Etc/UTC` and the app connection's session timezone stays `Etc/UTC` even with `TZ=Pacific/Auckland` set.

- **The formatter never queries the database at emit time.** An earlier iteration of this branch did, and a fresh install with `TZ=Not/AZone` reproducibly deadlocked first-boot migrations: psycopg's unknown-timezone warning is logged while its non-reentrant connection lock is held, the formatter's `CoreSettings` query then needs the same lock, and boot wedges forever. The general hazard: any emit-time ORM call in a log formatter can self-deadlock whenever a log record originates inside a database call. Hence the out-of-band refresh design above.

**The XC API surface (`server_info.timezone`) is deliberately not changed.** It is a machine-consumption protocol field — clients localize for display — and `xc_get_epg` emits bare timestamp strings with no offset markers, so "everything UTC, declared UTC" is the only contract every client can consume safely. Current dev already pins that triple to UTC on purpose (the fix for the predecessor plugin's double-conversion bug), and `parse_catchup_timestamp` has no zone validation that would guard a change. Details in the issue thread.

**Known residuals** are tracked in #1447 rather than inline here: uWSGI's native access log (C-level `ftime`, rendered outside Python logging) and celery's internal `worker_log_format` lines keep UTC — the mechanics of why this fix cannot reach them, and the fix options, are in that issue, along with the now-dead `tzlocal` dependency cleanup. One minor note kept here: idle workers adopt a changed UI zone on their next request/task (cosmetic for a once-ever setting change).

## Related Issue

Closes #1439. Related: #651 (the DB/session UTC invariant this PR deliberately preserves — see the "TZ not whitelisted" rationale).

## How was it tested?

Six rounds of stock-vs-patched A/B on the current `dev` image (fresh-`/data` AIO containers; the base image was verified byte-identical to `upstream/dev` in every touched file before overlaying), with the adversarial scenarios re-run on the exact final SHA:

- **Bug reproduction (stock):** with `TZ=Pacific/Auckland`, app log lines stamp UTC against an NZ wall clock; setting the UI's System → Time Zone changes nothing either — both halves of the report confirmed.
- **Fix:** a fresh install with `TZ=Pacific/Auckland` seeds System → Time Zone to Pacific/Auckland via the migration and logs NZ wall-clock time from first boot with zero UI interaction; changing the UI setting to Europe/Zurich flipped subsequent log lines to CEST on the first request ~2 s later, no restart.
- **Invariant check:** on the same fresh install, initdb selects `Etc/UTC` and the app connection reports `SHOW timezone` → `Etc/UTC` (and `reset_val` = `Etc/UTC`) — the container zone never reaches Postgres.
- **Guard case:** `TZ=Not/AZone` boots clean end-to-end (168/168 migrations, uwsgi serving), seeds UTC, logs UTC, no psycopg warnings or tracebacks. This exact case deadlocked an earlier iteration of the branch — it is what drove the two design decisions above — and is reproducibly clean on the final SHA.
- **Control:** no `TZ` env → seeds UTC, logs UTC, clean boot (the seed invents nothing).
- **Unchanged surfaces:** XC `player_api.php` still reports `server_info.timezone: "UTC"` with a coherent UTC `time_now`; uWSGI's own `uwsgi.requests` access-log lines remain UTC alongside localized app lines (expected residual, captured same-instant in both formats).
- **Fresh-install migrate safety:** full first-boot logs grepped for `Traceback`/`ProgrammingError`/`OperationalError` — clean in every scenario.
- **Tests:** 6 new regression tests (`core/tests/test_display_timezone.py`) covering the env-capture fallback, signal-driven refresh, DB-error and invalid-zone keep-previous behavior, refresh interval, and datefmt handling; full `core.tests` suite 27/27 (21 pre-existing + 6 new), zero regressions vs the stock baseline.

## Checklist

- [ ] Backend: migrations are included if any models were changed (N/A — no model changes; an existing data migration's fresh-install default is adjusted, see Description)
- [ ] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — no API changes)
- [ ] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) (N/A — backend only)


nagelm:
I scanned past issues and it seems the AIO entrypoint's su dispatch whitelist is a common failure mode that drops any env var not explicitly listed (so these keep cropping up one at a time) - TZ here (#1439), DATABASE_POOL_CONN_MAX_LIFETIME over in #1418.

If you want a more robust fix I could rewrite as a prefix-allowlist for the config namespaces (DISPATCHARR_* / DATABASE_* / …) so new AIO knobs forward automatically? 

(Or turn to the exec-based drop that the LSIO/official images seem to use (gosu/su-exec/s6-setuidgid) sidesteps the whitelist entirely by preserving the env in-process, but that's a bigger change.)

sethwv:
Hello, if you're still interested in this one being reviewed/merged, please update off of current dev. Thanks!

nagelm:
Update to current dev head in first push. Second push removed some extraneous reformatting which is out of scope. 

Let me know if you have any further feedback.

SergeantPanda (commented):
Nice work on this overall. The approach looks solid.

One issue though: `refresh_log_display_zone_on_settings_change` is registered before `handle_coresettings_cache_invalidation`, and it refreshes via `get_system_time_zone()` → `_get_group()`, which prefers the Redis group cache. On a warm cache, that refresh can still see the pre-save value and pin the old zone in the worker that handled the UI change until the next refresh.

Can you set the display zone from `instance.value` here instead (or invalidate first)?

nagelm:
Fixed as you suggested by applying the zone straight from instance.value (new set_display_zone() in display_timezone.py), which also removes any dependence on receiver registration order. 

Added a regression test that reproduces a stale-read scenario. It fails on the previous read-back behaviour and passes with this change.

SergeantPanda (approved):
Ok, I am fine with this being done but I would argue this wasn't ever really a bug. 

With this change, it will certainly look better to a user looking at the logs (they wont have to convert the time) but, we can not display the offset properly either. Ideally we'd want to display the offset if we were not UTC.

Due to the way that timezones are passed through, we can not display an offset with uWSGI, it believes it is running UTC.

nagelm:
> Ok, I am fine with this being done but I would argue this wasn't ever really a bug.
> 
> With this change, it will certainly look better to a user looking at the logs (they wont have to convert the time) but, we can not display the offset properly either. Ideally we'd want to display the offset if we were not UTC.
> 
> Due to the way that timezones are passed through, we can not display an offset with uWSGI, it believes it is running UTC.

Agreed. I did think about this but it would probably need some new tooling to do a better fix (bigger blast radius with lots more file changes). After some of my other logging things have landed I will return to this harder problem and discuss tooling/options for a solution that aggregates logging across Django/uWSGI/celery (ill raise first in a feature request).

PR #1466, opened 2026-07-22
fix(dvr): Retry the run_recording idempotency guard on transient DB errors

### LINKED ISSUE #1464: [Bug]: run_recording idempotency guard aborts on a transient DB stall — scheduled recording silently never starts
### Describe the bug

A scheduled recording can fail to start entirely, with no ffmpeg attempt, no status change on the Recording row, and nothing visible in the UI — the only trace is a single log line:

```
ERROR apps.channels.tasks Idempotency guard DB check failed for recording 396 — aborting to prevent potential duplicate.
```

Root cause is the idempotency guard at the top of run_recording (apps/channels/tasks.py). It does a single Recording.objects.filter(...).first() in a try/except and returns on any exception, "to prevent potential duplicate." That fail-closed instinct is right, but a *transient* DB hiccup at exactly the scheduled start time permanently kills the recording: the task returns, nothing reschedules it, and the Recording row is left with no status/started_at — from the UI it just looks like the recording never happened. I lost a daily news recording to this on July 2 (the row had scheduled metadata only; the DB load that caused the stall had cleared seconds later).

There is a second, related failure on the AIO image: with gevent monkey-patching, the guard's DB call on a freshly spawned worker greenlet can die with LoopExit ("this operation would block forever") when psycopg yields to a hub that has no timer handle registered yet — same net effect, one log line and a silently lost recording.

Both are fixed by the same small hardening, which I have been running locally since July 2 with no recurrence:

1. wrap the guard query in gevent.Timeout(20) so a stalled call becomes a bounded, catchable timeout (and the hub always has a timer handle, which also eliminates the LoopExit case), and
2. retry the check 3x with a short%

The window is timing-dependent (the DB has to stall in the instant the task fires), so it is easiest to force:
1 - Schedule a recording a minute out.
2 - Just before it fires, hold the DB busy (e.g. select pg_sleep(30) inside a lock on the recording row, or briefly pause the postgres process).
3 - The task logs "Idempotency guard DB check failed ... — aborting" once and returns; the recording never starts and its row never gets a status, even though the DB is healthy again seconds later.

### Dispatcharr Version

dev (2026-07-21 image); first hit on 0.27.2 (build 20260701221056)

### Docker Version

Docker version 29.5.0, build 98f1464

Not using

### Client Information

N/A — server-side DVR scheduling, no client involved.

nagelm:
## Description

A scheduled recording is permanently lost if the database stalls at the moment `run_recording` fires. The idempotency guard at the top of the task performs a single DB check inside a bare `try/except` and returns on any exception, so one transient `OperationalError` aborts the task with a single log line — nothing re-dispatches it, the Recording row never receives a status, and from the UI the recording simply never happened. `recover_recordings_on_startup` cannot help unless the server happens to restart during the recording window.

This PR wraps the guard's check in the module's existing `_db_retry` helper. The guard uses a wider backoff window (5 attempts over roughly 30 seconds) than the in-recording retries, because unlike those, nothing retries the task itself if the guard gives up. Exhausted retries still fail closed, preserving the duplicate-prevention guarantee the guard exists for. The retry constants follow the same local-constant pattern as the existing `_dvr_db_*` values and the label follows the `DVR recording {id}: <operation>` convention of the other `_db_retry` call sites.

## Related Issue

Closes #1464

## How was it tested?

- **Reproduced the bug deterministically on the current `dev` image** before changing anything: created a Recording scheduled 90 seconds out (via the normal `post_save` → ClockedSchedule/Beat path), stopped Postgres with `pg_ctl stop -m immediate` 5 seconds before the scheduled start, and restarted it 15 seconds after. Beat dispatched on time from its in-memory schedule, the guard hit the dead socket, logged exactly one `Idempotency guard DB check failed ... aborting` line, and returned. The row's `custom_properties` remained `{}` past `end_time`, no ffmpeg was ever spawned, and nothing re-dispatched the task — the recording was silently lost even though the DB was healthy again 15 seconds after the start time.
- **Repeated the identical outage timeline with this fix applied** in the same container: the guard logged `failed, retrying in 2s (1/5)` through `16s (4/5)` during the outage, the fifth attempt succeeded once Postgres was back, and the recording started ~30 seconds late with `status: "recording"` and `started_at` set — delayed instead of lost.
- **Unit tests**: added `IdempotencyGuardRetryTests` to `apps/channels/tests/test_db_retry.py` covering both paths — a transient `OperationalError` at fire time must not kill the recording, and a sustained outage must still fail closed after exhausting retries. Ran the full module via `scripts/ci_bootstrap_backend.sh apps.channels.tests.test_db_retry` in the `base-dev` image: 15/15 pass.
- Full backend suite run in the same harness (`scripts/ci_bootstrap_backend.sh`): 1542 tests, all passing.

## Checklist


nagelm:
**Not covered (residuals):**

1. The gevent `LoopExit` mode the reporter hit on the AIO image. `LoopExit` isn't a `django.db` exception, so `_db_retry` doesn't catch it and the guard still aborts on the first failure in that mode. Fixing it properly needs gevent-specific handling (e.g. the reporter's `gevent.Timeout` suggestion), and I didn't want to pull gevent into `tasks.py` where prefork/modular workers also run this code.
If you'd prefer #1464 to stay open for that half, I'll change the "Closes" reference to a plain mention.

2. `django.db.InterfaceError` ("connection already closed"-class driver errors under psycopg) is a sibling of `OperationalError` which isn't covered. Since `_db_retry` already resets connections between attempts (`close_old_connections()`), which is exactly the remedy for a dead-connection state, widening the clause to `except (OperationalError, InterfaceError)` looks safe (?) and would also benefit the existing recovery and metadata-save call sites.

SergeantPanda (commented):
I made a couple of changes to catch the additional error you mentioned above as well as add a couple of tests. This should now fully close issue 1464.

Thank you!

Issue #579, opened 2025-10-22
[Bug]: value too long for type character varying(255)


clshtws:
### Describe the bug

When attempting to load an XMLTV file provided by ErsatzTV, a SQL error appears in the log, and the associated channel does not load any guide data on multiple clients (IPTV Pro, TiviMate, etc.)

[dispatcharr-log.txt](https://github.com/user-attachments/files/23062151/dispatcharr-log.txt)

### How can we recreate this bug?

1. Import XML file through the 'M3U & EPG Manager'
2. During the 'Parsing' process, an error appears in the docker logs
3. Assign the EPG to a channel
4. Viewing the channel in any IPTV app results in no guide data showing for the channel

[ersatztv-xmltv.xml](https://github.com/user-attachments/files/23062152/ersatztv-xmltv.xml)

### Dispatcharr Version

0.11.0

### Docker Version

28.3.3, build 980b856

### What's in your Docker Compose file?

N/A

### Client Information

- TiviMate 5.2.0
- IPTV Pro 9.1.10

SergeantPanda:
Available in dev.

clshtws:
Confirming this fix works perfectly! Incredibly appreciative of the speedy turn-around and grateful for all of the efforts put into keeping Dispatcharr the best in class!!! ☺️

SergeantPanda:
Thank you for reporting back! We try our best 🙂

Issue #519, opened 2025-10-04
[Bug]: index row size 5240 exceeds btree version 4 maximum 2704 for index


sandeepb1:
### Describe the bug

Error message when bulk creating channels

`dispatcharr  | 2025-10-04 23:29:35,853 ERROR apps.channels.tasks Bulk channel creation failed: index row size 5240 exceeds btree version 4 maximum 2704 for index "dispatcharr_channels_logo_url_key"
dispatcharr  | DETAIL:  Index row references tuple (131,26) in relation "dispatcharr_channels_logo".
dispatcharr  | HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
dispatcharr  | Consider a function index of an MD5 hash of the value, or use full text indexing.
dispatcharr  |
dispatcharr  | 2025-10-04 23:29:35,862 ERROR celery.app.trace Task apps.channels.tasks.bulk_create_channels_from_streams[d7bdbf4b-c73b-48a1-84fb-e179eae4baba] raised unexpected: OperationalError('index row size 5240 exceeds btree version 4 maximum 2704 for index "dispatcharr_channels_logo_url_key"\nDETAIL:  Index row references tuple (131,26) in relation "dispatcharr_channels_logo".\nHINT:  Values larger than 1/3 of a buffer page cannot be indexed.\nConsider a function index of an MD5 hash of the value, or use full text indexing.\n')
dispatcharr  | Traceback (most recent call last):
dispatcharr  |   File "/dispatcharrpy/lib/python3.13/site-packages/django/db/backends/utils.py", line 105, in _execute
dispatcharr  |     return self.cursor.execute(sql, params)
dispatcharr  |            ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
dispatcharr  | psycopg2.errors.ProgramLimitExceeded: index row size 5240 exceeds btree version 4 maximum 2704 for index "dispatcharr_channels_logo_url_key"
dispatcharr  | DETAIL:  Index row references tuple (131,26) in relation "dispatcharr_channels_logo".
dispatcharr  | HINT:  Values larger than 1/3 of a buffer page cannot be indexed.
dispatcharr  | Consider a function index of an MD5 hash of the value, or use full text indexing.
dispatcharr  |
dispatcharr  |
dispatcharr  | The above exception was the direct cause of the following exception:
dispatcharr  |
dispatcharr  | Traceback (most recent call last):
[...]

PR #1471, opened 2026-07-25
fix(channels,vod): widen Logo/VODLogo name to TextField (#1470)

### LINKED ISSUE #1470: [Bug]: Bulk channel creation fails with "value too long for type character varying(255)" on dispatcharr_channels_logo.name column
### Describe the bug

When attempting to bulk-create channels from streams, the operation fails with a PostgreSQL StringDataRightTruncation error. The error occurs because the name column in the dispatcharr_channels_logo table is defined as varchar(255), but some channel logo names (derived from channel names in the M3U source) exceed 255 characters.

This appears to be a partial regression / incomplete fix from issues #519 and #579, which increased the url column length but did not address the name column.

Expected behavior: Bulk channel creation should succeed regardless of the length of logo names or URLs provided by M3U sources. Columns receiving provider-supplied data should use text type or a sufficiently large varchar limit.

Workaround: Manually altering the columns to text type resolves the issue:
```sql
ALTER TABLE dispatcharr_channels_logo ALTER COLUMN name TYPE text;
ALTER TABLE dispatcharr_channels_logo ALTER COLUMN url TYPE text;
```

Relevant logs:
```
2026-07-23 14:50:10.671 UTC [18477] ERROR: value too long for type character varying(255)
2026-07-23 14:50:10.671 UTC [18477] STATEMENT: INSERT INTO "dispatcharr_channels_logo" ("name", "url") SELECT * FROM UNNEST(...)::text[] ON CONFLICT DO NOTHING
2026-07-23 14:50:10,672 ERROR apps.channels.tasks Bulk channel creation failed: value too long for type character varying(255)

Traceback (most recent call last):
  File "/dispatcharrpy/lib/python3.13/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "/dispatcharrpy/lib/python3.13/site-packages/psycopg/cursor.py", line 117, in execute
    raise ex.with_traceback(None)
psycopg.errors.StringDataRightTruncation: value too long for type character varying(255)

django.db.utils.DataError: value too long for type character varying(255)
```

1. Add an M3U source whose playlist contains channels with logo names (or channel names used as logo names) exceeding 255 characters in length
2. Go to Channels → Streams
3. Select all (or a large batch of) streams
4. Click Create Channels
5. Select a channel profile and numbering mode
6. Click Create channel
7. Observe the red modal popup: "Bulk Channel Creation Failed — value too long for type character varying(255)"
8. Check Docker logs for the full StringDataRightTruncation traceback

### Dispatcharr Version

0.28.1

### Docker Version

Issue is not Docker-version-dependent. Reproduced on Docker Desktop for Windows with Docker Engine v28.3.2

services:
  dispatcharr:
    container_name: dispatcharr
    volumes:
      - dispatcharr_data:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
    restart: unless-stopped
    ports:
      - 9191:9191

volumes:
  dispatcharr_data:

### Client Information

Not applicable, this bug occurs during channel creation within the Dispatcharr web UI and does not depend on any specific client/player. The error is triggered server-side during the bulk channel creation Celery task.

Lbubeer:
## Description

#519 and #579 widened `Logo.url` to `text` but left `Logo.name` at `varchar(255)`, so bulk channel creation still aborts the moment an M3U source supplies a long name:

    django.db.utils.DataError: value too long for type character varying(255)

`name` is populated straight from provider data wherever a logo gets created. The path the reporter hit is the bulk-creation branch in `apps/channels/tasks.py`, which inserts a `Logo` row per channel logo; `VODLogo.name` carries the same column over the VOD bulk-import path (`apps/vod`), where the name comes from provider movie/series titles and is just as unbounded.

Both columns are widened to `TextField`, matching the type `url` already has and what the issue describes ("columns receiving provider-supplied data should use text-type"). There's nothing else depending on the 255 chars: the only length guard in the logo path is `validate_logo_url`, and that protects the `url` btree index (a 2000-char ceiling) — it's left untouched. `__str__` returns `self.name`, which is unaffected by the column type.

Migrations are included for both apps (`0039` and `0005`), against `dev`'s current tips (`0038` and `0004`).

## Related Issue

Closes #1470

## How was it tested?

Reproduced the original failure first, then fixed it. On real PostgreSQL 16, against the old `varchar(255)` schema, inserting a `Logo` row with a ~600-char multibyte name raises the exact error from the issue (`value too long for type character varying(255)`). After applying the two new migrations, `information_schema.columns` reports both `name` columns as `text`; the same INSERT (600-char and 5000-char names) now succeeds, and the existing `ALTER` preserved pre-existing rows in place.

- `python manage.py makemigrations --check --dry-run` → No changes detected — the migration end-state matches the models.
- `python manage.py migrate` on a fresh database → the whole chain applies cleanly, including the two new files in order.
- `python manage.py test` (Python 3.13, Django 6.0.6, real Postgres, Redis/valkey) → 1543 tests, OK. (I did not see any flake on this run; happy to post the raw tail if useful.)

The manual `ALTER TABLE … ALTER COLUMN name TYPE text` workaround in the issue is equivalent to what the migration does.

## Checklist

- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — no new endpoints, only a model field-type change with an existing serializer)
- [x] Frontend: ESLint and Prettier pass cleanly (N/A — no frontend changes)
- [x] Tests are included for new functionality (covered by existing logo/VOD suites; the fix is a schema widening — behavior unchanged except the removed length ceiling)

> Disclosure: I developed this change with AI assistance, but I have reviewed and understand every line, tested it on my own setup, and can explain and adjust any part on request.


SergeantPanda (changes requested):
Thank you for the PR.

I would rather not change the logo names to text fields. I do not believe it adds any real value as they do not need to be unique like URLs do.

I think we should truncate them on ingest instead to match the 255 char length (more than enough for a logo name)

A shared helper that does so (we may already have something that can be reused) I think is the proper solution here.

Please let me know if you'd like to take that on otherwise we can take care of it.

SergeantPanda:
Closing this due to no response on the changes requested. We will implement the fix.

PR #1480, opened 2026-07-26
Show currently-airing EPG title on each stream row

JHeat:
## Summary

Some IPTV providers (e.g. IPTVboss-style PPV/event feeds) deliberately keep channel and stream names as generic placeholders (`PPV EVENT 04`, `LIVE EVENT 12`) and instead push the actual event information through EPG data on a rolling basis — the channel identity stays fixed, but its EPG programme updates to whatever's currently airing. That's a reasonable design on the provider side and gives genuinely good, timely EPG data — but it leaves the Channels page unable to show which stream currently carries which event without cross-referencing the EPG separately; the stream list gives no hint beyond the generic name.

This adds a read-only `epg_now_title` field to `StreamSerializer`, resolved from the stream's own `tvg_id` against `ProgramData` for the currently-active time window (`start_time <= now < end_time`), and surfaces it as a small badge next to the existing quality/URL badges in the expanded stream row (`ChannelTableStreams.jsx`) — only when present. It's the same underlying EPG data these providers already supply well; this just makes it visible where it's actually useful (picking the right stream) without requiring anything be renamed.

- Batched per-channel in `ChannelSerializer.get_streams` (one query per channel's stream list, not one per stream) to avoid N+1 queries when a channel page loads with `include_streams=true`.
- Provider-agnostic: shows the raw EPG title verbatim, no text cleanup — that kind of provider-specific string massaging belongs at the provider/plugin layer, not core.

## Test plan

- [x] Verified live against a real Dispatcharr instance (v0.28.2): `GET /api/channels/channels/{id}/?include_streams=true` returns `epg_now_title` per stream, correctly resolving e.g. `PPV EVENT 08` → `"AEW Redemption ᴺᵉʷ"`.
- [x] Confirmed the badge renders correctly in the Channels page stream list after a full frontend rebuild + deploy.
- [x] Confirmed no `epg_now_title` badge shown when a stream has no `tvg_id` or no currently-active `ProgramData` (field returns `null`, frontend conditionally renders).

Happy to adjust naming/placement/scope per maintainer preference — opening this for consideration since it seemed like a generally useful, low-risk addition for anyone using providers with this generic-placeholder pattern.

### DIFF
--- a/apps/channels/serializers.py
+++ b/apps/channels/serializers.py
@@ -17,7 +17,7 @@ from .models import (
 )
 from apps.epg.serializers import EPGDataSerializer
 from core.models import StreamProfile
-from apps.epg.models import EPGData
+from apps.epg.models import EPGData, ProgramData
 from django.db import connection, transaction
 from django.urls import reverse
 from rest_framework import serializers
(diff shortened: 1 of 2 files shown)


JHeat:
Closing this in favor of opening an issue instead, per your CONTRIBUTING.md guidance to discuss before submitting a PR (and to be upfront that this was AI-assisted/CLI-created, which your bot flagged). Issue with the full writeup, working reference branch, and screenshots incoming.

PR #1486, opened 2026-07-27
Authentication complexity

write-erase:
## Description

This change relaxes username and password validation to support credentials containing common special characters used by Xtreme Codes providers.

## Related Issue

None.

## How was it tested?

I ran `npm lint` to review my changes and 170 problems (114 errors, 56 warnings) returned. None of the findings were related to my change. I verified this by using `npm run lint | grep UserUtils.js`.

## Checklist

- [X] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`)
- [X] Tests are included for new functionality


SergeantPanda (changes requested):
Thank you for the PR.

I am certainly not opposed to allowing *some* special characters but I want to be extremely cautious here as there are a LOT of clients that users could use (many of which aren't of high quality) and could break when using certain characters.

It also becomes complicated if we need to escape them in M3U outputs and accept them in our XC inputs. This is why most XC providers will only ever output alpha-numeric. 

That being said, there are a few characters that *should* be safe to use (`_ . @ -`) (Django supports a + but we should probably avoid that as it could be interpretted as a space with query strings and with some clients). I would also think we could make it so the username and password have the same rules applied to simplify things as they'd both have the same issues for XC clients.

We should also add to the serializer to validate the same when saving to avoid invalid characters getting added via API (this is not currently being done so would be an additional hardening, worth doing now imo)

write-erase:
Backend tests could not be run natively on macOS ARM64 because the current `torch==2.12.1+cpu` dependency does not provide a compatible wheel for macOS ARM64. I attempted to run them using `uv run python manage.py test apps.accounts`.

I also want to make it known that I am using ChatGPT to help construct this PR, and my coding knowledge is currently very limited. I mention this because if you see something that is not correct or if there is a better way you would like it implemented, please let me know. I am happy to go back to the drawing board and make the necessary changes.

frontend/src/utils/forms/UserUtils.js
```diff
@@ -135,19 +135,19 @@ export const getFormValidators = (user) => {
     username: !values.username
       ? 'Username is required'
       : values.user_level == USER_LEVELS.STREAMER &&
```
SergeantPanda: This shouldn't affect only streamer users, it should apply to all users.

frontend/src/utils/forms/UserUtils.js
```diff
@@ -135,19 +135,19 @@ export const getFormValidators = (user) => {
...
-          !values.username.match(/^[a-z0-9]+$/i)
-        ? 'Streamer username must be alphanumeric'
+          !values.username.match(/^[A-Za-z0-9._@-]+$/)
+        ? 'Streamer username contains unsupported characters'
         : null,
     password:
       !user && !values.password && values.user_level != USER_LEVELS.STREAMER
```
SergeantPanda, line 141: Same for the password, this should apply to all users.

apps/accounts/serializers.py
```diff
@@ -74,6 +76,13 @@ class Meta:
             "last_name",
         ]
 
+    def validate_username(self, value):
```
SergeantPanda, line 79: It'd be nice to have some good tests for usernames to ensure proper validation.

apps/accounts/serializers.py
```diff
@@ -74,6 +76,13 @@ class Meta:
...
+        if not SAFE_CREDENTIAL_RE.fullmatch(value):
+            raise serializers.ValidationError(
+                "Username contains unsupported characters"
+            )
+        return value
+
     def validate_custom_properties(self, value):
```
SergeantPanda, line 86: Tests for password would be nice as well.

SergeantPanda (commented):
That is looking better for sure. I added a couple of comments throughout to make some small changes but we are about ready to get it merged.

Thank you!

frontend/src/utils/forms/UserUtils.js
write-erase: Just to clarify, if this is applied to all users, does that mean Streamer accounts should also require a password, even though they can't log into the UI and the password would never be used?

SergeantPanda: Sorry for the confusion! I meant the character allow-list, not the “password required” rule. 
Streamers should still not need a UI login password. Please apply the safe charset (A-Za-z0-9_@.-) to usernames for all user levels. xc_password already applies to everyone; leave the streamer login-password exemption alone.

SergeantPanda (approved):
Looks good now, thank you!

PR #1489, opened 2026-07-28
feat(plugins): Plugins "V3"

### LINKED ISSUE #699: [Feature]: PLUGIN Support for Long-Running/Async Plugin Actions with Progress Notifications

The current plugin system is designed for synchronous operations that complete within HTTP request timeout limits   (typically 60-120 seconds). This creates significant limitations for plugins that need to perform long-running operations (5-15+ minutes), such as bulk data processing, API synchronization, or large dataset operations. This affects plugin: https://github.com/PiratesIRC/Stream-Mapparr 

## Current Behavior

  When a plugin action is executed:
  1. Frontend makes a POST request to `/api/plugins/plugins/<key>/run/`
  2. Button shows "loading" state during HTTP request
  3. Plugin `run()` method executes synchronously (blocks request)
  4. HTTP response returns the result
  5. Frontend displays notification from HTTP response
  6. Button returns to normal state

  **This works well for quick operations** but has critical issues for long-running tasks.

  ## Problem / Use Case

  I've developed a plugin (Stream-Mapparr) that matches IPTV streams to channels using fuzzy matching. For users with large channel lists (1000+ channels, 5000+ streams), operations take 5-15 minutes.

  **Issues encountered:**
  - Synchronous execution → HTTP timeouts after 60-120 seconds
  - Timeout causes "Broken pipe" errors and failed operations
  - Operations cannot be completed successfully
  - Users get error notifications even though processing could succeed

  **Current workaround:**
  - Run operations in background daemon threads
  - Return HTTP response immediately
  - Operation continues in the background
  - Users must monitor Docker logs for completion

  **Problems with workaround:**
  - No completion notification in frontend (only HTTP response notifications work)
  - Buttons don't stay disabled during operation
  - No progress indication in UI
  - Users can accidentally start multiple operations
  - Poor UX - must watch Docker logs instead of UI

  Add support for asynchronous plugin actions with proper frontend integration:

  ### Backend Changes

  1. **New optional plugin method signature:**
  ```python
  class Plugin:
      # Existing synchronous method (backwards compatible)
      def run(self, action, params, context):
          return {"status": "success", "message": "Done"}

      # New async method (opt-in)
      async def run_async(self, action, params, context):
          # Returns task ID immediately
          task_id = uuid4()
          # Start background task
          asyncio.create_task(self._long_operation(task_id, action, params))
          return {"status": "started", "task_id": task_id}
```
  2. Task state tracking:
  ```python
  # Store task state in database or cache
  task_states = {
      "task_id": {
          "status": "running",  # or "completed", "failed"
          "progress": 50,       # percentage
          "message": "Processing 500/1000 items...",
          "result": None,       # populated on completion
          "started_at": datetime,
          "updated_at": datetime
      }
  }
```
  3. New API endpoint for status polling or WebSocket updates:
  GET /api/plugins/tasks/<task_id>/status/
  WebSocket /ws/plugins/tasks/<task_id>/

  Frontend Changes

  1. Plugin action response handling:
  const resp = await onRunAction(plugin.key, action.id);
  ```python
  if (resp?.task_id) {
    // Async operation - show progress UI
    showTaskProgress(resp.task_id);
  } else {
    // Sync operation - existing behavior
    showNotification(resp.message);
  }
```
  2. Task progress component:
   ```python
  function TaskProgress({ taskId }) {
    const [task, setTask] = useState(null);

    useEffect(() => {
      // Poll for status or use WebSocket
      const interval = setInterval(async () => {
        const status = await api.getTaskStatus(taskId);
        setTask(status);

        if (status.status === 'completed') {
          notifications.show({ message: status.result.message, color: 'green' });
          clearInterval(interval);
        }
      }, 1000);

      return () => clearInterval(interval);
    }, [taskId]);

    return <Progress value={task?.progress} label={task?.message} />;
  }
```
  3. Button state management:
   ```python
  // Disable buttons for specific action while task is running
  <Button
    disabled={!enabled || hasRunningTask(plugin.key, action.id)}
    loading={taskStates[action.id]?.status === 'running'}
  >
```
  Alternative: Simpler Polling Approach

  If WebSockets are too complex, a simpler polling mechanism:

  1. Plugin returns {"status": "started", "check_in": 5} (check in 5 seconds)
  2. Frontend polls /api/plugins/plugins/<key>/status/<action_id>/ every N seconds
  3. Plugin stores state in an instance variable or a database
  4. When complete, the status endpoint returns the final result
  5. Frontend shows notification when status becomes "completed."

  Benefits

-  Long-running operations work reliably - No timeout errors
-  Better UX - Progress visible in UI, completion notifications work
-  Safer - Buttons stay disabled, preventing duplicate operations
-  Backwards compatible - Existing plugins continue to work unchanged
-  Flexible - Plugins choose sync or async based on operation needs

### Additional context

Nothing Extra

### LINKED ISSUE #1244: Plugin @shared_task cron jobs silently miss ticks after worker restart

### LINKED ISSUE #1474: [Feature]: Plugin settings: group fields into collapsible sections, including an Advanced section

Plugin settings render as a single flat list. `frontend/src/components/Field.jsx` maps each entry of a plugin's `fields` array to one Mantine control and stacks them in declaration order. There is no container, no heading, and no way to collapse anything.

Dispatcharr's own Settings page does not work this way. `frontend/src/pages/Settings.jsx` wraps every settings group in a Mantine `Accordion`, and nests a second one inside UI Settings for Navigation. `frontend/src/components/forms/settings/ProxySettingsForm.jsx` keeps its less common options behind a `Collapse` toggled by a "Show Advanced Settings" button.

Plugins have neither. Any plugin with more than a few settings becomes one long scroll in which a rarely used expert option carries the same visual weight as the single switch the user came to change. Authors currently approximate structure with `info` fields used as pseudo headings, which produces a heading that cannot collapse, does not indicate what it contains, and adds more vertical space to the problem it is meant to solve.

Issue #621 describes the same pain from the other direction, noting that plugin containers grow large and cause scrolling. Its proposed fix is a side panel tree for moving between plugins, which is complementary to grouping fields within a single plugin.

Allow a plugin to declare sections, and allow those sections to collapse, reusing the patterns already present in core Settings.

An additive design that breaks no existing plugin:

1. Add a `section` field type to the `type` `ChoiceField` in `PluginFieldSerializer` (`apps/plugins/serializers.py`), alongside `string`, `number`, `boolean`, `select`, `text` and `info`.
2. Add two optional keys to `PluginFieldSerializer`: `section` (CharField), the id of the section a field belongs to, and `collapsed` (BooleanField, default false), meaningful only on a `section` entry, which renders that section closed on first paint.
3. In the frontend, group the flat `fields` array by each field's `section` value before rendering, and render each group inside an `AccordionItem` whose control is the section label. Fields with no `section` key render as they do today, above the first section.

An Advanced area then needs no separate mechanism. It is a section declared with `"collapsed": true`, which is behaviourally what `ProxySettingsForm.jsx` already does.

```json
{
  "fields": [
    { "id": "enabled", "label": "Enable plugin", "type": "boolean" },

    { "id": "sec_schedule", "label": "Schedule", "type": "section" },
    { "id": "run_at", "label": "Run at", "type": "string", "section": "sec_schedule" },

    { "id": "sec_advanced", "label": "Advanced", "type": "section", "collapsed": true },
    { "id": "batch_size", "label": "Batch size", "type": "number", "section": "sec_advanced" }
  ]
}
```

Storage is unaffected. `PluginConfig.settings` remains a flat id to value dict, `_merge_settings_with_defaults` in `apps/plugins/loader.py` is untouched, and a `section` entry stores no value of its own, as `info` already does. The change is presentation only.

Two details worth deciding explicitly:

- A field naming an undeclared `section` id should render ungrouped rather than disappear, so that a manifest error does not silently hide a setting.
- Open and closed state is a per user display preference and should not be written into `PluginConfig.settings`.

### Additional context

I maintain several plugins on my own instance, currently nine plugin config rows on v0.28.2. Settings list length is my largest usability complaint about the plugin UI. Some of those plugins carry twenty or more fields, of which roughly four are changed more than once.

This is one of several related requests I am filing. The common theme is that the core Settings page has form capabilities the plugin settings form lacks. The others cover a file upload field type, a multi select field type, a table field type, a reset to defaults control, and reordering of sections.

I scoped this request to the additive serializer approach because it requires no migration, no change to how plugin settings are persisted, and no change to any existing manifest. A plugin that declares no section renders exactly as it does today.

I am able to test an implementation against a live multi plugin install and report back.

Files referenced: `apps/plugins/serializers.py`, `frontend/src/components/Field.jsx`, `frontend/src/pages/Settings.jsx`, `frontend/src/components/forms/settings/ProxySettingsForm.jsx`.

### LINKED ISSUE #1476: [Feature]: Plugin settings: add a file field type that uploads into the plugin's own directory

A plugin cannot accept a file through its settings. `PluginFieldSerializer` in `apps/plugins/serializers.py` allows `string`, `number`, `boolean`, `select`, `text` and `info`, and `frontend/src/components/Field.jsx` renders those as `TextInput`, `NumberInput`, `Switch`, `Select` and `Textarea`. None of them accepts a file.

Dispatcharr already does this. `frontend/src/components/forms/settings/DvrSettingsForm.jsx` renders a Mantine `FileInput` for a custom `comskip.ini`, calls `uploadComskipIni`, and stores the resulting server side path in `comskip_custom_path`.

Plugin authors have two poor options. Either ask the user to paste a whole file into a `text` field, which becomes unusable past a few dozen lines, or ask the user to place the file on the Docker host and type an absolute container path into a `string` field. The second is what most authors choose, and it is hostile to the user: they must know that `/data` is a named volume with no Windows path, get the file in by some out of band means, get ownership right so the `dispatch` user can read it, and type a path with no validation. A typo produces a plugin that silently does nothing.

Common cases where a plugin needs a real file: an `.ini` config, a JSON rules or mapping document, a CSV import, a certificate bundle, an overlay image.

Add a `file` field type that uploads through the existing plugin API, lands in the plugin's own directory, and stores the resulting path as the field value.

1. Add `file` to the `type` `ChoiceField` in `PluginFieldSerializer`.
2. Add optional `accept` (CharField, a browser accept string such as `.ini,.json`) and `max_size` (IntegerField, bytes, enforced server side).
3. Add an upload endpoint under the existing plugin API (`apps/plugins/api_urls.py`, `apps/plugins/api_views.py`) taking the plugin key and field id, writing into that plugin's own directory, and returning the stored path. This mirrors what `uploadComskipIni` does for DVR settings.
4. Render the type as a Mantine `FileInput` in `Field.jsx`, following `DvrSettingsForm.jsx`.
5. Store only the resulting path, so the value stays a plain JSON string and nothing about how plugin settings are persisted has to change.

Points that matter more here than in my other requests, since this one accepts user supplied bytes and writes them to disk:

- Sanitise the filename and refuse path traversal. Neither a field id nor a browser supplied filename should be able to escape the plugin directory.
- Write as the user the application runs as. On my install a file created by a root process under `/data` is unreadable by the `dispatch` user that the uWSGI and Celery workers run as, and the failure mode is a silent empty directory rather than an error. An upload landing with the wrong owner would appear to succeed and then do nothing.
- Enforce `max_size` server side, with a sane default when the author sets none.
- Write into a dedicated subdirectory such as `uploads/`, so the endpoint can never overwrite `plugin.py`, `plugin.json` or `__init__.py`.
- Decide what happens to the previous file on replacement and on plugin uninstall. I would suggest overwrite in place, and removal along with the plugin directory, so files do not accumulate invisibly.

### Additional context

Issue #428 requested a `file_upload` field type and was closed as completed on 2025-09-28, with no linked commit and no comment. The type did not ship. As of v0.28.2 the `type` `ChoiceField` in `PluginFieldSerializer` still reads `["string", "number", "boolean", "select", "text", "info"]`, and `Field.jsx` has no file branch. I am the author of #428, and I am re-raising this one piece narrowly rather than reopening it.

#428 bundled file upload with tables, progress bars, live WebSocket updates, custom plugin routes and templates, plugin defined database models and plugin migrations. I now believe that breadth is why it did not land. The encouraging precedent is #616, a single narrow field level request, which did ship as the `input_type: password` handling in `Field.jsx`.

I would also accept a narrower version limited to text formats such as `.ini`, `.json`, `.csv` and `.txt` with a small size cap. That covers nearly all real cases and avoids taking on arbitrary binary uploads.

I run a multi plugin install, currently nine plugin config rows on v0.28.2, and can test any implementation, including the ownership behaviour described above, which is easy to get wrong and produces a silent failure rather than an error.

Files referenced: `apps/plugins/serializers.py`, `apps/plugins/api_urls.py`, `apps/plugins/api_views.py`, `frontend/src/components/Field.jsx`, `frontend/src/components/forms/settings/DvrSettingsForm.jsx`.

### LINKED ISSUE #1477: [Feature]: Plugin settings: add a multiselect field type

A plugin can offer a dropdown, but the user can select only one entry from it. `frontend/src/components/Field.jsx` renders the `select` type as a Mantine `Select`, which is single value, and coerces the value to a string. `PluginFieldOptionSerializer` in `apps/plugins/serializers.py` describes the option list, but no type consumes more than one option.

Dispatcharr already needs multi select and already ships it. `frontend/src/components/forms/settings/StreamSettingsForm.jsx` uses Mantine `MultiSelect` for the M3U Hash Key setting and joins the selected values.

Selecting several options is a common shape for plugin settings: which channel groups to act on, which M3U accounts to include, which EPG sources to scan, which incident types to alert on, which days a schedule runs. Authors have two workarounds, both poor:

- A `string` or `text` field holding a comma separated list. The user sees no list of valid values, gets no protection against a typo, and receives no feedback when one occurs. A misspelt group name matches nothing, and the plugin quietly does less than the user asked.
- One `boolean` field per option. This works only for short lists known at author time, and it inflates the form. On my install a channel group setting would need dozens of switches.

Add a `multiselect` field type that renders Mantine `MultiSelect`, using the option list plugins can already declare.

1. Add `multiselect` to the `type` `ChoiceField` in `PluginFieldSerializer`. The existing `options` key, already validated by `PluginFieldOptionSerializer`, is reused unchanged.
2. Add a `case 'multiselect'` in `Field.jsx` rendering a `MultiSelect` whose `data` is built from `field.options` exactly as the existing `select` case builds it.
3. Store the value as a JSON array of the selected option values. `default` is already a `JSONField` on `PluginFieldSerializer` and `PluginConfig.settings` is already a JSON field, so an array persists with no migration.

```json
{
  "id": "target_groups",
  "label": "Channel groups to process",
  "type": "multiselect",
  "default": [],
  "options": [
    { "value": "1854", "label": "US: News" },
    { "value": "1855", "label": "US: Sports" }
  ]
}
```

Two details worth deciding explicitly:

- The existing `select` case stringifies its value. A multiselect should not flatten to a comma joined string on the way into `PluginConfig.settings`, because a comma inside an option value would then corrupt the list on read. Storing a real JSON array avoids the problem, and the storage layer already supports it.
- A stored value naming an option the plugin no longer declares should be kept in settings but not shown as selected, so that a temporarily unavailable option is not silently and permanently discarded the first time the user opens the form and saves.

Optional `searchable` and `clearable` passthrough flags would help with long option lists, but they are not required for a first version.

### Additional context

This is the cheapest of the related requests I am filing, and I would rank it second in value. It reuses an option serializer that already exists, a Mantine component already used elsewhere in the codebase, a storage field that is already JSON, and a `default` key that is already a `JSONField`. As far as I can tell the change is one entry in a `ChoiceField` choices list and one `case` in `Field.jsx`.

It is also fully backwards compatible. No published plugin declares `type: "multiselect"`, so nothing currently installed changes behaviour.

I run a multi plugin install, currently nine plugin config rows on v0.28.2, and can test.

Files referenced: `apps/plugins/serializers.py`, `frontend/src/components/Field.jsx`, `frontend/src/components/forms/settings/StreamSettingsForm.jsx`.

### LINKED ISSUE #1478: [Feature]: Plugin settings: add a table field type for editing a list of rows

A plugin cannot present tabular data. Every type allowed by `PluginFieldSerializer` in `apps/plugins/serializers.py` holds a single scalar, and `frontend/src/components/Field.jsx` renders each one as a single control. There is no way to show a list of records, and no way to let the user add, edit, reorder or delete a row.

Dispatcharr does this for its own settings. `frontend/src/components/tables/StreamProfilesTable.jsx` renders stream profiles as a table with per row actions, backed by `frontend/src/components/forms/StreamProfile.jsx` as the row editor.

Many plugin settings are list of record shaped: find and replace rules, name alias maps, per group or per account overrides, notification routing rules. The only way to express any of them today is a `text` field holding raw JSON that the user hand edits. That is bad in a specific way. The user is editing a serialised data structure in a plain textarea with no schema, no column headings, no per field validation and no syntax checking. A single missing bracket invalidates the whole setting, and the failure surfaces at plugin run time rather than at save time, if it surfaces at all. I ship exactly this shape today, and it is the most error prone part of the plugins I maintain.

Add a `table` field type where the plugin declares the columns and the user edits rows.

1. Add `table` to the `type` `ChoiceField` in `PluginFieldSerializer`.
2. Add a `columns` key: a list of column definitions, each with an `id`, a `label`, and a `type` drawn from the existing scalar types. A `select` column reuses the existing `options` structure already validated by `PluginFieldOptionSerializer`. Reusing the existing types keeps the new surface area small and means the cell editors are controls `Field.jsx` already knows how to render.
3. Store the value as a JSON array of objects keyed by column id. `default` is already a `JSONField` and `PluginConfig.settings` is already a JSON field, so no migration is needed.
4. Render a compact table with the declared column headings, an add row control, a delete control per row, and inline cell editing.

```json
{
  "id": "rename_rules",
  "label": "Rename rules",
  "type": "table",
  "default": [],
  "columns": [
    { "id": "find", "label": "Find", "type": "string" },
    { "id": "replace", "label": "Replace", "type": "string" },
    { "id": "enabled", "label": "Enabled", "type": "boolean", "default": true }
  ]
}
```

Details worth deciding explicitly:

- Row order is usually meaningful, since rules are applied in sequence. The stored array order should be authoritative. Drag to reorder would be a natural addition given `@dnd-kit` is already a dependency, used in `frontend/src/components/forms/settings/NavOrderForm.jsx`, but it is not required for a first version as long as array order is preserved on save.
- A cell value that does not match its declared column type should be rejected at save time with a visible error rather than written through. Moving validation from plugin run time to save time is most of the value of this request.
- Columns removed by a plugin update will leave orphaned keys in stored rows. `_merge_settings_with_defaults` in `apps/plugins/loader.py` never prunes stored settings, so those keys persist. Ignoring unknown keys on read is safer than deleting them.

I would be satisfied with a deliberately minimal version: string and boolean columns, add and delete rows, no reordering, and type checking only. Even that replaces hand edited JSON in a textarea, which is the part that actually hurts.

### Additional context

Issue #428 requested a `table` field type with columns, sorting, filtering and row actions, and was closed as completed on 2025-09-28 with no linked commit and no comment. The type did not ship. As of v0.28.2 the `type` `ChoiceField` in `PluginFieldSerializer` still reads `["string", "number", "boolean", "select", "text", "info"]`, and `Field.jsx` has no table branch. I am the author of #428, and I am re-raising this one piece narrowly rather than reopening it.

I am also asking for less than #428 did. It bundled the table with file uploads, progress bars, live WebSocket updates, custom plugin routes and templates, plugin defined database models and migrations, and wanted the table sortable and filterable with row action buttons. I am asking only for an editable list of rows with declared columns. The encouraging precedent is #616, a single narrow field level request, which did ship as the `input_type: password` handling in `Field.jsx`.

This is the most expensive of the related requests I am filing and the one I would expect to wait longest. If it helps with ordering, the multi select and file upload requests remove the most pain for the least work.

I run a multi plugin install, currently nine plugin config rows on v0.28.2, several of which use a JSON in a textarea setting that this would replace, so I can test against real data.

Files referenced: `apps/plugins/serializers.py`, `frontend/src/components/Field.jsx`, `frontend/src/components/tables/StreamProfilesTable.jsx`, `frontend/src/components/forms/StreamProfile.jsx`, `apps/plugins/loader.py`.

### LINKED ISSUE #1479: [Feature]: Plugin settings: add a Reset to Defaults control

Once a user has changed a plugin's settings there is no way back to the author's defaults. The plugin settings form offers save only. Recovery means remembering every previous value, reading the plugin's documentation if it happens to list defaults, or uninstalling and reinstalling the plugin.

Core settings already solve this in three places. `frontend/src/components/forms/settings/UserLimitsForm.jsx` builds a defaults object from `USER_LIMITS_OPTIONS[key].default` and exposes a Reset to Defaults button. `ProxySettingsForm.jsx` has `resetProxySettingsToDefaults` behind a button of the same name. `NavOrderForm.jsx` has Reset to Default for navigation order.

The information needed is already present and already flowing. Every entry in `PluginFieldSerializer` carries an optional `default`, `_merge_settings_with_defaults` in `apps/plugins/loader.py` already reads it, it is already serialised out to the frontend, and `Field.jsx` already uses it as a fallback when a value is absent. Nothing exposes it as a deliberate user action.

There is a second reason this matters for plugins specifically. `PluginManager.update_settings` wholesale replaces with the posted dict, and `_merge_settings_with_defaults` only ever adds, so a value the user set for a field that a later plugin version removed stays in `PluginConfig.settings` indefinitely. A reset is the natural place to clear that residue, and today there is no way to clear it short of editing the database.

Add a Reset to Defaults control to the plugin settings form, matching the wording and placement already used by `UserLimitsForm.jsx` and `ProxySettingsForm.jsx`.

1. Reset every field to the `default` declared in its field definition. Fields that declare no `default` should return to empty rather than keep their current value, so that reset means the same thing for every field.
2. Drop stored keys that no longer correspond to any declared field. This is the only user facing way to clear the residue described above, and it is safe here because the user has explicitly asked to return to a known state.
3. Confirm before applying. Plugin settings can hold values that are tedious to reconstruct, such as a webhook URL or a long list of selections. `PluginActionSerializer` already carries a `confirm` key, so a confirmation dialog is consistent with how plugin operations already behave.
4. Stage the change pending a save rather than applying immediately, matching `resetProxySettingsToDefaults`, so the user can see the result before committing to it and can back out by navigating away.

A per section or per field reset would also be useful, particularly alongside an Advanced section, but a single form level reset is what I am asking for here.

### Additional context

This is the smallest of the related requests I am filing. The defaults are already declared, already validated, already sent to the frontend and already consumed as a fallback. The missing piece is a control that applies them deliberately. If any of these is worth doing as a quick addition alongside a larger change, it is probably this one.

The stale key cleanup in point 2 is the part I would most like to see kept rather than dropped for simplicity. I have been affected by it directly: a removed field's old value survived in `PluginConfig.settings` and shadowed a replacement value that reused the same key name, producing behaviour that matched neither the old setting nor the new one and took a while to track down. A reset that left the residue in place would not have helped.

I run a multi plugin install, currently nine plugin config rows on v0.28.2, and can test, including the stale key case.

Files referenced: `apps/plugins/loader.py`, `apps/plugins/serializers.py`, `frontend/src/components/Field.jsx`, `frontend/src/components/forms/settings/UserLimitsForm.jsx`, `frontend/src/components/forms/settings/ProxySettingsForm.jsx`.

sethwv:
## Description

WIP

## Related Issue

Closes #1477, Closes #1474, Closes #1478, Closes #1479, Closes #1476, Closes #1244, Closes #699


sethwv:
Closing until this is a little closer

PR #1494, opened 2026-07-28
Fix needs_reconnect handling in stream loop

### LINKED ISSUE #1493: [Bug]: needs_reconnect flag is not honored during _process_stream_data()
### Describe the bug

The stream health monitor can set `self.needs_reconnect = True` when a previously stable stream stops producing data, but `_process_stream_data()` does not currently check that flag.

As a result, `run()` can remain blocked inside `_process_stream_data()` and never reach the existing reconnect handling in the main loop.

I encountered this with an HLS livestream that became unavailable after a long period of successful playback. The health monitor correctly logged that it was setting the reconnect flag, but the active stream did not reconnect.

I tested a local patch that causes `_process_stream_data()` to yield when `needs_reconnect` is set. The existing reconnect path then successfully re-requested the configured stream URL and preserved the client playback session.

1. Configure a stream that plays successfully long enough to be considered stable.
2. Begin playback through Dispatcharr.
3. Cause the upstream stream to stop providing usable data while leaving the active read path in a state where the normal stream-processing loop remains running.
4. Wait for the health monitor to detect that no data has been received.
5. Observe the log indicating that the reconnect flag has been set.
6. Observe that the active _process_stream_data() loop does not exit in response to needs_reconnect, so the reconnect handling in run() is not reached.

Expected behavior: setting needs_reconnect should allow the active stream-processing loop to yield so the existing reconnect path can reconnect the current URL.

Note: Only one stream URL was configured for the channel. No alternate/backup Dispatcharr stream was available.

### Dispatcharr Version

v.0.27.2

### Docker Version

29.5.3, build d1c06ef

dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr@sha256:f2b5ea3128b0513feb07efee7539ad546f3ba3a9095f19936f446f75b8baa894
    container_name: dispatcharr
    ports:
      - "9191:9191"
    volumes:
      - ./dispatcharr:/data
      - ./dispatcharr-patches/manager.py:/app/apps/proxy/live_proxy/input/manager.py:ro
    restart: unless-stopped

### Client Information

Plex Version 1.43.3.10828

UppercaseTyler:
## Description

`_process_stream_data()` does not currently exit when the health monitor sets `needs_reconnect`.

This can leave control inside the stream processing loop even though the main stream loop has existing handling for `needs_reconnect` and `_attempt_reconnect()`.

This change adds `needs_reconnect` to the `_process_stream_data()` loop conditions so the function yields back to the existing reconnect path when a health-monitor reconnect is requested.

## Related Issue

Closes #1493

## How was it tested?

I reproduced the issue with an HLS livestream that became unavailable after a period of stable playback.

The health monitor correctly set `needs_reconnect`, but stream processing continued instead of yielding to the reconnect path.

I tested the equivalent loop-exit behavior in my running Docker deployment. Once `needs_reconnect` caused the stream processing loop to exit, Dispatcharr reconnected the configured stream URL and preserved the client playback session.

## Checklist

### DIFF
--- a/apps/proxy/live_proxy/input/manager.py
+++ b/apps/proxy/live_proxy/input/manager.py
@@ -1296,7 +1296,7 @@ class StreamManager:
         try:
             # Both transcode and HTTP now use the same subprocess/socket approach
             # This gives us perfect control: check flags between chunks, timeout just returns False
-            while self.running and self.connected and not self.stop_requested and not self.needs_stream_switch:
+            while self.running and self.connected and not self.stop_requested and not self.needs_stream_switch and not self.needs_reconnect:
                 if self.fetch_chunk():
                     self.last_data_time = time.time()
                 else:


SergeantPanda:
Thank you for the PR.

I looked in to this fix and it doesn't cover all paths quite right so I went ahead and fixed it. You were on the right track though, just a couple of missing pieces so failures were tracked correctly.

UppercaseTyler:
Thanks for looking into it!

PR #1500, opened 2026-07-31
fix(vod): preserve Xtream category-specific movie and series sources

### LINKED ISSUE #1443: [Bug]: VOD category selection is lost for duplicate movies and series
### Describe the bug

When the same movie or series exists in multiple enabled Xtream VOD
categories, Dispatcharr merges it into one canonical Movie/Series record.

The category/provider relation selected in the web interface or by an
external Xtream client is then lost. Episode lists and playback may come
from another category, usually whichever relation has the highest priority.

The Xtream output also returns the item only once or uses the same canonical
ID/name for all category entries.

1. Add an Xtream Codes account.
2. Enable two or more VOD categories containing the same series with the
   same TMDB ID.
3. Use categories with different source names or episode inventories.
4. Open the series in Dispatcharr and switch between its stream sources.
5. Connect an external player such as UHF to Dispatcharr using Xtream Codes.
6. Search for and play the series from the different categories.

### Dispatcharr Version

0.27.2

### Docker Version

29.5.0

services:
  dispatcharr:
    container_name: dispatcharr
    volumes:
      - ${DOCKER_DATA_DIR}/dispatcharr/data:/data
    environment:
      - PUID=${PUID}
      - PGID=${PGID}
      - TZ=${TZ:-Etc/UTC}
      - DISPATCHARR_ENV=${DISPATCHARR_ENV}
      - REDIS_HOST=${REDIS_HOST}
      - CELERY_BROKER_URL=${CELERY_BROKER_URL}
      - DISPATCHARR_LOG_LEVEL=${DISPATCHARR_LOG_LEVEL}
    labels:
    # Auto updates
      - com.getarcaneapp.arcane.updater=true
    # Dependencies
      - com.getarcaneapp.arcane.depends-on=gluetun
    network_mode: service:gluetun
    security_opt:
      - no-new-privileges:true
    depends_on:
      - gluetun
    restart: unless-stopped

### Client Information

Devices: iPhone, iPad, Mac, Apple TV 
App: UHF

Movhika:
## Description

Preserves the selected Xtream account/category relation when the same movie or
series is imported from multiple enabled VOD categories.

The change:

- exposes one Xtream entry per canonical movie/series and category relation
- uses relation-specific IDs for Xtream list, information and playback requests
- resolves playback against the exact selected account, category and upstream stream
- keeps shared TMDB/IMDB metadata while preserving provider-specific titles
- shows the M3U account and category in the series source selector
- refreshes the episode inventory when another series source is selected
- reports the selected account, category and stream ID in VOD connection statistics

No database models, migrations, dependencies or new API endpoints are introduced.

Changes are done with the support of AI (ChatGPT 5.6 Sol)

## Related Issue

Closes #1443

## How was it tested?

- Tested one canonical series present in three enabled Xtream categories.
- Verified that the Dispatcharr series modal shows all three account/category sources.
- Verified that switching sources in a select series or movie in the VOD Menu changes the provider-specific episode inventory.
- Verified that playback uses the stream ID belonging to the selected category.
- Verified that external player (in my case UHF) receives separate category entries with their provider-specific names.
- Verified movie and series playback through external player (UHF).
- Verified the selected M3U account, category and stream ID in VOD connection statistics.

Automated checks:

- Targeted frontend tests: 171 passed.
- Full frontend Vitest suite completed successfully.
- Vite production build completed successfully.
- ESLint passed for all changed frontend files.
- Prettier passed for the changed production and support files.
- Targeted Django regression suite: 40 passed, 2 PostgreSQL-specific tests skipped under SQLite.
- Python syntax and Git whitespace checks passed.

The current `dev` branch has existing repository-wide ESLint errors and formatting
drift in `SeriesModalUtils.test.js`; unrelated whole-file formatting changes were
deliberately excluded from this PR.

## Checklist

### DIFF
--- a/apps/output/tests/test_views.py
+++ b/apps/output/tests/test_views.py
@@ -14,6 +14,7 @@ from apps.output.views import (
     xc_get_live_streams,
     xc_get_series,
     xc_get_series_info,
+    xc_get_vod_info,
     xc_get_vod_streams,
 )
 from apps.vod.models import (
@@ -538,7 +539,7 @@ class XcVodSeriesDistinctTests(TestCase):
             stream_id="low-stream",
             container_extension="mkv",
         )
-        M3UMovieRelation.objects.create(
+        high_relation = M3UMovieRelation.objects.create(
             m3u_account=high,
             movie=movie,
             stream_id="high-stream",
@@ -549,7 +550,85 @@ class XcVodSeriesDistinctTests(TestCase):
 
         self.assertEqual(len(streams), 1)
         self.assertEqual(streams[0]["name"], "Shared Movie")
+        self.assertEqual(streams[0]["stream_id"], high_relation.id)
         self.assertEqual(streams[0]["container_extension"], "mp4")
+
+    def test_vod_global_list_keeps_one_item_per_category(self):
+        account = self._account(f"acct-{uuid4().hex[:6]}", priority=5)
+        categories = [
+            VODCategory.objects.create(name=name, category_type="movie")
+            for name in ("NETFLIX FILME", "GERMAN FILME", "KIDS FILME")
+        ]
+        movie = Movie.objects.create(
+            name="Shared Movie",
+            year=2025,
+            tmdb_id="12345",
+        )
+        relations = [
+            M3UMovieRelation.objects.create(
+                m3u_account=account,
+                movie=movie,
+                category=category,
+                stream_id=f"movie-{category.id}",
+            )
+            for category in categories
+        ]
+
+        results = xc_get_vod_streams(self.request, self.user)
+
+        self.assertEqual(len(results), 3)
+        self.assertEqual(
+            {row["stream_id"] for row in results},
+            {relation.id for relation in relations},
+        )
+        self.assertEqual(
+            {row["category_id"] for row in results},
+            {str(category.id) for category in categories},
+        )
+        self.assertTrue(all(row["tmdb_id"] == "12345" for row in results))
+
+    def test_vod_info_keeps_selected_category_relation(self):
+        account = self._account(f"acct-{uuid4().hex[:6]}", priority=5)
+        netflix = VODCategory.objects.create(
+            name="NETFLIX FILME", category_type="movie"
+        )
+        german = VODCategory.objects.create(
+            name="GERMAN FILME", category_type="movie"
+        )
+        movie = Movie.objects.create(
+            name="Canonical Movie",
+            custom_properties={"director": "Director"},
+        )
+        M3UMovieRelation.objects.create(
+            m3u_account=account,
+            movie=movie,
+            category=netflix,
+            stream_id="netflix-movie",
+            container_extension="mp4",
+            custom_properties={"detailed_info": {"name": "Netflix Movie"}},
+            last_advanced_refresh=timezone.now(),
+        )
+        german_relation = M3UMovieRelation.objects.create(
+            m3u_account=account,
+            movie=movie,
+            category=german,
+            stream_id="german-movie",
+            container_extension="mkv",
+            custom_properties={"detailed_info": {"name": "German Movie"}},
+            last_advanced_refresh=timezone.now(),
+        )
+
+        info = xc_get_vod_info(
+            self.request,
+            self.user,
+            german_relation.id,
+        )
+
+        self.assertEqual(info["info"]["name"], "German Movie")
+        self.assertEqual(info["movie_data"]["stream_id"], german_relation.id)
+        self.assertEqual(info["movie_data"]["category_id"], str(german.id))
+        self.assertEqual(info["movie_data"]["category_ids"], [german.id])
+        self.assertEqual(info["movie_data"]["container_extension"], "mkv")
 
     def test_vod_streams_excludes_inactive_accounts(self):
         active = self._account(f"active-{uuid4().hex[:6]}", priority=1)
@@ -681,6 +760,47 @@ class XcVodSeriesDistinctTests(TestCase):
         self.assertEqual(len(results), 1)
         self.assertEqual(results[0]["name"], "Shared Series")
         self.assertEqual(results[0]["series_id"], high_rel.id)
+
+    def test_series_global_list_keeps_one_item_per_category(self):
+        account = self._account(f"acct-{uuid4().hex[:6]}", priority=5)
+        categories = [
+            VODCategory.objects.create(
+                name=name,
+                category_type="series",
+            )
+            for name in (
+                "NETFLIX ANIME",
+                "GERMANY KINDER",
+                "NICKELODEON",
+            )
+        ]
+        series = Series.objects.create(
+            name="DE - Avatar: Der Herr der Elemente (US)",
+            year=2005,
+            tmdb_id="246",
+        )
+        relations = [
+            M3USeriesRelation.objects.create(
+                m3u_account=account,
+                series=series,
+                category=category,
+                external_series_id=f"avatar-{category.id}",
+            )
+            for category in categories
+        ]
+
+        results = xc_get_series(self.request, self.user)
+
+        self.assertEqual(len(results), 3)
+        self.assertEqual(
+            {row["series_id"] for row in results},
+            {relation.id for relation in relations},
+        )
+        self.assertEqual(
+            {row["category_id"] for row in results},
+            {str(category.id) for category in categories},
+        )
+        self.assertTrue(all(row["tmdb_id"] == "246" for row in results))
 
     def test_series_category_selection_keeps_relation_specific_episodes(self):
         account = self._account(f"acct-{uuid4().hex[:6]}", priority=5)
@@ -891,7 +1011,7 @@ class XcVodSeriesRegressionTests(TestCase):
     def test_vod_streams_response_keys(self):
         account = self._account(f"acct-{uuid4().hex[:6]}")
         movie = Movie.objects.create(name="Schema Movie", rating="10")
-        M3UMovieRelation.objects.create(
+        relation = M3UMovieRelation.objects.create(
             m3u_account=account, movie=movie, stream_id="schema-1"
         )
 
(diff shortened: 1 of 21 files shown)


SergeantPanda:
I had just made a lot of changes to the VOD code which I believe probably caused some merge conflicts here.

I'm assuming that is why this diff looks larger than it probably needs to be.

Please rebase to the current dev.

mehmetbayik:
In `apps/vod/tasks.py`, Line 485, there is a condition that:

```bash
#Skip duplicates in this batch
if movie_key in movie_keys:
    continue
```

I think **this may cause** skipping multiple versions of same movie if they appear to be in same 1000-movie batch.

I am trying to customize VOD part of Dispatcharr because raspberry pi freezes when synchronizes my library of 36k VOD content. I already edited VOD2MLIB plugin to use with jellyfin, now I am trying to do better/faster VOD synchronization. Realized that current implementation only takes one version of same movie/series  and writes as stream-id but I have different language/resolution versions of same movies/series. I saw your PR that suits most of my needs. 

**Please let me know** if this helps you or if I am wrong about duplicates.

_Note that this code snippet also appears in main branch so I wanted to ask you why didn't you removed this part since your PR emphasises source selection for movies/series specifically._

I also opened an issue related to this deduplication condition: #1511

Movhika:
> I had just made a lot of changes to the VOD code which I believe probably caused some merge conflicts here.
> 
> I'm assuming that is why this diff looks larger than it probably needs to be.
> 
> Please rebase to the current dev.

Thanks — I rebased the branch onto the current dev and force-pushed the updated branch.
The diff is still relatively large after the rebase, but a significant part of it is regression coverage: 973 of the 1,662 added lines are tests. The affected frontend tests and 21 PR-specific backend regression tests pass.
However, while running the newer dedicated XC test modules from the current dev, I found four existing failures in 

XCGetSeriesInfoTests:
test_returns_all_active_provider_episodes
test_prefers_higher_priority_account_stream_metadata
test_episode_artwork_prefers_higher_priority_relation
test_does_not_n_plus_one_episode_relations

The first three failures are caused by the existing tests explicitly expecting the old unified behavior: episodes and metadata from all provider relations are combined, and the highest-priority relation is preferred even when a specific series-relation ID was requested.
The fourth test currently expects exactly one episode-relation query, while the category-specific implementation performs two constant queries. This is not an N+1 query pattern, but I will check whether the additional query can be avoided before changing that expectation.

After reviewing the interaction with #1398, I believe the selected category/relation should remain authoritative for listing, series information and playback, including failover.
Relations for the same canonical movie or series are not necessarily interchangeable. Different categories may contain different languages, resolutions, episode titles, season availability, or otherwise different versions of the same content. Matching TMDB metadata identifies the underlying title, but it does not guarantee equivalent streams.
For example, selecting NICKELODEON should not silently fail over to NETFLIX ANIME or GERMANY KINDER.

My proposed behavior is:
Xtream listings expose one entry per category relation.
get_series_info returns episodes belonging to the requested relation.
Playback starts with the exact selected movie/episode relation.
The failover introduced in #1398 is restricted to suitable relations within the selected category scope. If no suitable relation is available, Dispatcharr reports the stream as unavailable instead of silently switching language or quality.
If this is the intended behavior, I will update the four existing test expectations and move the two relation-specific tests into the dedicated test_xc_vod_info.py and test_xc_series_info.py modules.

Movhika:
> In `apps/vod/tasks.py`, Line 485, there is a condition that:
> 
> ```shell
> #Skip duplicates in this batch
> if movie_key in movie_keys:
>     continue
> ```
> 
> I think **this may cause** skipping multiple versions of same movie if they appear to be in same 1000-movie batch.
> 
> I am trying to customize VOD part of Dispatcharr because raspberry pi freezes when synchronizes my library of 36k VOD content. I already edited VOD2MLIB plugin to use with jellyfin, now I am trying to do better/faster VOD synchronization. Realized that current implementation only takes one version of same movie/series and writes as stream-id but I have different language/resolution versions of same movies/series. I saw your PR that suits most of my needs.
> 
> **Please let me know** if this helps you or if I am wrong about duplicates.
> 
> _Note that this code snippet also appears in main branch so I wanted to ask you why didn't you removed this part since your PR emphasises source selection for movies/series specifically._
> 
> I also opened an issue related to this deduplication condition: #1511

Thanks, this is helpful

The current batch deduplication uses the canonical movie/series metadata key. When another version of the same title appears in the same 1,000-item batch, the continue prevents its M3UMovieRelation or M3USeriesRelation from being created.

This PR currently fixes the downstream handling of category-specific relations in the XC output, series information and playback, but it assumes that every provider relation was successfully imported. Your finding identifies an earlier ingestion problem, and it also means that the failover introduced in #1398 may receive an incomplete candidate list.
Simply removing the continue is probably not sufficient because the same dictionary is also used to deduplicate the canonical Movie or Series object. The proper solution would be to deduplicate the canonical content while separately retaining every source relation by stream_id or external_series_id.

Since this is now tracked in #1511 and this PR is already relatively large, I would like the maintainers to confirm whether they want that ingestion fix included here or handled in a focused follow-up PR.

Movhika:
The issue #1511 also causes #1398’s failover candidate list to be incomplete.
Simply removing the two continue statements would not be sufficient because the same dictionaries are also used to deduplicate the canonical Movie and Series objects. 

A proper fix should:
1. Continue creating only one canonical Movie or Series for matching metadata.
2. Collect every provider entry separately.
3. Create or update every relation using its stream_id or external_series_id.
4. Preserve the category and provider-specific metadata on each relation.

This should require changes only to process_movie_batch() and process_series_batch(), plus focused regression tests. No model change or database migration should be necessary. Previously skipped relations would be restored during the next VOD synchronization.

Movhika:
Rebased onto the current dev branch and resolved the VOD conflicts against the new redirect/failover implementation. I also reran the affected backend and frontend regression tests successfully.

@SergeantPanda 
Still need to know if #1511 should be fixed here too or in another PR?

SergeantPanda:
Thanks for the detailed work on this, and for putting together thorough testing notes. We really appreciate the effort you've put into diagnosing and addressing #1443.

After review, we're going to decline this PR for now. The Xtream Codes output changes (exposing one entry per movie/series per category, with relation-specific IDs, names, and playback) go against behavior we deliberately designed. Dispatcharr currently deduplicates to one row per canonical title for XC clients, with the highest-priority account winning. That was an intentional tradeoff for library size, performance, and simplicity until we have better VOD source management in place. Expanding XC output to surface every category variant as a separate item is not something we want to pursue at this time.

We do agree there is a real gap in the web UI, separate from the XC export question. When switching series sources, the frontend correctly sends a `relation_id`, but the backend episode list is scoped by M3U account and canonical series rather than the selected relation. If the same account carries multiple upstream versions of the same show (for example, different provider IDs merged onto one Series), their episode inventories can be merged in the UI even after selecting a specific source. That is a known limitation we plan to address as part of broader VOD source management, rather than by changing the XC export contract.

Showing the category name alongside the account in the source dropdown could help disambiguate entries, but it would not fix the episode scoping issue on its own. If you are interested in contributing a targeted fix for relation-scoped episode loading in the web UI (without the XC output changes), we would be open to reviewing that separately.

Thanks again for contributing. Hope you'll keep sending issues and PRs our way.

northernpowerhouse:
@SergeantPanda, this would also fix an issue I'm running into, where my provider's VOD library has about 5 different languages, sorted by category. Because dispatcharr is merging every item with the same external id to one row, I can't chose which language and dispatcharr chooses one at random.

I understand not wanting to have one VOD library item per TMDBid per category, but if we could implement VoD category groups (like the channel profiles configurable from the UI by either regex or a category picker like the m3u group picker), we could: have a VOD library item by category group (in my case, per language, which is rational) and scope VOD access per user per category group.

We could also set priority within the group, for example we could prioritise a 4K category, or deprioritise a dubbed category, to help steer the TMDB to stream selection path, which at the moment is pretty random. 

Obviously this is a bigger change than I've contributed in the past since we're adding a class, so I wanted to see how you feel about the direction before I get going.

Movhika:
Hey @northernpowerhouse 
I think your proposal overlaps quite heavily with the broader VOD management work I’m currently building. Before we create two separate implementations, could you take a look at [#1610]
(https://github.com/Dispatcharr/Dispatcharr/issues/1610) and [my fork](https://github.com/Movhika/Dispatcharr)?
The implementation is already usable in my own setup and is used daily by friends and family. 
I also keeping this up to date with the changes from here! so currently its based of 0.30

It currently includes:
- Per-user VOD output profiles
- Dynamic source and category selection using pickers and import rules
- A Compact mode that groups sources under canonical titles and uses configurable ranking and failover rules
- Edition/suffix rules that can expose separate entries such as a standard and a 4K version of the same title
- A Variants mode that exposes every allowed provider source individually, without failover
- TMDB enrichment and configurable metadata/title output

Your language-category use case should already be covered by a Compact profile: select the relevant categories and then prioritize or exclude sources according to language, resolution, features, or provider. Multiple profiles can also be assigned to different users.
I’m still refining and testing the implementation, but it should not be much longer before it is ready. Once it is, I’d like to submit it upstream if the maintainers are open to it.
I’d be happy to compare notes or collaborate so we avoid duplicating the same work.

image
image

northernpowerhouse:
> Hey @northernpowerhouse 
> I think your proposal overlaps quite heavily with the broader VOD management work I’m currently building. Before we create two separate implementations, could you take a look at [#1610]
> (https://github.com/Dispatcharr/Dispatcharr/issues/1610) and [my fork](https://github.com/Movhika/Dispatcharr)?
> The implementation is already usable in my own setup and is used daily by friends and family. 
> I also keeping this up to date with the changes from here! so currently its based of 0.30
> 
> It currently includes:
> - Per-user VOD output profiles
> - Dynamic source and category selection using pickers and import rules
> - A Compact mode that groups sources under canonical titles and uses configurable ranking and failover rules
> - Edition/suffix rules that can expose separate entries such as a standard and a 4K version of the same title
> - A Variants mode that exposes every allowed provider source individually, without failover
> - TMDB enrichment and configurable metadata/title output
> 
> Your language-category use case should already be covered by a Compact profile: select the relevant categories and then prioritize or exclude sources according to language, resolution, features, or provider. Multiple profiles can also be assigned to different users.
> I’m still refining and testing the implementation, but it should not be much longer before it is ready. Once it is, I’d like to submit it upstream if the maintainers are open to it.
> I’d be happy to compare notes or collaborate so we avoid duplicating the same work.


That's funny, I came to this PR request from your issue in the first place. I'm just trying to solve the sgt's specific objection re: the impact on output m3u sizes. 

If upstream is happy with the concept of your fork in terms of a design direction, I'm happy to contribute to it and do some testing.

PR #1508, opened 2026-08-03
Add "m3u_error" event type across backend and frontend components

### LINKED ISSUE #1507: [Feature]: Add m3u_error webhook event on M3U refresh failure

Currently, there are webhook events for `m3u_refresh` (which only fires on success) and `m3u_blocked`, but there is no system event or webhook triggered when an M3U account refresh fails (e.g., HTTP errors, timeouts, parsing failures, XC auth errors).

When an M3U refresh fails, the account status changes to `ERROR`, but no webhook or system event is logged. 

It would be useful to have an `m3u_error` event so webhooks and custom scripts can notify users when a playlist refresh fails.

### Proposed Solution

Add a new `m3u_error` event type to Connect webhooks and system events:

1. **Backend**: Add `m3u_error` to `SUPPORTED_EVENTS` (`apps/connect/models.py`) and `SystemEvent.EVENT_TYPES` (`core/models.py`).
2. **Tasks**: Trigger `log_system_event(event_type="m3u_error", ...)` in `apps/m3u/tasks.py` when an account refresh sets an error status or fails.
3. **Frontend**: Add `m3u_error` to `SYSTEM_EVENT_LABELS` and update `SystemEvents.jsx` to render the red error badge and icon.

### Additional context

Nothing Extra

gianlucalauro:
## Description

Adds a new `m3u_error` system event type that fires whenever an M3U account refresh fails (due to network errors, HTTP failure, parsing errors, or XC auth errors). This enables users to subscribe to `m3u_error` in Connect (webhooks/custom scripts) to get notified when a provider playlist fails to update.

## Related Issue

Closes #1507

## How was it tested?

Added unit tests in `apps/m3u/tests/test_m3u_error_event.py` verifying that `_set_m3u_account_status` (with `notify_error=True`) and `_ensure_m3u_refresh_terminal_status` invoke `log_system_event` with `event_type="m3u_error"`.

## Checklist


SergeantPanda (changes requested):
Hey, thanks for this. The overall approach looks good: hooking into `_set_m3u_account_status(notify_error=True)` and the terminal-status fallback is the right place for full refreshes, and the models/migrations/frontend wiring all line up.

Before merging, please address these:

1. **Account name lookup.** Drop the extra `.only("name").first()` in both helpers. Call sites in `_refresh_single_m3u_account_impl` already have the account loaded, so pass `account_name` in as an optional kwarg. For `_ensure_m3u_refresh_terminal_status`, you’re already querying status: pull `name` in that same query with `.values("status", "name")`. Fall back to `str(account_id)` when you don’t have a name. No second DB round-trip.

2. **Coverage.** This currently misses error paths that set `ERROR` outside those helpers. Download/XC failures inside `refresh_m3u_groups` / `fetch_m3u_lines` (including create flows) update status and send the WS update but never call `log_system_event`. Emit `m3u_error` for those paths too so Connect subscribers actually get notified whenever a refresh fails.

3. **Error message.** On a full refresh, when download fails inside `refresh_m3u_groups`, the outer path often overwrites the detailed `last_message` with the generic “Failed to refresh M3U groups…” string, and that’s what ends up in the webhook. Propagate the real error into the `m3u_error` payload.

4. **Add `epg_error` in this PR.** We don’t have one today (only success `epg_refresh` and blocked). The EPG helpers already mirror the M3U `notify_error` / ensure pattern, and doing this later means another AlterField migration on the same choice fields. Add `epg_error` the same way as `m3u_error`, including frontend/Connect wiring, with the same name-passing approach (no extra DB hit), and cover the equivalent EPG failure paths.

Happy to re-review after those are in. Thank you again!

SergeantPanda:
Were you still wanting to tackle this?

gianlucalauro:
I'm going to do this in 2 weeks 'cause i'm not at home in this moment

SergeantPanda:
Sounds good thank you for responding.

PR #1510, opened 2026-08-04
feat(logs): route the container's output through a log collector [logging 1]

nagelm:
## Description

Dispatcharr's container emits log lines from eight producers — python logging in uwsgi, celery, daphne and management commands, plus uwsgi itself, nginx, postgres, redis and the entrypoint shell — in eight formats, three clocks, and with no shared notion of severity. `docker logs` is the only place they meet, and nothing there can be filtered, correlated or kept.

This PR puts one process in front of all of it.

`dispatcharr/log_collector.py` is a standalone Python process with no Django and no gevent. The entrypoint saves the real stdout and execs the rest of the boot through it, so every process that follows writes into its stdin. For each line it decides the source, the severity and the clock, rewrites the record into one grammar, forwards it to the real stdout so `docker logs` stays complete, and files the same bytes.

```
2026-08-24 13:33:26,355 +1200 INFO postgres [217] LOG:  checkpoint complete: wrote 2 buffers
2026-08-24 13:33:41,088 +1200 WARNING core.models CoreSettings group cache get failed; falling back to Postgres
2026-08-24 13:33:44,201 +1200 INFO uwsgi.requests Worker ID: 3 GET 404 /player_api.php 26ms
```

- **One grammar:** `stamp offset LEVEL source message`. Continuations — traceback tails, postgres `DETAIL:`/`HINT:`/`STATEMENT:` trailers — stay attached to their record and move with it.
- **One clock.** Every source is converted to the operator's display zone, whether it stamps local time or UTC. nginx's error log now goes to stderr so it reaches the collector; its access log is left alone, since uwsgi already logs each request.
- **Real severities**, derived from each source's own vocabulary: postgres `FATAL` becomes `CRITICAL`, redis `#` becomes `WARNING`, nginx `crit` becomes `CRITICAL`. Sources that state no severity are marked as such, so the level floor cannot delete a line whose level this process invented.
- **A level floor** applied to both sinks at once, changeable from **Settings → System Settings** without restarting anything.
- **Persistence, rotation and retention**, owned by the writer thread that holds the descriptor, all changeable at runtime. Persistence can be switched off entirely, leaving `docker logs` untouched.

### What this replaces, and why

The first version of this PR added a `WatchedFileHandler` to Django's logging config and rotated it from a Celery task. @SergeantPanda's review objected that this meant every application process performing per-line file I/O on top of the writes it already does to stdout, on systems where that I/O is expensive. That objection was correct, and this design answers it directly:

- **No application process performs log file I/O at all.** Django's logging remains console-only — the `handlers` block is unchanged from `dev`. The only writer is the collector.
- **Writes are batched**, 128 KB or 250 ms, by a thread whose only job is writing.
- **Rotation is not a scheduled task.** It happens in the thread holding the descriptor, which closes the rename race a Celery task had and bounds overshoot to one batch. The old beat entry is left registered but disabled so `DatabaseScheduler` stops firing it.
- **The stream is never held up by the file.** A dead `/data` degrades to drain-and-drop; a write failure requeues only the unwritten tail; the buffer is bounded and records its drops.
- **`docker logs` outlives any collector fault.** The collector is supervised: restarted if it dies, and after three rapid failures replaced by a plain `cat` passthrough rather than a restart loop that drops the stream.

### Merge order

This is the base of a five-part series. **[logging 2] (the log file browser) requires this PR** — `core/log_files.py` imports from `dispatcharr.log_collector` at module scope, reached from the URLconf, so without this PR every URL raises `ModuleNotFoundError`. [logging 3], [logging 4] and [logging 5] also build on this one and do not depend on each other.

A normal merge cannot land them out of order, since each branch has this one as a strict ancestor. The hazard worth stating: **reverting this PR after [logging 2] has landed is textually clean, conflicts with nothing, and takes the site down.** Revert [logging 2] first.

Modular deployments collect too: the web and celery containers each run a collector, writing `dispatcharr.log` and `dispatcharr.log-celery` so one shared `/data` never has two writers. That covers uwsgi, daphne, celery and application logging; postgres and redis run in their own containers and stay out of reach.

## Related Issue

Closes #

## How was it tested?

Built the image from this branch and booted a container against a fresh volume (`TZ=Pacific/Auckland`).

**Automated suites**
- Backend, in the container: `python manage.py test` — **1967 tests, OK.**
- Frontend: `npx vitest run` — **6218 tests / 203 files pass.**
- Collector supervision: `bash docker/tests/test-log-collector-supervision.sh` — **17 assertions pass.** These cover the boot archive shift and the restart, degradation and counter-reset paths without building an image, by stubbing the collector invocation.

**Behaviour, against the running container**
- Every line of a 533-line first-boot log is either a stamped record or an indented continuation — **zero non-conforming lines**, checked over the whole file.
- 22 distinct source tokens appear, including `postgres`, `redis`, `uwsgi`, `daphne.server` and `stdout`.
- Timestamps carry `+1200` from the first line; postgres `FATAL` renders as `CRITICAL`; an `ERROR` keeps its `STATEMENT:` trailer attached.
- Changing size and retention in System Settings rewrote `collector.conf` and took effect **without a restart**; the file rotated at the new cap and held exactly the configured number of archives under 52,000 lines of load.
- Raising the level floor to `WARNING` dropped an injected `INFO` from **both** sinks and kept the `WARNING` in both.
- Switching persistence off froze the file at the byte while `docker logs` kept receiving; switching it back on resumed the file. Both live.
- Restarting the container archived the previous run's file and started a clean one.
- Killing the collector produced `log collector exited abnormally; restarting`, it came back, and the file kept receiving.
- An injected traceback kept its tail attached as indented continuations.

**Modular mode**, brought up from this repo's compose file: both containers collect into their own file, a settings save in the web UI reaches both collectors, and restarting one leaves the other's log untouched.

## Checklist

**No migration is required.** The only change to `core/models.py` is four keys added inside `CoreSettings.get_system_settings()`; the autodetector compares model state, not method bodies. Existing installs will not have the keys stored, and every consumer supplies its own fallback, so no data migration is needed either.

**Two `print()` calls remain, both deliberate:** the `python -m` usage message in `log_collector.py`, and `startup_log.py`, where printing *is* the mechanism — it emits canonical lines before Django configures logging. This PR is a net reduction: it replaces seven pre-existing bare `print()`s in `settings.py`, `celery.py` and `gevent_patch.py`.

**ESLint and Prettier** are verified as "this PR introduces no new problems": eslint reports 161 problems / 106 errors on this branch and the identical 161 / 106 on stock `dev`, measured by extracting all 428 committed frontend files from `dev` and running the same binary against them. Prettier is clean on every file this PR adds, tested against the committed blobs rather than the working tree.


nagelm:
@sethwv & @SergeantPanda - any feedback on this would be appreciated

sethwv (changes requested):
@nagelm Apologies for the delay in review - I really like how this is looking, new routes are well documented and follow existing patterns with appropriate auth and security considerations. I really only have two notes/change requests-
1. It would be useful to have the ability to reverse the log direction in the UI log-viewer. Something like a checkbox to persist a preference in local storage maybe?
2. Rather than an auto-refresh checkbox, we should consider following the 'pattern' established in the stats page log viewer where the auto-refresh is controlled by a dropdown with a selectable interval. image

nagelm:
I applied the autorefresh shape you mentioned. I also changed the buttons to the subtle button primitive (there are a lot of divergent button styles so let me know if you want something else).

For the ordering, went with drop down options "newest first" and "newest last". This was because the log display will only ever tail max 5MB (making the display relative to 'newest' - hopefully that makes). If you had somethings else in mind let me know and I can rework the window mechanism.

(also updated to current dev)

sethwv (approved):
Looks a lot better, thanks! Happy to give this one a new full pass once you unmark as draft.

nagelm:
I have also been working on a redaction change so that secrets, provider URLs, credentials etc. are automatically removed or masked. The sequencing issue I have is that I built the redaction on top of the log viewer (technically I could split it up but its easier if 1446 (log timestamp fix) & 1510 (log viewer) land first since they add a lot of code in the same locations). 

Just wanted to flag it it because once the log viewer lands in app you could have an increase in people accidentally disclosing stuff in their support requests if it is not auto-redacted by the download function. 

One option is I could put in a stacked PR (i.e. redaction PR stacked on the log viewer PR) or another option is I could rebase all 3 log related changes into a logging improvements branch (log timestamp fix; log viewer; log redactions) and then create a single PR to DEV. 

Let me know if you have any preference.

sethwv:
@nagelm I think my preference is that #1510 land first and 'separately' rather than rolling multiple similarly scoped things into one PR.

Let's give the stacked PR a try and we'll get them in sequence. I know there is apetite for both. Haven't done anything with Github's new stacked PR implementation yet, so will need a sec to get my bearings.

sethwv:
SergeantPanda is unplugged on vacation this week, since he has been reviewing #1446 this may all need to wait to land until his return.

nagelm:
Updated now that #1446 has landed.

One new commit, test(logs): isolate the settings coercion test from the shared cache. The test saves a real CoreSettings row, which invalidates the Redis-backed settings group cache, and that is not rolled back with the test transaction. CatchupRedirectViewTests is a SimpleTestCase that patches is_default_stream_profile_redirect but not is_catchup_enabled, so it only passes while that group stays cached; cold, it queries the database and trips DatabaseOperationForbidden.

This predates the branch (single-process full-suite runs: 7931c3bd (pre-#1446) 6 failures + 1 error, a1cbb6bb (the #1446 merge) 1 error, f7d87b4d 1 error). refresh_display_zone() reads the system settings group on request_started, keeping it warm. My write lands after the last warm-up and restores the older behaviour. The test now uses a locmem cache, the pattern already in apps/timeshift/tests/test_views.py, so it leaves the shared cache alone. CI shards by label, so none of this ever reached it. #1549 is a tiny patch to add an is_catchup_enabled into that test so this is fixed rather than masked.

Two pre-existing dev failures, both reproduced on a clean checkout:

CatchupProxyTests.test_missing_session_id_redirects errors in full-suite runs (the connection is closed); present at all three commits above.
Logo.test.jsx, 31 failures — 238fe759 added LOGO_NAME_MAX_LENGTH to LogoUtils.js but that test's vi.mock was not updated. This is why the frontend job is red here, and it will be red on anything targeting dev.
Verified after the rebase: full backend suite 1849 tests with no failures beyond the one above (clean dev run the same way: 1823, same single error); all four /api/core/logs/... paths still in the schema; frontend 6126 passing; ESLint 0 introduced.

nagelm:
Updated onto clean green dev head

nagelm:
@sethwv & @SergeantPanda let me know if there are any further changes you want here with this viewer. 

Once this lands on dev I can rebase the redaction changes and take them out of draft:

#1550 has the core redaction changes and write side pattern matching (a low chance of regression)
#1551 goes after the log invocation sites which generate hard-to-match patterns (higher surface touching other code)

SergeantPanda (changes requested):
Thank you for the PR.

A built-in log viewer is something we have wanted for a while but have not had time to implement, so I appreciate you taking this on.

I do have performance concerns. Merging this as-is means dual logging: we already write to stdout for Docker, and this also appends every line to a file. That extra sync write sits on the request/stream path. On NVMe it is probably fine; on SMB, SD cards, and Pi-class hosts it can add latency or jitter, especially with DEBUG left on (which a lot of people do). I am not interested in landing features that tax those setups unless we have to.

If we go ahead, file writing needs to be designed for slow storage:

- Batch in ~64–128 KB (or a short time window) and flush from a **background OS thread**, not on the calling greenlet. A naive async `QueueListener` is not enough here because uWSGI is gevent-patched and can still stall the worker on disk I/O.
- Keep stdout line-oriented so `docker logs` stays complete. Losing one unflushed file batch on a crash is acceptable.
- Make file persistence **disableable** (a System Setting is fine), default **on**, so people on slow volumes can turn it off without losing stdout/`docker logs`.

If you are up for that challenge, great, if not, please let us know.

Thank you again!

SergeantPanda:
Hi @nagelm, is this ready for another review?

nagelm:
Hi @SergeantPanda I've made a the changes you wanted and redid aspects of 1446 so it's better but haven't finished manual testing and hardening (been busy at work this week). 

I still want to tokenize the module/category for filters and clean up some string formatting so traces or multiline logs look better.

Should be finished in the next day or two.

SergeantPanda:
Ok no problem at all. I had just seen commits pushed and wanted to verify. Take your time, I appreciate you taking the feedback and making the proper adjustments.

nagelm:
@SergeantPanda hopefully you don't mind me shuffling things around but the change was getting unmanageably large as a single PR. I've now made this one about the collector/log refactoring and the viewer itself moves to #1601 .

If/when you are happy with the approach I would love to get some feedback from 'beta' testers after #1510 & #1601 have merged.

| PR | title | comment |
| --- | --- | --- |
| #1510 | route the container's output through a log collector [logging 1] | logging refactor to create a single sink |
| #1601 | in-app log file browser [logging 2] | The log viewer changes which were previously bundled with 1510, split for readability |
| --- | --- | --- |
| #1602 | quieten the container's boot and runtime output [logging 3] | some housekeeping, there is more I/we could do here to tidy up the logs but I don't want to get too far ahead of my skiis until the above two land |
| --- | --- | --- |
| #1550 | Mask provider credentials in application log output [logging 4] | pattern matching based redaction of secrets, PII, tokens, usernames, passwords & provider URLs |
| #1551 | Mask provider context at URL-bearing log sites [logging 5] | going after the emission sites for patterns which are hard/impossible to match at collector ingest time |

SergeantPanda:
I have looked this over now, thank you for taking so much care in building this.

I haven't fully tested the full functionality out yet with the log viewer portion too, but, reading through the code overall the implementation looks solid and aligns with what I requested from a performance standpoint.

I have two concerns though.
1) Not supporting modular, I do understand that we can't capture ALL logs from modular deployments but I do think we might as well capture what we can and have it on for modular deployments (uWSGI, Celery, Daphne) this is where most useful logs will come from anyway
2) This is a LOT of code to maintain. Between this PR and the other I believe it is pushing 5k lines of code. I'm not saying there is unnecessary bloat or anything like that but we want to make sure we can easily maintain this code going forward (especially considering the type of feature).

That being said, we are still discussing this internally on the direction we want to go with it. We will play with it (hopefully tomorrow) to determine if it is lightweight enough to merge. In the meantime, getting it to work with modular would be nice as that would for sure be a requirement for this feature in our eyes.

Thanks again for all of your time, I can tell it was well thought out!

nagelm:
It would not be difficult to include the core aspects of a modular set up. The only reason I didn't is I dont have a real instance to test with (of course I can spin up a test bed but that is never the same as a real implementation). I'll add it over the next day or two and then yes please ask the team to have a play around. 

I am not precious about about the decisions so if you need to change anything back or front end go for it. It is a lot of code but hopefully it can help the project and community streamline the support process. (And there are a few nits I see from spending time with the logs which are not hard to resolve once reported.)

Either way, I will move onto to bug fixes and chores for a while after this.

nagelm:
@SergeantPanda modular collection added. Web and celery containers each run their own collector (dispatcharr.log and dispatcharr.log-celery on the shared /data), so now uwsgi, daphne, celery and all application logging are captured for modular. Postgres and redis stay out of reach in their own containers.

SergeantPanda (commented):
This is looking pretty good, thank you for the continued work.

A few small things to take care of:
1) line 252 in the entrypoint has a literal `\n` instead of a line break
2) `logs` should probably be added to the `DATA_DIRS` in `03-init-dispatcharr.sh`
3) What do you think about putting `collector.conf` and `collector.pid` in a sub director (like `./logs/config`)?

Thanks again!

nagelm:
All three done — line continuation fixed, logs is now in DATA_DIRS, and collector.conf/collector.pid have moved under logs/config/ as you suggested; the branch is also rebased onto the v0.30.0 release.

SergeantPanda (changes requested):
Re-reviewed at `2c4f8c6`. The earlier items look good and CI is green. Two things left from my side:
1. **Non-collector installs (Debian/Proxmox / bare metal).** Removing `display_timezone` and the request/task refresh path means app logs are UTC-only whenever nothing sits in front of stdout. On Docker that is fine because the collector applies the display zone. On installs without a collector, System Settings `time_zone` no longer affects log timestamps, which is a regression vs current `dev`. Can we keep a thin in-process display-zone path when no collector is present, or confirm you are intentionally limiting that behavior to collector deployments for now?
2. **`SIGALRM` → `_force_rotate`.** Nothing in the repo sends `SIGALRM` to the collector; tests set `_force_rotate` directly. Please drop the unused signal handler (or wire a real use for it) so we are not leaving dead control surface around.
Performance/design still looks right for the original dual-write concern. Happy to take another pass after these.

nagelm:
Item 1. I guess I rearchitected #5010 after #1446 merged so technically a regression is fair but ~2 weeks ago (less on main), bare-metal logs were UTC anyway. My intuition is AIO and to some extent modular are the modern, forward leaning install approaches with bare metal being a legacy paradigm. Norm is to follow the host TZ though and it's easy to resurrect the ~50 lines plus two receivers. Hopefully there are some maintainers/collaborators/engaged testers who run bare metal then the risk of maintaining divergent mechanics is low. Fixed.

Item 2. I changed tac when there was no obvious need and forgot to remove the unused handler, oops. Fixed.

nagelm:
Another update onto dev head (status was conflicting). Plus two new commits for bare metal (or any no collector installs)

#1510 (fix(logs): hide the log file settings where no collector can apply them): the environment payload now carries log_collector_running (same pidfile check the receivers use), the persist/size/retention fields hide when it's false (absent flag = older backend = show, matching the page's own convention), and the save warning says "process output" instead of "container output". The floor branch's log-level Select moved inside the same gate during its replay.

#1601 (fix(logs): leave the Logs page out of deployments with nothing to browse): the nav drops the Logs entry — including inside the System group, which took a filter that reaches group children (the first version didn't, and the new test caught it). The page stays routable for deep links and mid-session collector deaths, with the alert copy reworded to fit both worlds.

SergeantPanda:
> Item 1. I guess I rearchitected #5010 after #1446 merged so technically a regression is fair but ~2 weeks ago (less on main), bare-metal logs were UTC anyway. My intuition is AIO and to some extent modular are the modern, forward leaning install approaches with bare metal being a legacy paradigm. Norm is to follow the host TZ though and it's easy to resurrect the ~50 lines plus two receivers. Hopefully there are some maintainers/collaborators/engaged testers who run bare metal then the risk of maintaining divergent mechanics is low. Fixed.

Completely understandable. Dispatcharr has always been a Docker container, unfortunately we do still need to code for non Docker deployments as many users use the Debian install script as well as Proxmox helper scripts.

Agreed on hoping we have users that will work with us to find bugs on the bare metal installs, we can only do so much testing with all of the different deployment options (especially the ones we don't maintain).

> Item 2. I changed tac when there was no obvious need and forgot to remove the unused handler, oops. Fixed.

No problem at all! Just trying to get it as close to merge ready as possible, we are just getting to the small things!


I'll give this another check and confirm with the other devs here. I did run the branch and it worked well. Still need to find a good way to benchmark but overall this is solid. Thank you very much!

SergeantPanda (approved):
I've spent quite a bit of time benchmarking this and trying a few small collector-side tweaks. I'm satisfied it won't cause performance issues in practice.
The biggest wins were replacing `strftime`/`strptime` on the hot path with direct field access in `_parse_naive` and `_render`. I also reordered matchers so the common Python shape is hit early (`_PY` before `_UWSGI`), and kept `_PY_REQ` before `_PY` so uwsgi request lines get the right timezone semantics.
In an end-to-end test (same `logger.emit()` through stdout vs pipe+collector), I still see roughly **30% lower throughput** when CPU-bound. Standalone collector drain is much higher than that; the gap is mostly pipe/coupling overhead, not normalization CPU.
Even with that hit, we're nowhere near capacity at expected log rates, including DEBUG/VOD bursts.

SergeantPanda:
I noticed while doing my final testing that the log collector wasn't cleanly shutting down so Docker was exiting with a 137 code.

Should be good now, merged it in, thank you!

Issue #1299, opened 2026-05-28
[Bug]: Auto Channel Sync doesn't preserve channel profile enable/disable selections on refresh


ChilledSlim:
### Describe the bug

While I understand the purpose of Auto Channel Sync is to address groups that change often (ie: PPV), it forces all channels to be enabled in all profiles every time it refreshes.  This is awful and makes the feature not very useful.

I have a channel profile with all of my channels that I pass through to my TV.  I also have a channel profile with 50 channels or so to pass through a limited channel profile/playlist and all other channels which have been added are disabled. [example uses: the living room gets all channels, the kids room gets some kids channels; the living room gets all the channels, something passed through to Jellyfin just gets sports].

Every time the M3U refreshes, it recreates the channels in the limited channel profile, when these were previously disabled.

* At a minimum, while I understand Auto Channel Sync overrides changes (such as channel name, number, etc) shouldn't it preserve the enable/disable selection per channel profile? (side-effect: this would add new channels but at least not re-enable reused/old ones).
* Next best, shouldn't it recognize that all of the existing channels in the group are disabled and keep any new channels disabled?
* To be feature complete, should the Auto Channel Sync have a list of channel profiles it should enable new/existing channels to vs syncing to all of them and enabling them?

### How can we recreate this bug?

1. Make a channel profile (ie: Kids TV)
2. Add 50 channels
3. Turn on Auto Channel Sync for the M3U playlist for a group
4. Disable those channels in your Kids TV channel profile
5. Refresh the M3U playlist
6. They get re-enabled constantly

### Dispatcharr Version

v0.25.1

### Docker Version

Docker version 29.2.1, build a5c7197

### What's in your Docker Compose file?

services:
  dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:latest
    container_name: dispatcharr
    ports:
      - 9191:9191
    volumes:
      - /srv/data/dispatcharr:/data
      - /media/pool/Recordings:/data/recordings
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - CELERY_BROKER_URL=redis://localhost:6379/0
      - DISPATCHARR_LOG_LEVEL=info
    devices:
      - /dev/dri:/dev/dri  # For Intel/AMD GPU acceleration (VA-API)

### Client Information

N/A

MotWakorb:
Just to be clear on your steps:

[...]

PR #1512, opened 2026-08-04
Add a No Profiles auto-sync option and validate oversized logos

raclo:
## Description

Add an explicit **No Profiles** option to auto channel sync.

When **No Profiles** is selected, auto-sync continues creating and updating channels but skips channel-profile membership management entirely. It does not create, enable, disable, or otherwise reconcile `ChannelProfileMembership` rows. Existing memberships therefore remain under manual control.

The existing behavior is preserved when one or more real profiles are selected. Clearing the selection continues to use the historical default behavior of assigning all profiles.

The selection is stored in the existing `custom_properties` JSON as `skip_channel_profile_memberships: true`, so no database migration is required.

This PR also validates oversized provider logo values before querying the uniquely indexed logo field. Rejected provider values do not prevent channel creation and do not replace an existing valid logo.

## Related Issue

Addresses the auto-sync profile-control use case discussed in #1299.

Follow-up to #519: apply oversized-logo validation to auto-sync as well.

## How was it tested?

Backend:
- 10 targeted auto-sync/profile/logo regression tests passed.
- 95 related backend tests passed across:
  - `apps.m3u.tests.test_sync_correctness`
  - `apps.m3u.tests.test_sync_compound`
  - `apps.m3u.tests.test_rename_preview_parity`
  - `apps.m3u.tests.test_xc_empty_fetch_guard`
- Tests used isolated PostgreSQL 17 and Redis instances without production volumes.
- `git diff --check` passed.
- Python `compileall` passed.

Frontend:
- 38 Vitest tests passed for `AutoSyncAdvanced`.
- ESLint passed.
- Prettier check passed.

The regression coverage verifies that:
- normal profile selection retains the historical reconciliation behavior;
- **No Profiles** creates no memberships for new channels;
- existing enabled and disabled memberships remain unchanged;
- memberships in unrelated profiles remain unchanged;
- repeated refreshes with **No Profiles** are idempotent;
- switching from a profile to **No Profiles** preserves the current membership state;
- switching back from **No Profiles** to a real profile resumes historical reconciliation;
- the explicit `skip_channel_profile_memberships: true` flag takes precedence over stale profile IDs;
- oversized logo values are rejected before indexed queries;
- new channels are still created when a logo is rejected;
- an existing valid logo is preserved.

One lifecycle limitation remains unchanged: if auto-sync deletes the channel itself, its membership rows are naturally removed by the existing database cascade.

## Checklist

  - N/A — no models were changed.
  - N/A — no API endpoints were added or changed.


raclo:
I reviewed the contribution guidelines and rebased these two commits locally onto the current dev branch. I also reran the PostgreSQL/Redis regression suites, Docker build, and image smoke tests successfully. I am now updating the Draft PR branch with force-with-lease and changing the target to dev.

SergeantPanda (changes requested):
Thank you for the PR. 

I thought on this a bit and am not sure I want the behavior to change that much on the channel profiles for auto channel sync.

That being said, what if we instead offer a "No Profiles" option when selecting what profiles to auto the auto sync group to? If "No Profiles" is selected we could skip the assignments all together, giving the user more control over which profiles each channel belongs to.

Please let me know what you think and if that works for your use-case or not.

raclo:
Thanks for taking the time to think this through.

Yes, I think a “No Profiles” option would work well for my use case, provided that selecting it means auto-sync completely skips channel profile membership reconciliation.

In other words, I would expect auto-sync to continue creating and updating the channels themselves, but not create, enable, disable, or remove any "ChannelProfileMembership" entries. Profile membership would then remain entirely under manual control.

The trade-off that newly discovered channels would not automatically be added to any profile is perfectly fine for me. I would rather assign those manually than have existing manual profile selections changed during refreshes.

Keeping the current behavior for users who select one or more profiles, while adding “No Profiles” as an explicit opt-out, also sounds like a good way to provide more control without changing the existing behavior for everyone.

The oversized-logo validation in this PR is independent of the profile membership behavior, so I believe that part can remain as-is.

If this matches what you had in mind, I’m happy to adjust the profile-related part of the PR accordingly.

One small note: I’m currently away and won’t be able to dedicate proper time to the implementation until around August 24, but I can pick it up then.

SergeantPanda:
Sounds like we are on the same page then. I will mark this as a draft for now but yeah when you get time, update this PR and give me a ping 🙂.

raclo:
I have now implemented the agreed No Profiles approach locally, rebased it onto the current dev branch, and rerun the backend and frontend regression checks successfully. I am about to update the PR branch with force-with-lease, then I will refresh the PR description and mark it ready for review again.

raclo:
Updated as discussed. The PR now adds an explicit No Profiles option that leaves channel-profile memberships entirely untouched while preserving the historical reconciliation behavior for normal profile selections. I also kept the oversized-logo fix independent. Backend and frontend regression checks are passing. @SergeantPanda, this is ready for another look when you have time.

SergeantPanda (approved):
This looks much better, thank you for making the adjustments.

I did make a couple of changes to the logo validation code so it doesn't need to run twice.

I also moved "No Profiles" to the bottom of the list.

PR #1514, opened 2026-08-04
fix: stable VOD episode/movie IDs, fix 500 crash on stale/missing relation

jstevenscl:
## Description

Fixes both the VOD "episode not found" 500 crash and its underlying cause: VOD episode/movie playback ids handed to XC clients are not stable across a catalog refresh.

`xc_get_vod_streams`, `xc_get_vod_info`, and `xc_get_series_info` currently emit the underlying `Movie`/`Episode` row's own database PK as the client-facing playback id. That PK is not stable -- a catalog refresh can delete and recreate the row. `stream_xc_movie`/`stream_xc_episode` then look up `M3UMovieRelation`/`M3UEpisodeRelation` by that same unstable PK, so once a refresh happens, every id a client (e.g. TiviMate) has cached goes stale. Combined with a `.first()`/`DoesNotExist` mismatch in `stream_xc_episode` (`.filter(...).first()` returns `None` on no match, it never raises `DoesNotExist`, so the `except` clause is dead code), a stale id crashes with `AttributeError: 'NoneType' object has no attribute 'episode'` instead of failing cleanly with a 404.

`xc_get_series` already avoids this: it exposes `M3USeriesRelation`'s own PK, not `Series.id`, because relations are upserted in place (keyed by `(m3u_account, stream_id)` from the provider) and are therefore both stable across a refresh and globally unique. This PR brings movies and episodes in line with that existing pattern, rather than switching to the provider's raw `stream_id` -- which would introduce a *new* cross-account collision risk, since `stream_id` is only unique per `(m3u_account, stream_id)`, not globally, and provider data can put the literal string `"None"` there when a field is missing.

**Changes:**
- `xc_get_vod_streams`: `"stream_id"` is now the `M3UMovieRelation` PK.
- `xc_get_vod_info`: same for both the lookup (`vod_id`) and the emitted `movie_data.stream_id`.
- `xc_get_series_info`: per-episode `"id"` is now the `M3UEpisodeRelation` PK (falls back to `episode.id` only when there's no active relation at all, which isn't playable either way).
- `stream_xc_movie` / `stream_xc_episode`: look up by the relation's own PK instead of `movie_id`/`episode_id`, and both now cleanly 404 when no relation matches instead of the episode path crashing.
- Removed `apps/output/views.py`'s `xc_movie_stream`/`xc_series_stream` (and their `urls.py` import) -- confirmed dead code, never wired to a URL route (`CHANGELOG.md` documents they were superseded by the `vod_proxy` versions), so fixing them too would just leave a second, still-unreachable copy of the same bug pattern to maintain.
- Updated the one existing test asserting on the old value (`test_vod_streams_response_keys`) and the `xc_get_vod_info` test fixture that passed `movie.id` where it now needs `relation.id`.
- Added `apps/proxy/vod_proxy/tests/test_xc_stream_ids.py`: direct coverage for `stream_xc_movie`/`stream_xc_episode` -- valid relation id streams, the exact pre-fix stale-PK crash case now 404s cleanly, inactive account, garbage id.

Related to #820 (same crash signature, in the now-removed `xc_series_stream` redirect function). That issue's resolution stopped it from reproducing there by routing around the dead function, but never addressed the root cause -- which was still live and reachable via `stream_xc_episode` today, on the current `dev` branch.

## Related Issue

Related to #820 (not a direct duplicate -- different function, and the root cause was never fixed, just no longer reachable in the code path that issue reported).

Closes #1513 (my own earlier, narrower PR against `main` for just the crash -- closed in favor of this one once I found the deeper cause; the crash-only fix would have conflicted with this one since they touch the same lines).

## How was it tested?

1. **Reproduced the live crash** against a disposable `ghcr.io/dispatcharr/dispatcharr:latest` container (no production data involved): with a `stream_id` that has no matching relation, `GET /series/<user>/<pass>/<id>.mp4` 500s with the exact traceback described above.
2. **Verified the fix removes the crash**: same request against the patched code returns a clean 404 `{"error": "Episode not found"}`.
3. **Verified against real production data** (read-only queries, no writes) that `M3UEpisodeRelation.stream_id` never equals the Episode FK `episode_id` across 238 live relations on a real instance -- confirming the two id spaces are genuinely disjoint, not usually-coincidentally-equal.
4. **Ran the full Dispatcharr backend test suite** for the affected apps against the patched code, using the project's own `base-dev` CI image and `scripts/ci_bootstrap_backend.sh` (internal Postgres + Redis, `manage.py test`):
   ```
   docker run --rm --entrypoint '' \
     -e DISPATCHARR_ENV=aio -e DJANGO_SECRET_KEY=ci-test-secret-key \
     -e POSTGRES_DB=dispatcharr -e POSTGRES_USER=dispatch -e POSTGRES_PASSWORD=secret \
     -v <repo>:/app -w /app \
     ghcr.io/dispatcharr/dispatcharr:base-dev \
     bash scripts/ci_bootstrap_backend.sh apps.output.tests apps.proxy.vod_proxy.tests -v2
   ```
   Result: **117 tests, all passing** (including every existing `apps.output`/`apps.proxy.vod_proxy` test after updating the two that asserted on the old id scheme, plus the 8 new tests added by this PR).

## Checklist

Notes on items that are trivially satisfied rather than actively exercised: no models changed (no migrations needed), no new endpoints added (existing ones only changed which id value they return/accept, no OpenAPI shape change), and this is a backend-only change with no frontend files touched (lint/format not applicable, vacuously passing).

### DIFF
--- a/apps/output/tests/test_views.py
+++ b/apps/output/tests/test_views.py
@@ -804,7 +804,7 @@ class XcVodSeriesRegressionTests(TestCase):
     def test_vod_streams_response_keys(self):
         account = self._account(f"acct-{uuid4().hex[:6]}")
         movie = Movie.objects.create(name="Schema Movie", rating="10")
-        M3UMovieRelation.objects.create(
+        relation = M3UMovieRelation.objects.create(
             m3u_account=account, movie=movie, stream_id="schema-1"
         )
 
@@ -812,7 +812,9 @@ class XcVodSeriesRegressionTests(TestCase):
 
         self.assertEqual(set(stream.keys()), XC_VOD_STREAM_KEYS)
         self.assertEqual(stream["stream_type"], "movie")
-        self.assertEqual(stream["stream_id"], movie.id)
+        # The relation's own PK, not the Movie's -- stable across a catalog
+        # refresh (relations are upserted in place; Movie rows are not).
+        self.assertEqual(stream["stream_id"], relation.id)
         self.assertEqual(stream["rating_5based"], 5.0)
         self.assertEqual(stream["custom_sid"], None)
         self.assertEqual(stream["direct_source"], "")
--- a/apps/output/tests/test_xc_vod_info.py
+++ b/apps/output/tests/test_xc_vod_info.py
@@ -39,7 +39,9 @@ class XCGetVodInfoArtworkTests(TestCase):
 
     def _info(self):
         request = self.factory.get('/player_api.php')
-        return xc_get_vod_info(request, self.user, str(self.movie.id))
+        # vod_id is the M3UMovieRelation's own PK, not the Movie's -- see the
+        # note in xc_get_vod_info on why the lookup changed.
+        return xc_get_vod_info(request, self.user, str(self.relation.id))
 
     def test_cover_big_and_movie_image_are_identical(self):
         """Real XC servers return the same URL for both fields; clients may read either."""
--- a/apps/output/urls.py
+++ b/apps/output/urls.py
@@ -1,5 +1,5 @@
 from django.urls import path, re_path, include
-from .views import m3u_endpoint, epg_endpoint, xc_get, xc_movie_stream, xc_series_stream
+from .views import m3u_endpoint, epg_endpoint, xc_get
 
 app_name = "output"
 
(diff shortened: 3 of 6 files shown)


SergeantPanda:
Thank you for the PR.

For this. I do not want to change to outputting the pk for the relation table. That sort of defeats what we are trying to do with Dispatcharr.

If our canonical PKs are changing after refresh, I’d rather fix that ingest/dedup/cleanup bug than change what we expose to XC clients.

The playback crash you found (stale/missing ID returning 500 instead of 404) would be a great standalone PR if you’re up for splitting that out.

Issue #799, opened 2026-01-02
[Feature]: Add complete DVR recordings to VOD library


91jme:
### Is your feature request related to a problem?

DVR recordings can only be viewed via the Dispatcharr Web UI, not clients like TiviMate. A simpler request as requested in #505 (I couldn't find a follow-up to that via search)

### Describe the solution you'd like

Once a recording is complete, add it as a VOD show episode so that it can be consumed via client like TiviMate.

### Additional context

Nothing Extra

toddzim33:
Wanted to add some validation and technical detail. Everything dispatcharr needs is already built:

List completed recordings with metadata: GET /api/channels/recordings/
Stream files with Range/seeking support: GET /api/channels/recordings/{id}/file/ 
Channel logos for poster fallback: GET /api/channels/logos/{id}/ 
M3U/XMLTV generation pattern: Already exists for channels

I validated this end-to-end using only existing dispatcharr API endpoints, no modifications to dispatcharr required. Queried the recordings API, generated an M3U pointing at /api/channels/recordings/{id}/file/, loaded it into TiviMate as a VOD playlist.

Result: recordings appear, playback works, seeking works. The file endpoint handles Range requests correctly out of the box.

The only missing piece is the M3U output endpoint itself, applying the same generation logic that already exists for channels to the recordings list.

UI suggestion: The DVR page could expose this the same way the Channels page already does, M3U and EPG copy-URL buttons at the top. Users paste once into their client; new recordings appear automatically on the next playlist refresh. No new UI patterns needed,  just a second instance of what's already there.
Poster art: The recordings API doesn't store show poster URLs. Channel logos work as a fallback but are landscape-oriented and look stretched in portrait VOD UIs. Capturing a poster URL from EPG program data at record time would improve this significantly.

PiratesIRC:
I am also looking for this: I record live events with the DVR and want to browse and play them from TiviMate on a Shield, ideally in the Movies / Series section with a poster and per-show grouping, and with the DVR access levels (none / view / manage) still respected for my non-admin users.
A few things I measured on 0.30.0 that change the picture since the March comments above:
The recording file endpoint, GET /api/channels/recordings/{id}/file/, now requires authentication (JWT, API key header, or a ?token= query parameter). Without a token it returns 403. With a token it returns 200 and supports Range requests correctly (206 with Content-Range), so seeking works.
[...]

PR #1515, opened 2026-08-05
feat(output): expose completed DVR recordings as Xtream VOD

a-monteiro:
## What

Exposes completed local DVR recordings through the existing Xtream Codes VOD output as a virtual **Recordings** category.

Completed recordings now:

- appear in `get_vod_categories` only when an accessible recording exists;
- appear in both category-scoped and unfiltered `get_vod_streams` responses (for clients that filter locally);
- provide `get_vod_info` metadata; and
- play through the existing credentialed, range-capable `/movie/...` route.

Incomplete, missing, hidden, or unauthorized recordings remain absent.

## Why

Addresses [#799](https://github.com/Dispatcharr/Dispatcharr/issues/799)

Completed DVR recordings were available only through the web UI, rather than normal Xtream clients.

## Implementation

- Uses a virtual `recording-<id>` stream ID and a `dispatcharr-recordings` VOD category; no database migration or new public API route is introduced.
- Reuses the existing recording-file response, including Range support, after Xtream credentials and source-channel visibility checks.
- Keeps the collection lazy for category discovery, filters completed rows in the database, and treats file disappearance during discovery as unavailable.

## Validation

- `python3 -m py_compile` on each touched Python module
- Docker image build and publish completed successfully
- Deployed the image into an isolated TrueNAS staging app and verified against its live Xtream API:
  - two independently completed recordings are returned in both unfiltered and `Recordings`-scoped VOD lists;
  - VOD detail metadata resolves for a recording;
  - credentialed MKV Range playback returns HTTP `206` and 1024 requested bytes.
- Added Django `TestCase` coverage for category/listing, unfiltered client behavior, incomplete/missing files, file-stat race handling, playback Range support, metadata, and hidden-channel authorization.

The local VM cannot run the Django test process because its NumPy wheel requires x86-64-v2 CPU support. Please let GitHub Actions run the test module on a compatible runner.

## Checklist

- [x] Linked to an existing issue
- [x] Targets `dev`
- [x] No migrations or dependencies added
- [x] No secrets, provider URLs, or local infrastructure details included
- [x] I understand and agree to the contributor license terms

### DIFF
--- a/apps/output/tests/test_recordings_xtream_vod.py
+++ b/apps/output/tests/test_recordings_xtream_vod.py
@@ -65,6 +65,10 @@ class XcRecordingVodTests(TestCase):
         self.assertEqual(streams[0]["container_extension"], "mkv")
         self.assertNotIn(self.file.name, str(streams[0]))
 
+    def test_unfiltered_vod_streams_include_completed_recordings_for_clients_that_filter_locally(self):
+        streams = xc_get_vod_streams(self.factory.get("/player_api.php"), self.user)
+        self.assertEqual([row["stream_id"] for row in streams], [f"recording-{self.recording.id}"])
+
     def test_incomplete_or_missing_recording_is_not_exposed(self):
         now = timezone.now()
         Recording.objects.create(
--- a/apps/output/tests/test_recordings_xtream_vod.py
+++ b/apps/output/tests/test_recordings_xtream_vod.py
@@ -7,6 +7,7 @@ from uuid import uuid4
 from django.test import TestCase
 from django.utils import timezone
 from django.test.client import RequestFactory
+from django.http import Http404
 from rest_framework.test import APIClient
 
 from apps.accounts.models import User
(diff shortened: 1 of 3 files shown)


SergeantPanda (commented):
Thank you for the PR with a minimal diff, this is actually something we've been working on but it's part of a much larger feature that is (hopefully) coming soon.

I'll verify with the dev of the feature but I highly suspect this will directly conflict with what he is working on.

a-monteiro:
> Thank you for the PR with a minimal diff, this is actually something we've been working on but it's part of a much larger feature that is (hopefully) coming soon.
> 
> I'll verify with the dev of the feature but I highly suspect this will directly conflict with what he is working on.

Happy to contribute!
 I've been using this myself but realized there was an open issue requesting it, so opened the PR. But please do let me know If I should expect this to get reviewed/merged or should close in favor of the upcoming feature.

SergeantPanda:
Sorry for the delay, yes this will be a part of an upcoming larger feature. I will close this. Thank you for the contribution though!

PR #1525, opened 2026-08-08
feat(channels): improve channel profile assignment workflow

### LINKED ISSUE #1080: [Feature]: Create an empty profile

What I would like is the ability to create a completely empty profile, without all channels from the "All" profile being automatically added.

_**Create an in new profile**_
Currently, when creating a new profile, all existing channels are automatically inherited from the "All" profile. This quickly becomes overwhelming, especially when creating a profile exclusively for PPV events or for a friend with their own XC account – as you end up with 150+ irrelevant TV channels in the profile.

The Hide function does not fully solve the problem, even though each profile contains all channels from "All". It does not change the fact that the channels are still there in the background – they are simply hidden, not removed. The desire is still to have a truly empty profile, where you decide which channels should be available.

_**Delete function in new profiles**_
When deleting a channel from a new profile, it should not automatically be deleted from the "All" profile, as "All" functions as the primary channel collection.

_**Create an in new profile**_
Requested functionality:
- A new profile starts completely empty with no channels
- You can manually select and add channels from "All" to the new profile

Example: 
I create a new profile called "XC account":
- The profile starts with no channels
- I manually add the desired channels from "All", or add all channels at once.

_**Delete function in new profiles**_
Instead, the following behavior is requested:
When deleting a channel from a new profile, a confirmation dialog should appear with two options:

Delete from this profile only – the channel is removed from the current profile only, but remains in "All"
Delete from this profile and ALL – the channel is removed from both the current profile and "All"

Example:
I have a channel in my "XC account" profile that I want to remove. I delete the channel, and the system asks whether I only want to remove it from "XC account", or if it should also be deleted from "All". This way I have full control over what happens to the channel in the "All" profile.

This would make it far more flexible to manage profiles with different content and access levels.

### Additional context

Nothing Extra

Veneziaisking:
## Description

This PR improves the channel profile workflow by allowing users to create empty profiles and selectively add channels to them.

### Changes

- Add a "Start with no channels" option when creating a channel profile.
- Add an "Add to Profile..." action for selected channels.
- Make the "Add to Profile..." action visible in the main toolbar when channels are selected.
- Change the default channel view for real profiles to show only channels assigned to that profile.
- Keep the existing "Show Disabled" functionality available for users who want to browse and enable unassigned channels.

### Example workflow

1. Create a profile, e.g. `Bo`, with "Start with no channels".
2. The new profile starts with no channels.
3. Select the `All` profile.
4. Select the channels that should be available to `Bo`, e.g. YES Network US and We TV US.
5. Click "Add to Profile...".
6. Select `Bo`.
7. `Bo` now contains only the selected channels.

The `All` profile is not modified.

The delete-confirmation portion of issue #1080 is intentionally not included in this PR and can be addressed separately.

## Related Issue

Addresses #1080

## How was it tested?

- Backend channel profile tests: 4/4 passed.
- Existing channels test suite: 397/397 passed.
- Frontend relevant tests: 187/187 passed.
- Full frontend suite: 6117/6117 passed.
- ESLint and Prettier checks were run after the final changes.
- Manual testing completed successfully using an isolated `dispatcharr-test` Docker installation.
- Verified that the existing Dispatcharr installation and database were not modified.

## Checklist

allchannels
BOprofiles
addtoprofil
channelsadded
bochannels


SergeantPanda (approved):
Thank you for the PR!

Overall this looks really good. I'm going to merge after the changes I just made and make some more minor adjustments to the channel profile flow.

I changed the verbiage slightly on the channel profile create modal and adjusted some code comments.

PR #1533, opened 2026-08-11
fix(epg): stop Schedules Direct guide refresh from silently dropping …

sethwv:
What: Adds retry-with-backoff and proper failure handling to Schedules Direct's batched /schedules/md5, /schedules, and /programs requests.

Why: A failed MD5 batch was treated as "zero changes" and reported false success, a failed program metadata batch wrote placeholder "No Title" rows instead of retrying, and transient failures (timeouts, 5xx, SD's own embedded error codes on HTTP 200) weren't retried at all, so channels could silently end up with no guide data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


sethwv:
This PR is an attempt to add more information + transparency for issue reported in Discord and a first pass at a retry mechanism to see if that fils in the data from transitory api errors. https://discord.com/channels/1340492560220684331/1536704877575675934

Issue #1537, opened 2026-08-12
[Feature]: Pass upstream provider expiration through XC user_info


ayala:
### Is your feature request related to a problem?

Dispatcharr's XC `user_info.exp_date` always reports a synthetic date 90 days in the future, even when active M3U account profiles already contain the actual upstream provider expiration. Generic Xtream-compatible clients therefore cannot display the expiration Dispatcharr has already discovered.

### Describe the solution you'd like

Return the earliest known expiration from active `M3UAccountProfile` records attached to active M3U accounts in `user_info.exp_date`. Preserve the current 90-day value as a fallback when no upstream expiration is known. This uses the existing standard Xtream field and requires no client-specific behavior or new API surface.

### Additional context

The change should be a generic passthrough for every Xtream-compatible client. Inactive accounts and profiles should not influence the value.

Issue #1536, opened 2026-08-12
Pass upstream provider expiration through XC user_info


ayala:
## Problem

Dispatcharr's Xtream-compatible `player_api.php` response currently sets `user_info.exp_date` to an artificial `now + 90 days` value, even when active upstream M3U/XC profiles already have a provider-reported expiration stored in `M3UAccountProfile.exp_date`.

That prevents generic Xtream clients from displaying the actual subscription expiration.

## Proposed change

When building XC `user_info`:

- return the earliest known expiration among active profiles belonging to active M3U accounts;
- serialize it as the Unix timestamp string expected by Xtream clients;
- retain the existing 90-day value only when no active profile has a known expiration.

This requires no schema or endpoint change and remains client-agnostic. The earliest date is conservative for installations aggregating multiple upstream accounts.

## Testing

Add output-view tests covering provider expiration passthrough, inactive profile/account exclusion, earliest-expiration selection, and the existing fallback.

PR #1539, opened 2026-08-12
Pass provider expiration through XC output

ayala:
## What

Pass the earliest known upstream provider expiration from active M3U account profiles through the standard Xtream `user_info.exp_date` field.

## Why

Dispatcharr already stores provider expiration dates, but its XC output always returns a synthetic date 90 days in the future. Xtream-compatible clients therefore cannot display the real expiration Dispatcharr has discovered.

## How

- Select the earliest non-null expiration from active profiles on active M3U accounts.
- Return it as the existing Unix timestamp string in `exp_date`.
- Preserve the existing 90-day fallback when no provider expiration is known.
- Ignore inactive accounts and profiles.

## Tests

- Earliest active provider expiration is passed through.
- Inactive accounts and profiles are excluded.
- The legacy fallback remains unchanged when no expiration is known.

Related to #1536, #1537, and #1538. Those issues were automatically closed because GitHub's issue-form provenance cannot be supplied through the CLI; this PR contains the requested implementation and test coverage.

### DIFF
--- a/apps/output/tests/test_views.py
+++ b/apps/output/tests/test_views.py
@@ -8,8 +8,13 @@ from django.test.utils import CaptureQueriesContext
 from apps.channels.models import Channel, ChannelGroup, ChannelOverride, ChannelProfile, ChannelProfileMembership
 from apps.epg.models import EPGData, EPGSource
 from apps.accounts.models import User
-from apps.m3u.models import M3UAccount
-from apps.output.views import xc_get_live_streams, xc_get_series, xc_get_vod_streams
+from apps.m3u.models import M3UAccount, M3UAccountProfile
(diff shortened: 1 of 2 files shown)


SergeantPanda:
Thank you for the PR.

I'm not sure we want to do this and here is why: Some apps refuse to work if it thinks the account is expired. So, if you have 1 account that expires but have others that have not, the app would refuse to work even though Dispatcharr would still happily serve channels from other providers

IF we were going to do something like this, a floor of maybe 7 days would prevent that behavior but then becomes unreliable.

I'd probably rather have a webhook that could send a notification when accounts get close to expiration.

PR #1547, opened 2026-08-14
Fix/truncate logo names

SergeantPanda:
This PR adds a reusable helper to truncate strings and log a warning when it does. 

It is wired in to Logo names, M3U, EPG and SD.


nagelm:
dev's Frontend Tests have been failing since this merged — the LogoForm suite's LogoUtils mock is missing the new LOGO_NAME_MAX_LENGTH export, so every render in that file crashes. Fix is in #1552.

SergeantPanda:
Yes I got the email just didn't have time to fix last night. I will be fixing the tests.

PR #1555, opened 2026-08-16
fix(docker): raise worker RLIMIT_NOFILE past the pam_limits reset in entrypoint.sh

### LINKED ISSUE #1554: [Bug]: uwsgi/daphne/celery are pinned to RLIMIT_NOFILE soft=1024 — compose `ulimits` cannot raise it; daphne fd exhaustion kills all streaming while the container stays "healthy"
### Describe the bug

**Summary.** In the AIO image, `uwsgi`, `daphne` and `celery` always run with `RLIMIT_NOFILE` **soft = 1024**, and an operator **cannot raise it**. Setting `ulimits: nofile:` in `docker-compose.yml` has no effect on them — it only applies to PID 1. Under sustained live-streaming load `daphne` accumulates file descriptors until it hits 1024, after which **every stream start fails** with `[Errno 24] Too many open files` while the web UI keeps returning HTTP 200 and the container stays `healthy`.

**Why the limit can't be raised.** `docker/entrypoint.sh` launches uwsgi through a login shell:

```bash
nice -n "$UWSGI_NICE_LEVEL" su - "$POSTGRES_USER" -c "cd /app && exec $VIRTUAL_ENV/bin/uwsgi $uwsgi_args" &
```

`/etc/pam.d/su` line 52 is `session required pam_limits.so`, and the image ships no `nofile` entry in `/etc/security/limits.conf` or `limits.d/`. `pam_limits` therefore resets `RLIMIT_NOFILE` to 1024 for that session, discarding whatever the container was given. Measured on a stock 0.28.1 AIO container started with `ulimits: nofile: {soft: 65536, hard: 524288}`:

```
$ docker exec dispatcharr bash -c 'ulimit -n'
65536                                   <- what compose set (PID 1 namespace)
$ docker exec dispatcharr bash -c 'su - dispatch -c "ulimit -n"'
1024                                    <- what uwsgi/daphne/celery actually get
```

Per-process confirmation from `/proc/<pid>/limits` on that same container:

```
PID 1    Max open files  65536   524288    /bin/bash /app/docker/entrypoint.sh
PID 259  Max open files  1024    524288    /dispatcharrpy/bin/uwsgi --ini /app/docker/uwsgi.ini
PID 276  Max open files  1024    524288    /dispatcharrpy/bin/python .../daphne -b 0.0.0.0 -p 8001
PID 274  Max open files  1024    524288    /dispatcharrpy/bin/python .../celery -A dispatcharr
```

Because `daphne` is an `attach-daemon` of the uwsgi master (`docker/uwsgi.ini` line 17), it inherits the same capped limit. The hard limit is already 524288, so **nothing but the soft limit is in the way.**

**What the failure looks like in production.** After ~3 weeks uptime, `daphne` reached 1023/1024 fds and the streaming path died while every health signal stayed green:

```
ERROR live_proxy.manager Error establishing transcode connection for channel <uuid>: [Errno 24] Too many open files
OSError: [Errno 24] Too many open files
INFO  live_proxy.manager URL http://<provider>/live/<user>/<pass>/2393.ts failed after 3 attempts, trying next stream
ERROR live_proxy.manager Failed to find alternative streams after 7 attempts for channel: <uuid>
WARNING core.utils Failed to send WebSocket update: [Errno 24] Too many open files
```

Two things make this expensive to diagnose:

1. **The healthcheck cannot see it.** `docker/docker-compose.aio.yml` health is an HTTP probe of the web root; serving that page opens almost no descriptors, so the container reports `healthy` for the entire outage. `player_api.php` also answers normally.
2. **It reads as a dead provider.** Each channel exhausts all its sources on the way down, so the log fills with failing provider URLs. The provider was fine — the proxy simply could not open a socket.

**This is not #1416/#1418 or #408.** Both of those resource-leak fixes (PR #1420, and #1356 for #408) are already in this build — I verified `64912c6` and `1918b7d` are ancestors of the 0.28.1 build commit `67ccbd7`. The fd ceiling is a separate problem, and it is the one an operator cannot work around from compose.

**Expected behavior.** The workers should either start with a sensible fd limit, or honour the `ulimits:` an operator sets in compose. Today neither is true, and the only remedy is restarting the container every few weeks.

The capped limit reproduces immediately on any AIO container — no load required:

1. Start the stock AIO compose, but add an explicit generous limit so it's clear the operator asked for one:
   ```yaml
   ulimits:
     nofile:
       soft: 65536
       hard: 524288
   ```
2. `docker exec dispatcharr bash -c 'ulimit -n'` → **65536** (looks correct)
3. `docker exec dispatcharr bash -c 'su - dispatch -c "ulimit -n"'` → **1024** (what the workers really get)
4. Confirm on the live processes: `docker exec dispatcharr bash -c 'grep "Max open files" /proc/$(pgrep -f bin/daphne | head -1)/limits'` → soft **1024**

To reach the outage, run live streams until daphne's descriptor count reaches 1024:

```bash
docker exec dispatcharr bash -c 'ls /proc/$(pgrep -f bin/daphne | head -1)/fd | wc -l'
```

On our instance (~40 channels, 620 dual-sourced failover channels, continuous viewing) that took about three weeks. Channel zapping and client reconnects accelerate it.

### Dispatcharr Version

0.28.1

### Docker Version

Docker version 28.5.2, build ecc6942 (Compose v2.40.3)

```yaml
services:
  dispatcharr:
    container_name: dispatcharr
    hostname: dispatcharr
    restart: unless-stopped
    cpu_shares: 512
    mem_limit: 24g
    memswap_limit: 26g
    ulimits:              # <-- added while debugging; does NOT reach uwsgi/daphne/celery
      nofile:
        soft: 65536
        hard: 524288
    healthcheck:
      test: ["CMD-SHELL", "curl -fsS -o /dev/null http://localhost:9191/ || exit 1"]
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 90s
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - TZ=Asia/Ho_Chi_Minh
      - DISPATCHARR_LOG_LEVEL=info
    volumes:
      - /opt/dispatcharr/data:/data
    ports:
      - 9191:9191
```

### Client Information

Emby Live TV (HDHR tuner emulation) and SparkleTV / TiviMate on Android TV, plus Xtream Codes clients. The failure is server-side and client-independent — no client can start a stream once the limit is reached.

### Workaround (for anyone hitting this before a fix lands)

Bind-mount a `pam_limits` drop-in so the `su -` session keeps a usable limit:

```yaml
volumes:
  - ./99-dispatcharr-nofile.conf:/etc/security/limits.d/99-dispatcharr-nofile.conf:ro
```

```
dispatch soft nofile 65536
dispatch hard nofile 524288
```

Verified this raises the running uwsgi/daphne/celery processes to soft 65536. Note the username must match `POSTGRES_USER` (default `dispatch`). A plain `docker restart` clears the leaked descriptors but does not change the limit, so the outage returns.

I have a patch for `docker/entrypoint.sh` and will open a PR against `dev` referencing this issue.

brendongl:
## Description

On every AIO deployment, `uwsgi` and its attach-daemons (`daphne`, `celery`, `redis`) start with `RLIMIT_NOFILE` **soft = 1024**, and an operator has no way to raise it.

`docker/entrypoint.sh` launches uwsgi through a login shell:

```bash
nice -n "$UWSGI_NICE_LEVEL" su - "$POSTGRES_USER" -c "cd /app && exec $VIRTUAL_ENV/bin/uwsgi $uwsgi_args" &
```

`su -` opens a PAM session; `/etc/pam.d/su` loads `pam_limits.so`, and the image ships no `nofile` entry in `/etc/security/limits.conf` or `limits.d/`. The limit is therefore reset to 1024 *after* Docker has applied the container's limits, so a `ulimits: nofile:` entry in `docker-compose.yml` only ever affects PID 1 and never reaches the processes that actually open descriptors.

Measured on a stock 0.28.1 AIO container started with `ulimits: nofile: {soft: 65536, hard: 524288}`:

```
$ docker exec dispatcharr bash -c 'ulimit -n'
65536                                   <- what compose set
$ docker exec dispatcharr bash -c 'su - dispatch -c "ulimit -n"'
1024                                    <- what the workers actually get
```

```
PID 1    Max open files  65536   524288    /bin/bash /app/docker/entrypoint.sh
PID 259  Max open files  1024    524288    /dispatcharrpy/bin/uwsgi --ini /app/docker/uwsgi.ini
PID 276  Max open files  1024    524288    .../daphne -b 0.0.0.0 -p 8001
PID 274  Max open files  1024    524288    .../celery -A dispatcharr
```

`daphne` holds a descriptor per proxied stream and websocket, so a busy instance saturates 1024 and then fails **every** new stream:

```
ERROR live_proxy.manager Error establishing transcode connection for channel <uuid>: [Errno 24] Too many open files
ERROR live_proxy.manager Failed to find alternative streams after 7 attempts for channel: <uuid>
WARNING core.utils Failed to send WebSocket update: [Errno 24] Too many open files
```

Meanwhile the web UI still returns 200 and the healthcheck still reports `healthy`, so nothing surfaces the outage. Because each channel exhausts all of its sources on the way down, the logs fill with failing provider URLs and it reads like a dead upstream.

**The change.** Raise the soft limit inside the su'd shell, before `exec`, so it lands after PAM:

```bash
UWSGI_MAX_FD="${UWSGI_MAX_FD:-65536}"
nice -n "$UWSGI_NICE_LEVEL" su - "$POSTGRES_USER" -c "ulimit -n $UWSGI_MAX_FD 2>/dev/null || true; cd /app && exec $VIRTUAL_ENV/bin/uwsgi $uwsgi_args" &
```

Notes on the specific choices:

- **Why inside the `su -c` string** rather than before `su`: anything set in the parent is discarded by `pam_limits` when the session opens. It has to run inside the session.
- **Why this works unprivileged**: the hard limit is already 524288, and lowering-or-raising the *soft* limit up to the hard limit requires no privileges, so it is fine after the drop to `$POSTGRES_USER`. I did not touch the hard limit, which would need root.
- **Why `2>/dev/null || true`**: `entrypoint.sh` runs under `set -e`. If someone sets `UWSGI_MAX_FD` above the hard limit (or runs with a restrictive hard limit), an unguarded `ulimit` failure would abort startup. Guarded, the container keeps its inherited limit and still boots.
- **Why an env var**: matches the existing `UWSGI_NICE_LEVEL` / `CELERY_NICE_LEVEL` convention, and lets operators tune it without a rebuild. Default 65536.
- **Scope**: only this one launch path needs it. `docker/entrypoint.celery.sh` (modular mode) runs celery directly without `su -`, so it already inherits the container's limit — I left it alone.
- The `daphne`/`celery`/`redis` attach-daemons are children of the uwsgi master (`docker/uwsgi.ini`), so they inherit the raised limit without separate changes.

## Related Issue

Closes #1554

## How was it tested?

**Verified on a live 0.28.1 AIO container** (the instance where the outage happened), by running the patched construct verbatim with a harmless `exec` target in place of uwsgi:

```
--- default (65536) ---
worker_nofile=65536
--- env override (UWSGI_MAX_FD=32768) ---
worker_nofile=32768
--- above hard limit (UWSGI_MAX_FD=99999999): must not abort under set -e ---
worker_nofile=65536
SURVIVED set -e after over-limit request
```

Also confirmed the underlying mechanism on that container before writing the patch:

- `docker exec … 'ulimit -n'` → 65536 vs `su - dispatch -c "ulimit -n"` → 1024, i.e. PAM is what resets it
- adding a `pam_limits` drop-in at `/etc/security/limits.d/` raises the running `uwsgi`/`daphne`/`celery` to 65536, which independently confirms the diagnosis
- after applying the equivalent fix, `/proc/<pid>/limits` for the live `uwsgi`, `daphne` and `celery` processes all read soft 65536 (previously 1024), and the instance has been serving streams normally since

`bash -n` passes on both changed files.

**What I have not run:** the new `docker/tests/test-nofile-limit.sh` end-to-end, because that builds the image from scratch and I only had the running production container to work against. The scenarios encode exactly the assertions I verified by hand above (`default_limit`, `pam_reset`, `env_override`, `above_hard_limit`, `compose_ulimits`), and it follows the structure of `test-puid-pgid.sh` / `test-tls-postgres.sh`, but it deserves a real run in CI or by a maintainer before you trust it. Happy to iterate on it.

## Checklist

- [x] Backend: migrations are included if any models were changed *(n/a — no model changes)*
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema *(n/a — no new endpoints)*
- [x] Frontend: ESLint and Prettier pass cleanly *(n/a — no frontend changes)*
- [x] Tests are included for new functionality *(see the caveat above — added but not executed end-to-end)*
- [x] Existing tests still pass *(no Python/JS touched; both changed shell files pass `bash -n`)*

---

### Disclosure

I should be transparent given the "On AI-Assisted Code" section of CONTRIBUTING.md: **I am an AI agent (Claude). I found this bug, diagnosed it, and wrote this patch, the test script, and this description**, while investigating a real streaming outage on my operator's production instance. The measurements and command output quoted above are real, taken from that instance — not illustrative.

Flagging it plainly because your policy asks that submitted code be understood rather than pattern-matched, and you should be able to weigh that when reviewing. Two things I'd particularly like a maintainer's eye on: whether `UWSGI_MAX_FD` is the naming you'd want, and whether 65536 is the right default for you versus something lower. My operator (@brendongl) is running the fix in production and can answer questions or test changes on request.


SergeantPanda:
Curious, what level of usage do you have when you start seeing these issues?

SergeantPanda:
Thanks for the excellent diagnosis and PR. The PAM/`su -` reset explanation and the production measurements made this straightforward to review. The core fix is solid and we are keeping it.

A few small maintainership edits on top of your branch:

1. Env var rename: `UWSGI_MAX_FD` to `DISPATCHARR_NOFILE`. The limit applies to the whole worker tree (uwsgi plus daphne/celery attach-daemons), not just uwsgi, and matches other `DISPATCHARR_*` knobs.
2. Shorter entrypoint comment with the same intent.
3. Compose examples: documented `DISPATCHARR_NOFILE` as its own optional block, not under process priority / nice. Nice and nofile are unrelated.
4. Changelog entry under Unreleased Fixed, credited to you, Fixes #1554.
5. `above_hard_limit` test: after PAM, soft is still 1024 when the raise fails, so the assertion now expects `1024` instead of the pre-`su` Docker soft of `4096`.

We left the approach as-is (`ulimit` inside the `su -c` string). Alternatives like replacing `su -` with `gosu`/`setpriv`, a `limits.d` drop-in, or uWSGI `max-fd` are either a larger refactor or worse under our privilege drop. Your approach is the right scoped fix.

Appreciate you tracking this down on a live outage and writing it up so clearly.

PR #1567, opened 2026-08-18
fix: reconnect live player on network error + fix stale-event race

jstevenscl:
## Description

The live-channel web player (`mpegts.js`, in `FloatingVideo.jsx`) had no recovery logic
on `mpegts.Events.ERROR` at all — a transient network error just showed a message and
stopped playback, permanently, until the user manually reopened the channel. Unlike the
VOD/recording player (`hls.js`, which self-heals via `startLoad()` /
`recoverMediaError()`), there was no equivalent for the live player.

This PR:
1. Adds reconnect-on-`NetworkError` for the live player: since `mpegts.js` has no
   in-place recovery API, destroy the player and recreate it pointed at the same stream.
   Exponential backoff (1s → 2s → 4s → 8s → 10s), capped at 5 attempts. Only
   `NetworkError` retries — `MediaError` (codec issue) is deterministic and would just
   fail identically every time, wasting the retry budget before showing the real message.
2. Fixes a stale-event race found while live-testing #1: `mpegts.js` can emit an event
   asynchronously from a player instance that's already been destroyed and replaced (a
   worker message already in flight when `.destroy()` was called). None of the event
   handlers checked whether the player an event belonged to was still the active one, so
   a late/stale event could destroy a newer, working player and revert it, or show a
   stale "Auto-play was prevented" message after a later player had already started
   playing successfully. Fixed with a `playerRef.current === player` guard on every
   handler.

## Related Issue

Part of #1566 (this PR covers items 1 and 3 from that issue; item 2, the AAC
output-profile auto-fallback, is a separate follow-up PR referencing the same issue).

## How was it tested?

**Automated**: 7 new Vitest tests (5 for reconnect backoff/retry-limit/give-up behavior,
2 for the stale-player-instance guard). The stale-guard tests use distinct mock player
instances per `createPlayer()` call — every other test in this file reuses one shared
mock instance across destroy+recreate cycles, so `playerRef.current !== player` is
trivially always false there and those tests can't catch this class of bug at all.
Verified the new tests actually catch the regression: temporarily reverted the guard
locally and confirmed both new tests fail (one recreates a third player where it must
not; the other shows the stale message where it must not), then restored the fix and
confirmed all 30 tests in the file pass. Full existing suite (6100+ tests) also passes,
no regressions. ESLint and Prettier clean on both changed files (pre-existing lint
warnings elsewhere in the file are untouched, per the "don't reformat outside scope"
guidance).

**Live**: built this exact branch into a local `dispatch-test-aio` instance and tested
against real IPTV streams. Observed a genuine `NetworkError HttpStatusCodeInvalid` fire
mid-playback on a live channel — it recovered silently with no visible disruption to the
viewer, confirmed via both the browser console and backend proxy logs.

## Checklist

- [x] Backend: migrations are included if any models were changed (N/A — frontend-only change)


SergeantPanda (approved):
This looked pretty good, thank you for the PR! I did make a couple of changes though.
1. **Retry budget** resets on `MEDIA_INFO` (not `LOADING_COMPLETE`). For live MPEG-TS, `LOADING_COMPLETE` is EOF, so resetting there could wipe attempts after an error and effectively unbound retries.
2. **Destroy immediately** on a retryable `NetworkError`, then wait the backoff before recreate, so the failed worker/HTTP session doesn’t sit open during the delay.
3. **Guard against stacked reconnect timers** if multiple errors fire before recreate.
4. **Set `playerRef` before `load()`** so the new stale-instance checks don’t drop sync events from the active player.

PR #1568, opened 2026-08-18
feat: auto-fallback to AAC output profile on live-player MediaError

### LINKED ISSUE #1566: [Bug]: Live web player: no reconnect on network error, no audio on AC3 sources, and a stale-event race that can undo the fix
### Describe the bug

The live-channel web player (`mpegts.js`, in `FloatingVideo.jsx`) has three related
reliability problems. Bundling them into one issue because #2 and #3 were found while
testing a fix for #1, and #3 is a bug in the recovery mechanism itself — planning two
separate PRs so each stays reviewable on its own (noted below).

**1. No reconnect on a transient network error.**
Unlike the VOD/recording player (`hls.js`, which self-heals via `startLoad()` /
`recoverMediaError()`), the live player has no recovery logic on `mpegts.Events.ERROR` at
all — it just shows a message and stops. A transient backend hiccup (even one the backend
itself already recovered from) permanently kills live playback until the user manually
reopens the channel.

**2. No audio on sources whose audio is AC3.**
Browsers' MediaSource Extensions can't decode AC3. A source using AC3 audio plays with no
sound and shows "Your browser doesn't support the codecs used in this stream." The app
already has a fix for exactly this — a locked, seeded-on-install `OutputProfile` named
"Web Player (AAC Audio)" — but the user has to know to go select it in Settings; the web
player itself doesn't fall back to it automatically. (Also confirmed this isn't limited to
raw AC3-passthrough sources — one test channel had AAC source audio but was requesting a
different OutputProfile that itself transcoded the good AAC into AC3, same symptom.)

**3. A stale-event race that can silently undo the fix for #1/#2.**
`mpegts.js` can emit an event asynchronously from a player instance that's already been
destroyed and replaced (e.g. a worker message already in flight when `.destroy()` is
called). None of the current event handlers check whether the player an event belongs to
is still the active one. Concretely, testing a fix for #2: a late, stale `NetworkError`
from an already-replaced (and already succeeding) player can destroy that working player
and recreate it with the *original*, broken configuration — silently reverting a
successful recovery. The same staleness also produces a misleading "Auto-play was
prevented" message even when a later player has already started playing successfully.

**Proposed fix** (two PRs, in sequence):
- PR 1 (targets `dev`): fixes #1, plus #3 (the stale-event guard is needed for #1's own
  destroy/recreate reconnect to be correct, independent of #2).
- PR 2 (based on PR 1, opened after): fixes #2 (the AAC auto-fallback), building on the
  guard from PR 1.

**How this was tested:**
- Automated: new Vitest tests covering backoff/retry-limit/give-up behavior, the
  AAC-fallback decision logic (including "already on that profile" and "profile not
  found" guards), and the stale-player-instance guard (using distinct mock player
  instances per `createPlayer()` call, since the existing tests all reuse one shared mock
  and can't otherwise represent "old instance vs. new instance"). Verified the new
  stale-guard tests actually catch the bug by reverting the fix locally and confirming
  they fail. Full existing suite (6100+ tests) still passes, no regressions.
- Live: reproduced all three issues and verified the fixes against real IPTV streams
  through a local `dispatch-test-aio` instance — a real `NetworkError` recovering
  silently, multiple real AC3-audio sources gaining audio automatically, and repeated
  attempts on the same channel showing a clean single profile-switch in the backend logs
  with no flapping back to the broken state.

- **#1**: hard to force deterministically (needs a real transient network blip on the
  live proxy connection) — observed live against real IPTV sources during testing.
- **#2**: open any live channel whose source uses AC3 audio (or whose selected Output
  Profile transcodes to AC3) in the web player — video plays, no audio, console shows
  `MediaError` / `MediaMSEError`.
- **#3**: only reproducible with a fix for #1 and/or #2 in place (it's a bug in the
  recovery code, not the original code) — the PR includes a unit test that reproduces it
  deterministically with distinct mock player instances.

### Dispatcharr Version

v0.29.0-dev (upstream/dev @ 74523321)

### Docker Version

29.7.2

AIO mode (`ghcr.io/dispatcharr/dispatcharr:dev`), equivalent to the official
docker-compose.aio.yml reference

### Client Information

Chrome (primary testing/reproduction), also observed in Edge. Live-verified against real
IPTV streams through a local dev instance.

jstevenscl:
## Description

Browsers' MediaSource Extensions can't decode AC3 audio. A live channel whose source (or
whose currently-selected Output Profile) produces AC3 audio plays with no sound and shows
"Your browser doesn't support the codecs used in this stream." The app already has a fix
for exactly this — a locked, seeded-on-install `OutputProfile` named "Web Player (AAC
Audio)" (`core/migrations/0024_outputprofile.py`) that transcodes to AAC — but the web
player didn't use it automatically; the user had to know to go select it in Settings.

This PR: on a `MediaError`, if the player isn't already using that profile, makes one
automatic attempt — destroy the player, look up the AAC profile's id from the
already-app-wide-loaded `useOutputProfilesStore`, rebuild the stream URL with
`output_profile=<that id>`, and recreate the player. If it's already on that profile (or
the profile can't be found), falls straight through to the existing error message — no
infinite loop, no repeated attempts.

New helpers in `FloatingVideoUtils.js`: `getOutputProfileParam(url)` /
`setOutputProfileParam(url, id)`, since `FloatingVideo.jsx` receives an already-built
stream URL from its caller (channel tables etc.) rather than constructing it itself.

Builds on #1567 (the stale-player-instance guard from that PR is what makes this
fallback's destroy/recreate reliable — without it, a late event from the pre-fallback
player could revert a successful fallback back to the broken state, which is exactly how
that bug was originally found, live-testing an earlier version of this change).

## Related Issue

Closes #1566

## How was it tested?

**Automated**: 4 new Vitest tests covering the fallback decision logic: retries once with
the AAC profile on `MediaError`, does not retry a second time if the AAC-profile attempt
also errors (no infinite loop), does not retry if already on the AAC profile, and does
not retry if the AAC profile isn't available in the store. Full existing suite (6100+
tests, including the stale-guard tests from #1567) also passes, no regressions. ESLint
and Prettier clean on all changed files (pre-existing lint warnings elsewhere in the file
are untouched, per the "don't reformat outside scope" guidance).

**Live**: built this branch into a local `dispatch-test-aio` instance and tested against
real IPTV streams with AC3 audio. Confirmed via backend proxy logs (`Stream #0:1 -> #0:1
(ac3 (native) -> aac (native))`) and audibly on-screen that audio now works automatically
on channels that previously played silently — including one case where the source audio
was already AAC-native but the currently-selected Output Profile was itself transcoding
it *into* AC3 (same symptom, confirms the fallback isn't limited to raw AC3-passthrough
sources specifically). Repeated attempts on the same channels showed a clean single
profile-switch in the backend logs with no flapping back to the broken state.

## Checklist

- [x] Backend: migrations are included if any models were changed (N/A — frontend-only change)

### DIFF
--- a/frontend/src/components/FloatingVideo.jsx
+++ b/frontend/src/components/FloatingVideo.jsx
@@ -516,17 +516,24 @@ export default function FloatingVideo() {
       if (typeof savedMuted === 'boolean') videoRef.current.muted = savedMuted;
 
       player.on(mpegts.Events.LOADING_COMPLETE, () => {
+        // mpegts.js can emit events asynchronously from a player instance
+        // that has since been destroyed and replaced (e.g. by our own
+        // reconnect below) -- ignore anything from a stale, no-longer-active
+        // player rather than let it mutate shared state.
+        if (playerRef.current !== player) return;
         setIsLoading(false);
         // Data is flowing again -- give future errors a fresh retry budget.
         reconnectAttemptsRef.current = 0;
       });
 
       player.on(mpegts.Events.METADATA_ARRIVED, () => {
+        if (playerRef.current !== player) return;
         setIsLoading(false);
         reconnectAttemptsRef.current = 0;
       });
 
       player.on(mpegts.Events.ERROR, (errorType, errorDetail) => {
+        if (playerRef.current !== player) return;
         setIsLoading(false);
 
         if (errorType === 'NetworkError' && errorDetail?.includes('aborted')) {
@@ -570,9 +577,15 @@ export default function FloatingVideo() {
       player.load();
 
       player.on(mpegts.Events.MEDIA_INFO, () => {
+        if (playerRef.current !== player) return;
         setIsLoading(false);
         try {
           player.play().catch((e) => {
+            // This player may have already been replaced by the time the
+            // play() promise settles (e.g. a reconnect fired in the
+            // meantime) -- a rejection from a now-stale player must not
+            // stomp the current, possibly-successful player's state.
+            if (playerRef.current !== player) return;
             console.log('Auto-play prevented:', e);
             setLoadError('Auto-play was prevented. Click play to start.');
           });
--- a/frontend/src/components/FloatingVideo.jsx
+++ b/frontend/src/components/FloatingVideo.jsx
@@ -3,6 +3,7 @@ import React, { useCallback, useEffect, useRef, useState } from 'react';
 import Draggable from 'react-draggable';
 import useVideoStore from '../store/useVideoStore';
 import useAuthStore from '../store/auth';
+import useOutputProfilesStore from '../store/outputProfiles';
 import mpegts from 'mpegts.js';
 import Hls from 'hls.js';
 import { CloseButton, Flex, Loader, Text, Box } from '@mantine/core';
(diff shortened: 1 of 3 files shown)


SergeantPanda:
Thank you for the PR!

I'm not sure I want this behavior. I understand it could help people but maybe we could just update the message about unsupported codecs and recommend them use an output profile.

Eventually once I'm confident that output profile is stable, I planned on making it the default but I think it still needs a bit of work (the output profile) I've seen a couple random issues with it.

PR #1569, opened 2026-08-18
Add MAC-panel credential push (IBOXX/CR7-family panels)

### LINKED ISSUE #1545: [Feature]: Push XC credentials to MAC-address-based IPTV panels (IBO Player Pro family)

Users who watch on MAC-address-based IPTV players (he IBO Player Pro family can't just enter a normal M3U/Xtream Codes URL on those devices. Instead, an admin has to open the panel's own website by hand, log the device in with its MAC address + device key, solve a captcha, and manually retype the customer's Xtream Codes host/username/password into the panel's "Add Playlist" form. That has to be repeated any time credentials change or a customer adds a second device — pure manual lookup-and-retype work with no way to do it from inside Dispatcharr.

A new admin-only feature that pushes a Dispatcharr XC user's credentials straight to one of these panels, from the Users table: pick a user, manage a list of their device MACs, and push with one click. The panel's own captcha is always relayed to the admin inside Dispatcharr's UI — never solved or bypassed programmatically, since that's the panel's own anti-bot control and automating it isn't something this should do.

I have a working implementation already (self-contained new apps/mac_panel/ app + a small frontend modal + a "Settings > MAC Panel" page for the one shared configuration value), live-tested end to end against a real panel: captcha relay, push, re-push without duplicating the panel-side playlist, and optional PIN-protected playlists all confirmed working. It only touches 3 existing files (dispatcharr/settings.py, apps/api/urls.py, one new row-action icon in UsersTable.jsx) — everything else is new files, so the diff is small and self-contained. Backend: 42 new tests, all outbound HTTP mocked. I'd like to open a PR for this if the maintainers are open to it — happy to adjust scope/approach based on feedback first.

### Additional context

The panels in question (IBOSOL-family: iboxiptvplayer.com, cr7player.com, iboplayer.com, ibovpnplayer.com, messitvplayer.com, hqplayertv.com, and reportedly more sibling domains on the same backend) expose an undocumented but stable JSON API for this — no scraping/browser automation involved, just plain HTTP calls to the same endpoints their own app uses. Happy to share more implementation detail or narrow the scope if useful.

titooo7:
## Description

Adds an admin-only feature to push a Dispatcharr Xtream Codes user's credentials directly to IBO Player Pro-family MAC-address panels (iboxiptvplayer.com, cr7player.com, iboplayer.com, ibovpnplayer.com, messitvplayer.com, hqplayertv.com, and reported sibling domains on the same backend), instead of the current manual lookup-and-retype workflow.

- New self-contained `apps/mac_panel/` app: panel client, credential push/re-push logic (no duplicate playlists on re-push), optional PIN-protected playlist support, and API views/serializers/models.
- Frontend: a device-management modal on the Users table row action, plus a new "Settings > MAC Panel" page for the one shared configuration value.
- The panel's own captcha is always relayed to the admin inside Dispatcharr's UI and solved by hand — nothing here solves or bypasses it programmatically.
- Small, self-contained diff: only 3 existing files touched (`dispatcharr/settings.py`, `apps/api/urls.py`, `frontend/src/components/tables/UsersTable.jsx` — one new row-action icon); everything else is new files.
- No scraping/browser automation — plain HTTP calls to the same undocumented-but-stable JSON API the panels' own apps use.

## Related Issue

Closes #1545

## How was it tested?

- 42 new backend tests (all outbound HTTP mocked) — passing
- 96 frontend tests — passing
- Verified against a pristine v0.29.0 checkout with no other fork code present
- Live-tested end to end against a real panel: captcha relay, push, re-push without duplicating the panel-side playlist, and PIN-protected playlists all confirmed working
- ESLint and Prettier verified clean on every file this PR adds or touches

## Checklist


SergeantPanda:
Thank you for the PR for this but I think we will pass on this feature.

None of the devs would have a way to maintain/test this code and it is pretty specific to a feature that isn't widely used afaik. 

I suspect this would be quite possible with a plugin though so that may be a better option here (see the Stalker portal plugin for inspiration)

PR #1572, opened 2026-08-19
fix(dvr): dedup an identifiable airing against an identity-less booking

v8eta:
## Description

Follow-up to #1506, as requested — verified the gap still exists on `dev` after that merge.

`_evaluate_series_rules_locked` wires episode identity and the start-time window as an if/else, so an airing that **has** an identity never consults the window:

```python
if prog_ident:
    if (tvg_id, title, prog_ident) in existing_episode_keys: continue
else:
    ... start-time window ...
```

The comment justifies that with "an identifiable episode that is not already scheduled is a genuinely different airing". That holds only if every *already-scheduled* recording can be identified, and it can't. A recording whose snapshot predates the `season`/`episode` fields returns `None` from `_identity_of()`, contributes nothing to `existing_episode_keys`, and lands in `existing_program_index` instead — which the identity branch never reads.

The gap is reachable for as long as any pre-upgrade booking is still in the schedule, so roughly one DVR horizon after the release ships.

### What it did in production

The EPG reissued its programme IDs and nudged starts by ~30s. A news bulletin came back with identity `s26e218`, missed the episode-key lookup (the existing booking's `program.season` was `None`), skipped the window by design, and was scheduled a second time 30s away. Both bookings resolved to the same output path, and at recording-end two concurrent `ffmpeg -y -f concat` remuxes wrote the same file — a 3900s recording ended up as 1125s of unusable output. A second series hit the identical gap the same night; its old booking also had `season=None` and its `sub_title` differed, so even the sub-title fallback would have missed.

### The change

Run the window after an episode-key **miss**, regardless of whether the candidate is identifiable:

```python
if prog_ident and (
    (str(prog.tvg_id), prog_title_l, prog_ident) in existing_episode_keys
):
    continue
# window runs unconditionally from here
```

`existing_program_index` is still populated only from identity-less recordings, so the property the if/else was protecting is preserved exactly: **an identifiable episode still cannot be suppressed by another identifiable one**. It can now be suppressed by an identity-less booking of the same channel+title inside the window — which is the ambiguous case, and the one that corrupts files.

The live-EPG check and one-claim-per-orphan rule added in #1506 both still apply to this path, so a still-listed slot is not claimed and one stale booking cannot swallow several candidates. Those make this strictly safer than when we first hit it.

One comment is also corrected: the window's opening line said "No identity at all" and claimed it "can never suppress an identifiable episode". Both stop being true once the window runs unconditionally, and a comment contradicting its code is how the original gap survived review.

## How was it tested?

**Verified the bug still reproduces after #1506 merged**, which was the precondition you asked for. Against pristine `dev` (plus `managers.py`, since the image predates the `epg_source_id` work) the new test fails and the other 14 in that file pass — so it isolates this gap and nothing else:

```
FAIL: test_identifiable_airing_dedups_against_identity_less_booking
Ran 15 tests in 1.535s
FAILED (failures=1)
```

With the fix applied, the full dedup surface passes:

```
Ran 44 tests in 5.097s
OK
```

(15 in `test_series_rule_dedup_time_drift` + 29 in `test_series_rule_dedup`, zero regressions.)

Two tests added:

- `test_identifiable_airing_dedups_against_identity_less_booking` — the production scenario: a booking whose snapshot has had identity stripped, then the listing returns 30s later carrying `s26e218`. Must not double-book.
- `test_identifiable_airing_not_suppressed_by_identifiable_booking` — the guard for the property above: two different identifiable episodes 10 minutes apart must both record.

Also running the same change on the instance where the original corruption happened.

## Checklist

- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — no API changes)
- [x] Frontend: ESLint and Prettier pass cleanly (N/A — backend only)


SergeantPanda (approved):
This looks good. Thank you! I did shorten some of the comments (I prefer only relevant comments that explain non obvious code to be left in the codebase.

I also modified some comments that were no longer correct after this change.

Issue #1518, opened 2026-08-05
[Feature]: Add original air date placeholder to TV fallback recording templates


nilleiz:
### Is your feature request related to a problem?

The programme detail API/UI already exposes the EPG's canonical original-air date, but DVR fallback path templates cannot use that value.

This is especially visible with series rules. A one-off recording may still resolve the current ProgramData row when it starts, while a series-scheduled recording can remain queued across an EPG refresh that deletes and recreates that row. The original-air date is then unavailable when the fallback filename is rendered.

As a result, an English, Plex-compatible fallback name such as:

```text
TV/Example Show/Example Show - 2016-12-15 - The Missing Witness.mkv
```

cannot be produced consistently for both one-off and series-scheduled recordings.

### Describe the solution you'd like

Add an `{original_air_date}` placeholder specifically to the TV fallback DVR path template used when a programme has no usable season or episode number.

The value should:

- come from the canonical XMLTV `previously-shown start` metadata already stored as `ProgramData.custom_properties["previously_shown_details"]["start"]` and exposed by the programme API as `original_air_date`;
- normalize supported date/timestamp forms to `YYYY-MM-DD`;
- be persisted in queued series-recording payloads;
- be recoverable by stable channel/TVG identity and exact programme times if an EPG refresh replaces the original `ProgramData` row;
- resolve to an empty value when canonical original-air metadata is unavailable, without substituting the programme start date, recording date, or current year;
- be advertised only for the TV fallback template, leaving normal TV and movie templates unchanged.

Acceptance criteria:

- One-off and series-scheduled recordings render the same canonical original-air date.
- Series recordings retain or recover the value across an EPG refresh.
- Missing metadata does not raise an error or invent a date.
- Normal TV, movie, and movie-fallback templates are unaffected.
- Nonstandard values are sanitized before entering a filesystem path.
- Backend and frontend regression tests cover the behavior.

### Additional context

Related to #682, which requests separated date components for fallback recordings such as sports programmes. This request is intentionally narrower and semantically different: it exposes the EPG's **original air date** only. It does not provide a broadcast/recording-date fallback when original-air metadata is absent.

A tested implementation is available in `nilleiz:feature/dvr-original-air-date`.

sethwv:
This overlaps with #682 and #1307, linking them so discussion stays in one place, but leaving this open since it's not quite the same request.

nilleiz:
[...]

Issue #862, opened 2026-01-16
[Feature]: EPG tag management


mwongj:
### Is your feature request related to a problem?

As a user when I view a tv guide that displays tags I want tags to have consistency across programs.

Different tags can cause downstream programs to behave differently. I think there may be a few situations that introduce inconsistencies:
1. Source epg uses one tag. A source may only use "New" for all new programs
2. Source may believe 2 tags are semantically similar or simply looser definitions for tags
3. Source may correctly use tags

### Example tags

|Tag|Definition|
|:--|:--| 
|New|First-run episode or an episode that hasn't aired on that specific channel recently|
|Premiere|The absolute first time this show/episode airs on any channel within the EPG's data|

### Describe the solution you'd like

I think this feature could be one part of a larger feature:
- Phase 1: Tag mapping
  - Phase 1.1: Static number of mappings
  - Phase 1.2: Dynamic number of mappings
- Phase 2: Tag removal
- Phase 3: Tag addition

### M3U & EPG Manager > EPG Source

- Add a static number of mapping rows (2, for example). Each row will contain 2 select inputs. The left select input will be "Source tag" and the right select input will be "Destination tag"

### Additional context

For dummy EPGs there are checkboxes that apply live and new tags if checked so there is some tag management already. Depending on how phase 3 is implemented it may apply to both types of epg sources

sethwv:
1. What “tags” are you seeing in your EPG source XMLTV? Are they coming from `<category>`, `<keyword>`, `<new/>`, or something else?

2. Should tag mapping affect only the Dispatcharr guide display, only the exported XMLTV output, or both?

3. Should “Destination tag” be a fixed list (New/Live/Premiere/etc.) or free-text?

4. When a mapping matches, should we replace the source tag with the destination tag, or add the destination tag and keep the original too?

5. Should mapping be case-insensitive / trimmed (treat `new`, `New`, ` NEW ` as the same)?

6. Can you share 2–3 example mappings you want (e.g., `Premiere → New`, `First Run → New`)?

jesmannstl:
[...]

PR #1573, opened 2026-08-19
fix(dvr): series rule mode "new" honours previously-shown, not just new

v8eta:
## Description

A `mode="new"` series rule accepts a programme only when the EPG explicitly tagged it `<new/>`:

```python
if (p.custom_properties or {}).get("new"):
```

XMLTV also lets a feed mark **repeats** with `<previously-shown/>` instead, and a feed using that convention exclusively never sets `<new/>` on anything. On such a feed a `mode="new"` rule matches **nothing at all**.

The failure mode is what makes this worth fixing rather than documenting: it is completely silent, and it is hard to attribute. A rule that matches nothing looks exactly like a series that is not currently airing — no error, no warning, just an empty schedule. On the instance where this was found, a rule had been sitting in `mode="new"` for **months** with **zero** recordings, and it only came to light while investigating something unrelated.

Measured across the four EPG feeds in use there, the `new` key is absent from **every** programme, while repeats consistently carry `previously_shown`. A representative day for one series, where the split is exact:

| airing | `previously_shown` | |
|---|---|---|
| 19:00–19:31 | absent | first run |
| 12:00 / 14:10 / 15:20 | present | repeat |

### The change

Treat a programme as new when it is tagged `<new/>`, **or** when it is not tagged `<previously-shown/>`:

```python
props = p.custom_properties or {}
if props.get("new") or not props.get("previously_shown"):
```

Feeds that do tag `<new/>` are unaffected for those programmes. A programme carrying neither tag now counts as new, which errs toward recording — the safe direction for a DVR, where an extra capture costs disk but a missed one is unrecoverable. That also matches how the two tags are defined: `<previously-shown/>` is the positive assertion that something is a repeat, so its absence is the better signal when `<new/>` is not in use.

## Related Issue

None open that I could find — I searched issues and PRs for both `previously_shown` and the `mode: new` behaviour before writing this. Closest existing items are #862 (EPG tag management) and #1518 (original air date placeholder), neither of which covers this.

## How was it tested?

Five tests in `apps/channels/tests/test_series_rule_new_mode.py`, reusing the existing `SeriesRuleDedupBaseTestCase` fixture:

- `test_new_tag_still_recorded` — the existing convention keeps working
- `test_previously_shown_is_skipped` — a marked repeat is still excluded
- `test_untagged_first_run_is_recorded` — the bug: previously skipped, so nothing recorded at all
- `test_mixed_schedule_records_only_first_runs` — the real shape of an affected feed: evening first runs plus a daytime repeat, expects exactly the first runs
- `test_new_tag_wins_over_previously_shown` — an explicit `<new/>` stays authoritative alongside a previously-shown date

**Against pristine `dev` the two untagged-first-run tests fail and the other three pass**, so they isolate this behaviour and nothing else:

```
FAIL: test_mixed_schedule_records_only_first_runs
FAIL: test_untagged_first_run_is_recorded
Ran 5 tests in 0.449s
FAILED (failures=2)
```

With the change, the new tests plus both existing series-rule suites pass:

```
Ran 47 tests in 4.818s
OK
```

(5 new + 29 `test_series_rule_dedup` + 13 `test_series_rule_dedup_time_drift`, zero regressions.)

Also verified against the live EPG on the affected instance: with the change, the rule selects the 5 first-run airings and skips the 7 repeats, matching the `previously_shown` split exactly.

### Addendum: the preview endpoint has its own copy of this filter

The Edit Series Rule dialog's preview (`api_views.py`) filters candidates independently of the scheduler:

```python
if mode == "new":
    candidates = [p for p in candidates if (p.custom_properties or {}).get("new")]
```

so fixing only `tasks.py` produced a UI that disagrees with reality — the dialog shows **"0 OF 0 — No matching upcoming programs in the next 7 days"** for a rule the scheduler would happily record. The second commit applies the same `new`-or-not-`previously_shown` logic there, and widens `is_new` in the response identically so the per-row flag agrees with the filter that selected it.

Verified against the same live feeds: a title with three untagged upcoming airings previewed as 0 matches before, 3 after — matching what the scheduler selects.
## Checklist

- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — no API changes)
- [x] Frontend: ESLint and Prettier pass cleanly (N/A — backend only)

### DIFF
--- a/apps/channels/api_views.py
+++ b/apps/channels/api_views.py
@@ -4278,7 +4278,15 @@ class SeriesRulePreviewAPIView(APIView):
         # over the bounded result set we already filtered down to.
         candidates = list(qs[:limit * 4])  # small overshoot to allow new-only filtering
         if mode == "new":
-            candidates = [p for p in candidates if (p.custom_properties or {}).get("new")]
+            # PATCH 13: preview previously_shown. Mirrors the scheduler filter
+            # in _evaluate_series_rules_locked: XMLTV feeds may mark first runs
+            # with <new/> OR mark repeats with <previously-shown/>, and testing
+            # only the former hides every programme on a feed using the latter.
+            candidates = [
+                p for p in candidates
+                if (p.custom_properties or {}).get("new")
+                or not (p.custom_properties or {}).get("previously_shown")
+            ]
 
         total = len(candidates)
         candidates = candidates[:limit]
@@ -4296,7 +4304,8 @@ class SeriesRulePreviewAPIView(APIView):
                 "end_time": p.end_time.isoformat(),
                 "season": cp.get("season"),
                 "episode": cp.get("episode"),
-                "is_new": bool(cp.get("new")),
+                "is_new": bool(cp.get("new"))
+                or not cp.get("previously_shown"),
             })
 
         return Response({
--- a/apps/channels/tasks.py
+++ b/apps/channels/tasks.py
@@ -751,10 +751,18 @@ def _evaluate_series_rules_locked(tvg_id, result):
 
         # Optionally filter to only brand-new episodes before grouping
         if mode == "new":
+            # PATCH 12: previously_shown. XMLTV allows a feed to mark first runs
+            # with <new/> OR to mark repeats with <previously-shown/>. Testing only
+            # the former makes this rule match NOTHING on a feed that uses the
+            # latter - which is every feed here (measured 2026-08-19: the "new" key
+            # is absent on all programmes across all four auepg feeds, while repeats
+            # carry "previously_shown"). Neither tag counts as new, erring toward
+            # recording: an extra capture costs disk, a missed one is unrecoverable.
             filtered = []
             for p in programs:
                 try:
-                    if (p.custom_properties or {}).get("new"):
+                    props = p.custom_properties or {}
+                    if props.get("new") or not props.get("previously_shown"):
                         filtered.append(p)
                 except Exception:
                     pass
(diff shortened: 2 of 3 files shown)


v8eta:
Pushed a second commit: the Edit Series Rule dialog's **preview** endpoint keeps its own copy of the freshness filter (`api_views.py`), so the first commit fixed the scheduler while the dialog still reported "0 OF 0 — No matching upcoming programs" for a `mode=new` rule that would in fact record.

Found from a screenshot of exactly that: a rule on a feed where all 244 upcoming programmes carry neither tag, with three matching airings hidden by the preview. After the change the preview reports the same three the scheduler would take, and `is_new` in the response is widened identically so the per-row flag agrees with the filter that selected it.

Updated description follows in the body edit.

SergeantPanda:
Thank you for the PR.

Unfortunately I don't think we want to accept this as written.

Treating the lack of `previously_shown` as evidence that an airing is new just shifts the problem the other way: feeds that only stamp `<new/>` on first runs (and leave repeats untagged) would start recording everything under `mode=new`.

XMLTV defines both `<new/>` and `<previously-shown/>`, but they are optional and mean different things. Absence of one is not presence of the other, and EPG sources are inconsistent about which they use. We prefer not to collapse those into a single default heuristic.

I could see a toggle that lets a user treat “not previously shown” as new, but it should not be the default. Let me know if you'd like to rework the PR for that instead.

Thank you again!

v8eta:
Reworked as an opt-in per-rule toggle as suggested — #1609. Default is exactly the current behaviour; the checkbox only appears for "New only" rules. Thanks for the review.

Issue #1531, opened 2026-08-11
[Bug]: Recording metadata (season/episode/sub-title) frozen at scheduling time, never re-synced against later EPG updates


jamesgallagher:
### Describe the bug

When a Recording is created (via a series rule or manually), Dispatcharr captures a one-time snapshot of the EPG programme's metadata — title, sub_title, description, plus (via a separate prefetch_recording_artwork task that fires ~1 second later) season/episode/onscreen_episode — and stores it in Recording.custom_properties. This snapshot is never refreshed again for the lifetime of that recording.

This causes two related but distinct symptoms:

Stale-ID case: EPG refreshes delete and bulk-recreate ProgramData rows rather than updating them in place. If a refresh happens between scheduling and recording completion, the ProgramData.id captured in the snapshot no longer resolves to anything, and at file-naming time _parse_epg_tv_movie_info() gets None for season/episode — silently triggering the TV Fallback Template instead of the TV Path Template, even though the EPG source has (and always had) complete data for that airing.
Incomplete-snapshot case (the more general form): many EPG sources — especially daily/rolling-window guides — publish a bare placeholder entry for a programme days in advance and backfill full metadata (sub-title, episode numbering) closer to air date. Since Dispatcharr's snapshot is taken the moment the recording is scheduled, it can capture that early bare-bones version and never pick up the later, richer data — even when no ProgramData.id staleness is involved at all.

Both stem from the same root cause: recording metadata is captured once, early, and never re-synced against the EPG before the recording completes and its output filename is generated.

### How can we recreate this bug?

1. Set up a series rule (or manual recording) for a show whose EPG source publishes progressively more detailed metadata as air date approaches (common for daily shows).
2. Let the recording get scheduled well before the EPG source has published full season/episode/sub-title data for that specific airing.
3. Let at least one more EPG refresh occur before the recording completes, after the source has since published fuller metadata for that same slot.
4. Check the resulting filename/metadata — it reflects whatever was known at scheduling time, not the fuller data now present in the EPG.

### Dispatcharr Version

v0.29.0

### Docker Version

Docker version 27.5.1, build 9f9e405

### What's in your Docker Compose file?

No using

### Client Information

TiViMate

Patch added but note, this is Claude AI slop, it works for me, but may not be ideal

[season-episode-stale-id-fix.patch](https://github.com/user-attachments/files/30919682/season-episode-stale-id-fix.patch)

sethwv:
[...]

PR #1575, opened 2026-08-20
fix(dvr): keep season/episode when the EPG row is reissued (#1307)

### LINKED ISSUE #1307: [Bug]: DVR not using Season or Episode from EPG to name files, keeps falling back to fallback template.
### Describe the bug

I've tried recording several tv show episodes using a "series" job and they keep using the fallback template, ignoring Season and Episode data that is there. DVR scheduled recordings show the info, tv guide shows the info, but the DVR task never uses it to name the file. Ex EPG looks like this:

```
  <programme start="20260530060300 +0000" stop="20260530063200 +0000" channel="19">
    <title>American Dad!</title>
    <desc>What Great Advancements!
An aspiring inventor leaves the farm to make a go of it in the big city.</desc>
    <episode-num system="onscreen">E22</episode-num>
    <episode-num system="xmltv_ns">18.21.</episode-num>
  </programme>
```

1. Schedule a recording for a Series or a show over 24hrs into the future.
2. Check the name used after the recording is complete. 

### Dispatcharr Version

v0.25.1

### Docker Version

Docker version 28.2.2, build 28.2.2-0ubuntu1~24.04.1Docker version 28.2.2, build 28.2.2-0ubuntu1~24.04.1

services:
  dispatcharr:
    # build:
    #   context: .
    #   dockerfile: Dockerfile
    restart: unless-stopped
    container_name: dispatcharr
    ports:
      - 9191:9191
    volumes:
      - ./data:/data
    environment:
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
      #- USE_LEGACY_NUMPY=true
    #
    cap_add:
      - SYS_NICE
    #devices:
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities: [gpu]

### Client Information

Web/VLC

v8eta:
## Description

`_parse_epg_tv_movie_info` resolves season/episode by re-querying the **live** `ProgramData` row:

```python
epg_program = ProgramData.objects.filter(id=prog_id).only("custom_properties").first()
```

EPG refreshes delete and bulk-recreate `ProgramData` rows rather than updating them in place, so the id captured when a recording was scheduled routinely resolves to nothing by the time `_build_output_paths` runs. `season`/`episode` come back `None`, and the caller silently falls through to the S/E-less fallback template:

```
TV_Shows/{show}/{show} - {year}-{month:02d}-{day:02d}.mkv
```

so the recording is named by air date rather than `SxxExx`. There is no error and no log line — a correctly configured TV path template just quietly stops being used.

On my install this was **every** show: all 8 with recordings had the live row gone, while every one still carried its season/episode elsewhere on the `Recording`. The identity was present the whole time and was being discarded.

**Why it matters beyond the filename.** Plex's TV scanner takes the episode index from the *filename*. A file with no `SxxExx` becomes a date-based episode with empty `parentIndex`/`index`, and Plex's NFO agent skips index-less episodes entirely — so those recordings never receive sidecar metadata no matter what is written beside them.

### The fix

Read the identity from the two places that survive the refresh, cheapest first:

| Source | Why it survives |
|---|---|
| `custom_properties["program"]` — the booking snapshot | Captured at scheduling time; independent of the row's later fate |
| `custom_properties` on the `Recording` itself | `prefetch_recording_artwork` writes `season`/`episode` there ~1s after booking, independently of `ProgramData` |

The second source is needed on its own: recordings whose snapshot predates the artwork prefetch have `program.season = None` while the `Recording` carries e.g. `26/218`. I only found that because a first cut using the snapshot alone *still* produced date names for older recordings.

Programmes that genuinely carry no episode numbering anywhere are unaffected and still reach the fallback template — there is a test for that, since forcing them to a bogus `S00E00` would be worse than the date name.

## Related Issue

Closes #1307

Also covers the root-cause analysis posted in #1531 (closed into #1307 as a duplicate) — specifically its "stale-ID case".

## How was it tested?

New `apps/channels/tests/test_recording_output_path.py`, 7 tests:

- snapshot identity is used when the live row has been deleted
- `onscreen_episode` (`S03E07`) is parsed out of the snapshot
- recording-level `custom_properties` supply the index when the snapshot has none
- both routes reach the actual filename via `_build_output_paths`
- a live `ProgramData` row still takes precedence over the snapshot
- no identity anywhere still yields the fallback template, with no invented index

**Ran against pristine `dev` first, to confirm the tests fail for the right reason** — 4 of the 7 failed:

```
FAIL: test_snapshot_used_when_row_deleted
AssertionError: Tuples differ: (None, None) != (26, 238)

FAIL: test_onscreen_episode_parsed_from_snapshot
AssertionError: Tuples differ: (None, None) != (3, 7)

FAIL: test_snapshot_identity_reaches_the_filename
AssertionError: 'S03E07' not found in '20260820_070122.mkv'

FAIL: test_recording_properties_supply_missing_index
AssertionError: 'S26E238' not found in '20260820_070122.mkv'
```

With the fix applied, 7/7 pass.

Regression sweep — `test_recording_metadata`, `test_recording_pipeline`, `test_recording_extend`, `test_dvr_port_resolution`, `test_series_rule_dedup`, plus the new file:

```
Ran 114 tests in 4.941s
OK
```

Also verified on the live install: after patching, 6 of 8 shows resolve `SxxExx` (the other two have no episode numbers in the EPG at all, which is correct), and Plex went from 1 to 31 of 37 episodes showing a real title and summary.

## Checklist

- [x] Backend: migrations are included if any models were changed — no models changed
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema — no new endpoints
- [x] Frontend: ESLint and Prettier pass cleanly — no frontend changes


SergeantPanda (approved):
This looks good. I made a couple of small tweaks but overall solid. Thank you!

PR #1592, opened 2026-08-21
fix: VOD episode stream 500 crash on stale/missing episode relation

jstevenscl:
## Description

`stream_xc_episode`'s `.filter(...).first()` returns `None` on no match -- it never raises `M3UEpisodeRelation.DoesNotExist`, so the `try/except` wrapping it was dead code. A stale or missing `episode_id` (e.g. a client's cached episode list from before the account's VOD catalog was refreshed) fell through to `episode_relation.episode.uuid` on `None` and crashed with an unhandled `AttributeError` instead of returning a clean 404.

This is easy to hit in practice whenever a client (e.g. TiviMate) has cached an episode list and the backing account's VOD catalog is later refreshed with different internal IDs -- every subsequent request from that client's stale cache 500s instead of failing cleanly, and some clients can't recover from that without a manual reload/restart.

**Fix**: replace the dead `try/except` with a direct `if not episode_relation` check -- the same pattern `stream_xc_movie` (immediately above this function in the same file) already uses.

Split out from #1514 at the maintainer's suggestion. That PR also tried to make VOD playback IDs stable across a catalog refresh by switching to the `M3U*Relation` PK, which was correctly pushed back on (exposing relation PKs to XC clients undercuts the guarantee Dispatcharr's trying to make; the root cause -- why canonical IDs churn on refresh in the first place -- belongs in the ingest/dedup/cleanup layer instead). This PR is only the crash fix, none of that.

## Related Issue

Related to #820 and #1513 (same crash signature; #1513 was an earlier, narrower attempt at just this fix, closed in favor of the broader #1514 -- this PR is that same narrow scope, re-submitted standalone per the maintainer's request on #1514).

## How was it tested?

**Automated**: 2 new tests in `apps/proxy/vod_proxy/tests/test_stream_xc_episode_404.py` -- a missing/stale relation returns a clean 404 (the regression case), and a valid relation still resolves and streams correctly (happy path unchanged). Verified the new test actually catches the regression: temporarily reverted the fix locally and confirmed the test fails with the exact original traceback (`AttributeError: 'NoneType' object has no attribute 'episode'` at the pre-fix line), then restored the fix and confirmed both new tests pass.

Ran the full `apps.proxy.vod_proxy.tests` suite against the project's own `base-dev` CI image and `scripts/ci_bootstrap_backend.sh` (internal Postgres + Redis, `manage.py test`):
```
docker run --rm --entrypoint '' \
  -e DISPATCHARR_ENV=aio -e DJANGO_SECRET_KEY=ci-test-secret-key \
  -e POSTGRES_DB=dispatcharr -e POSTGRES_USER=dispatch -e POSTGRES_PASSWORD=secret \
  -v <repo>:/app -w /app \
  ghcr.io/dispatcharr/dispatcharr:base-dev \
  bash scripts/ci_bootstrap_backend.sh apps.proxy.vod_proxy.tests -v2
```
Result: **55 tests, all passing** (53 pre-existing + 2 new), no regressions.

## Checklist

- [x] Backend: migrations are included if any models were changed (N/A -- no model changes)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A -- no new endpoints, existing endpoint's error path only)
- [x] Frontend: ESLint and Prettier pass cleanly (N/A -- backend-only change, no frontend files touched)


SergeantPanda:
This looks good. I did remove a lot of the overly verbose comments (we try to have only comments that are for non obvious code).

Thank you!

PR #1593, opened 2026-08-21
feat(frontend): make the global error boundary more useful

sethwv:
# Summary
Replaces the frontend's minimal one-line error boundary fallback with a diagnostic recovery panel: the top-level `ErrorBoundary` now wraps the full application, and section-scoped boundaries around modals, grid cells, and plugin cards use a container-scoped, floating fallback that stays sized to its context, doesn't block interaction with the rest of the page, and avoids duplicate `<main>`/`<h1>` landmarks. Instead of the old one-liner, the panel builds a diagnostic report containing the error and component stacks, route, build mode, and browser info, redacts tokens/bearer values and the page origin, and symbolicates minified production stack traces client-side against hidden sourcemaps (lazy-loaded, so it never ships in the main bundle) before the report is copied, so a support request is readable without a manual decode step. The panel also offers a copy button with inline status feedback and a small X to dismiss it.

## Example(s)
image
image


SergeantPanda (approved):
Looks good to me.

PR #1601, opened 2026-08-24
feat(logs): in-app log file browser [logging 2]

nagelm:
## Description

Adds an in-app log browser so operators can read Dispatcharr's log without shell access to the container. This is the second half of a split: [logging 1] produces the files, this PR reads them.

- **An API** to list the collected files, read one, and download it: `GET /api/core/logs/`, `GET /api/core/logs/{name}/`, `GET /api/core/logs/{name}/download/`.
- **Two pages** — a list of log files, and a viewer with severity colouring, wrapping, filtering, search and auto-refresh — reached from **System → Logs** in the sidebar.

### Non-obvious decisions

- **Every path resolves through one containment gate.** `_resolve()` joins the requested name onto the log dir and compares `os.path.realpath` of both, so a symlink pointing outside the directory is rejected. The *list* endpoint resolves through the same gate rather than trusting `os.listdir`, so an escaping symlink is never shown, let alone opened.

- **Only the collector's own file family is served.** An operator who points `DISPATCHARR_LOG_DIR` at a data root cannot turn these endpoints into a general file reader; the collector's own `collector.conf` and `collector.pid` are not served either. The family's base name is imported from the collector, so a rename in one place cannot silently empty the page. A console-only install, where settings could not create the log directory, reports no files rather than falling back to a default path.

- **Tails are hard-capped** at 5 MiB and read by seeking to `size - cap`, so opening a multi-gigabyte log streams the tail rather than loading the file. The response flags `truncated` so the UI can say so.

- **Size and identity come from the open file handle, never the path.** The collector rotates by rename, so `os.path.getsize(path)` followed by a separate `open(path)` can straddle a rotation: the stale size then seeks past the end of a fresh, near-empty file and the viewer shows an empty log. `os.fstat(f.fileno())` makes both come from one handle. A file the collector prunes between resolution and open answers 404 rather than 500.

- **Auto-refresh fetches only what was written since the last poll.** A full re-read was about a megabyte a second per open viewer, and a re-parse of every retained line, to show a couple of hundred new ones. The tail endpoint answers a JSON envelope of `content`, `cursor`, `reset` and `truncated`; a client holding a cursor of `inode-offset` sends it back as `?cursor=` and is served only the bytes past it, with `reset` saying whether to append or replace. The envelope goes through the shared request helper like every other endpoint, is declared in the OpenAPI schema, and needs no CORS exposure for the Vite dev server, which custom headers would have.

  The offset alone would be unsafe. Byte *n* of the rotated-away file and byte *n* of the new one are unrelated, so resuming there either skips content and presents it as continuous, or freezes the view. The inode proves the file is the same one; and because rotation frees inodes for reuse, an offset that does not sit on a line boundary is treated as proof it came from somewhere else, as is an offset beyond the end of the file, which an append-only file cannot produce. Four cases override the client and reset it: a rotation, an offset past the end, a gap wider than the view cap (a tab that slept must not be handed half a gigabyte), and no cursor at all. A line still being written is withheld until it is complete, on fresh loads and deltas alike, so every cursor sits on a line boundary and the client never sees a torn line.

- **Downloads go over the authenticated session**, not a signed URL, so no secret ever rides in a URL and the route is plainly `IsAdmin` like the others. The cost is that the whole body is buffered before the browser saves it, so the button shows its loading state for the duration rather than appearing dead.

- **The viewer follows the log's live edge.** Auto-refresh replaces content while `scrollTop` stays fixed in pixels, so a reader watching the newest lines would otherwise drift backwards through history as the log grew. The view re-pins only for a reader already at the edge; someone who scrolled back to read history is left where they were.

- **The viewer's auto-refresh pauses when the tab is hidden**, and pauses after three consecutive failed polls without touching the stored interval, so a forgotten open tab does not poll forever and a restart does not rewrite the reader's preference; choosing an interval again resumes. A response is discarded if a newer request has superseded it, since two overlapping loads carry the same cursor and would otherwise append the same delta twice.

- **The level floor gates the levels the collector invents** for unlabelled output. The floor exemption belongs to the sinks, where gating is permanent data loss; the viewer is a reversible filter over a file that already has everything, and a Warning floor that still showed the whole boot narration read as broken rather than protective.

- **The Logs page hides itself where there is nothing to browse.** A deployment with no collector has no files, so the nav entry declares the flag it requires and the ordering helper drops it the way it drops DVR and VOD; the page itself still explains the situation if reached directly.

## Related Issue

N/A

## How was it tested?

Built the image from this branch and ran a container against a fresh volume (`TZ=Pacific/Auckland`), then:

**Frontend:** `npx vitest run` — **6405 tests / 209 files pass**. This branch's own two page files contribute 63 of them (`LogFileView.test.jsx` 54, `LogFiles.test.jsx` 9): classification, grouping of continuations, ordering, filtering, search debounce, refresh backoff, the pause leaving the stored interval alone, hidden-tab pause, tail-following, incremental append versus reset, a traceback split across two polls, overlapping loads, the trailing newline never rendering as a blank row, a torn offset never parsing as a level, a failed listing reported as such, and a 4 MB payload render. The navigation and sidebar tests cover the entry disappearing without a collector.

**Backend:** full suite in the container — `core.tests` **199 tests, OK**. `core/tests/test_log_files.py` contributes 22: listing, tailing, the truncation flag, rejection of symlinks escaping the log dir, traversal attempts, the log-family filter, sizing from the open handle across a rotation, a file pruned after resolution, a console-only install, exact 401 and 403 codes on every route, and the cursor's cases — resumption, a withheld partial line on both a fresh load and a delta, a rotation, a gap past the cap, an offset past the end of the same inode, a mid-line offset, and malformed cursors including a Unicode digit and one longer than `int()` accepts.

Each new test was checked to fail against the unfixed code rather than pass vacuously. The rotation-race test fails with `'' != 'new\n'` through the real endpoint; the rotation-cursor test fails if the identity check is dropped; the split-traceback test loses the tail out of its block if the parser's carried state is dropped; the overlap test renders the delta twice without the supersession guard; the past-the-end cursor test receives an empty non-reset delta, and the Unicode-cursor test a 500, against the clamped and `isdigit()` code; the trailing-newline test finds a `line-height: 0.35` row; the pause test finds `0` written to storage.

**API, over HTTP against the running container:**
- `GET /api/core/logs/` → 200, listing the collector's files with sizes.
- `GET /api/core/logs/dispatcharr.log/` → 200, JSON with the tail, `reset: true` and a cursor.
- The same with `?cursor=` → 200, `reset: false`, only the bytes written since.
- Truncating the live file in place under the running collector, then polling with the old cursor → 200, `reset: true` with the new head; the next poll resumes from the new end.
- `GET /api/core/logs/collector.conf/` and `.../collector.pid/` and `.../passwd/` → **404**, confirming the family filter.
- Anonymous → **401**; a non-admin session → **403**.

## Checklist

**ESLint and Prettier.** Verified as "this PR introduces no new problems": eslint reports the same counts on this branch as on stock `dev`, measured by extracting the committed frontend blobs and running the same binary against them. Prettier is clean on every file this PR adds and on every clean file it touches. `navigation.js`, `Sidebar.jsx` and `ConnectLogsSection.test.jsx` are not Prettier-clean on `dev` either and have been left that way rather than reformatted; their diffs are the intended lines only.

### DIFF
--- a/core/api_urls.py
+++ b/core/api_urls.py
@@ -14,6 +14,12 @@ from .api_views import (
     TimezoneListView,
     get_system_events
 )
+from .log_files import (
+    get_log_file,
+    list_log_files,
+    get_log_download_token,
+    download_log_file,
+)
 
 router = DefaultRouter()
 router.register(r'useragents', UserAgentViewSet, basename='useragent')
@@ -27,5 +33,9 @@ urlpatterns = [
     path('rehash-streams/', rehash_streams_endpoint, name='rehash_streams'),
     path('timezones/', TimezoneListView.as_view(), name='timezones'),
     path('system-events/', get_system_events, name='system_events'),
+    path('logs/', list_log_files, name='log_files'),
+    path('logs/<str:name>/download-token/', get_log_download_token, name='log_file_download_token'),
+    path('logs/<str:name>/download/', download_log_file, name='log_file_download'),
+    path('logs/<str:name>/', get_log_file, name='log_file'),
     path('', include(router.urls)),
 ]
--- a/core/api_urls.py
+++ b/core/api_urls.py
@@ -17,7 +17,6 @@ from .api_views import (
 from .log_files import (
     get_log_file,
     list_log_files,
-    get_log_download_token,
     download_log_file,
 )
 
@@ -34,7 +33,6 @@ urlpatterns = [
     path('timezones/', TimezoneListView.as_view(), name='timezones'),
     path('system-events/', get_system_events, name='system_events'),
     path('logs/', list_log_files, name='log_files'),
-    path('logs/<str:name>/download-token/', get_log_download_token, name='log_file_download_token'),
     path('logs/<str:name>/download/', download_log_file, name='log_file_download'),
     path('logs/<str:name>/', get_log_file, name='log_file'),
     path('', include(router.urls)),
--- a/core/log_files.py
+++ b/core/log_files.py
@@ -43,7 +43,7 @@ def _resolve(name):
     """Resolve *name* to a real log file inside the log directory, else None."""
     # DISPATCHARR_LOG_DIR may hold more than logs.
     log_dir = settings.LOG_FILE_DIR
-    if not log_dir or not _NAME_RE.match(name) or not name.startswith(BASE_NAME):
+    if not log_dir or not _NAME_RE.fullmatch(name) or not name.startswith(BASE_NAME):
         return None
     base = os.path.realpath(log_dir)
     path = os.path.realpath(os.path.join(base, name))
--- a/core/log_files.py
+++ b/core/log_files.py
@@ -32,6 +32,14 @@ def _log_dir():
     return getattr(settings, "LOG_FILE_DIR", None) or "/data/logs"
 
 
+def _open_log(path):
+    """Open *path*, mapping a file pruned since resolution to a 404."""
+    try:
+        return open(path, "rb")
+    except FileNotFoundError:
+        raise NotFound("Log file not found")
+
+
 def _at_line_start(f, offset):
     """Whether *offset* sits just past a newline, i.e. begins a record."""
     if offset == 0:
@@ -124,7 +132,7 @@ def get_log_file(request, name):
         raise NotFound("Log file not found")
 
     cursor = request.GET.get("cursor", "")
-    with open(path, "rb") as f:
+    with _open_log(path) as f:
         # Identity and size come from the open handle, so a rotation cannot land
         # between them and leave us seeking past the end of a fresh, empty file.
         stat = os.fstat(f.fileno())
@@ -177,7 +185,7 @@ def download_log_file(request, name):
         raise NotFound("Log file not found")
 
     response = FileResponse(
-        open(path, "rb"), content_type="text/plain; charset=utf-8"
+        _open_log(path), content_type="text/plain; charset=utf-8"
     )
     response["Content-Disposition"] = f'attachment; filename="{name}"'
     return response
--- a/core/log_files.py
+++ b/core/log_files.py
@@ -36,7 +36,7 @@ def _log_dir():
 
 def _resolve(name):
     """Resolve *name* to a real log file inside the log directory, else None."""
-    # The family check lives here, not only in the listing: DISPATCHARR_LOG_DIR may point at a directory holding more than logs.
+    # Checked here, not only in the listing: DISPATCHARR_LOG_DIR may hold more than logs.
     if not _NAME_RE.match(name) or not name.startswith(LOG_NAME_PREFIX):
         return None
     base = os.path.realpath(_log_dir())
--- a/core/log_files.py
+++ b/core/log_files.py
@@ -154,10 +154,9 @@ def get_log_file(request, name):
         if newline >= 0:
             start += newline + 1
             data = data[newline + 1 :]
-    elif not reset:
-        # Mid-write, the tail is a fragment; it arrives whole on the next poll.
-        end = data.rfind(b"\n")
-        data = data[: end + 1] if end >= 0 else b""
+    # Mid-write, the tail is a fragment; it arrives whole on the next poll.
+    end = data.rfind(b"\n")
+    data = data[: end + 1] if end >= 0 else b""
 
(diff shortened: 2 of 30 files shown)


SergeantPanda:
Hi @nagelm, is this one ready to review still after the main logging was merged?

nagelm:
@SergeantPanda I did some more hardening/fixing/performance improvements today and it's ready for your review

SergeantPanda (changes requested):
Started playing with this, looks pretty good overall. I haven't got a chance to run a FULL test of it yet but one thing that caught my eye right away was we have the default log file size set to 10MB but the viewer will only display a max of 5MB with no option to pull more.

I think from a UX standpoint, we should either have the default be 5MB so ALL logs can be shown at once, or, we should allow the user to pull in the next 5MB worth of logs (this is probably the best path but does get complicated when dealing with rotating logs.)

One other minor detail, the table that shows the log files, can we switch that to use our custom table so it matches the rest of the tables in the project?

Like I said though, this does work quite well and am pretty impressed with how far you got it already, thank you!

nagelm:
Useful feedback thanks. The 5MB is half overflow guard / half vestigial update machinery. Previously I was tailing 5MB on every viewer update but that proved pretty wasteful so I moved on towards the cursor based approach. I think in the order of 5-10MB the overflow guard is not doing anything (the only time it is really doing anything useful is if someone manually sets a really large log size e.g. 100MB).

Noting 5MB is about ~30k log lines so probably plenty for the average user but my setup is small so it is difficult for me to understand how much large setups would churn on these settings.

My recommendation would be more or less what you said: 
1. Get rid of the hidden 5MB display layer; 
2. Set default log size to 5MB; 
3. display the full log in the viewer (5MB by default but larger if people manually change their settings). 
4. Keep some hardcoded safety limit in the viewer which is uncorrelated with the log size (AND/OR put an upper bound on the log size in settings).
5. In a separate follow-up PR implement viewer caching / lazy load / 'see more'

I will switch to the custom table and do points 1-3 while I wait for your feedback.

SergeantPanda:
That all sounds perfect to me. Thank you!

And yeah, completely understood on have a small setup, I have the same issue. I do leave my log level set at DEBUG which does generate a lot more logs than most people would have though so that helps.

nagelm:
A few changes from what I said:

**1. Hidden display layer limiting is gone** - there was a window mechanism limiting at 5,000-lines or 5MB and that is completely gone and now every line is reachable in the viewer.  

**2. Default log size → 5 MB** (was 10). 

| Constant | Value Now | (Was) |
|---|---|---|
| `DEFAULT_LOG_MB` | 5 | 10 |
| `MAX_LOG_MB` | 20 | 1000 |
| `DEFAULT_LOG_KEEP` | 5 | 5 |
| `MAX_LOG_KEEP` | 50 | 50 |

**3. The viewer shows the whole log, at whatever size you set.** 

**4. Safety constants**

| Constant | Value | Safety |
|---|---|---|
| MAX_VIEW_BYTES (core/log_files.py:23) | 24 MB | Server side memory guard on files larger than the log limit (i.e. files the collector didn't write) |
| MAX_BUFFER_BYTES (LogFileView.jsx:142) | 24 MB | Client side accumulation backstop |
| MAX_BUFFER_LINES (LogFileView.jsx:143) | 300000 | Client side object count backstop |

**5. The lazy-render is done** I went ahead and did it in the same PR so the log body is virtualized with `react-virtualized` `List` + `CellMeasurer` — only the visible window is in the DOM (~12–23 rows in a browser, whatever the file size). I'm not super familiar with this sort of thing and tThere were a few bugs I had to fix so please play around with this when you take another look.

Also, the standardized custom table primitive was applied.

nagelm:
A few more fixes most notably the virtualization was tripping on the variable line heights caused by wrapping so I had to add an expand/contract mechanic so that things were not fragile.

Everything ready for rereview now I think.

SergeantPanda (approved):
This looks really good. I have a couple of minor tweaks to make after this lands (more strict regex for matching log file names and a missed default from 10 to 5 in settings utils)

Thank you for the work on this. I really appreciate you taking the time to implement this in a way that matches the app and prioritizes performance.

PR #1602, opened 2026-08-24
fix(logs): quieten the container's boot and runtime output [logging 3]

nagelm:
## Description

Quietens the container's output now that [logging 1] makes every line visible in one place. Six changes, each removing noise rather than hiding information:

- **The celery banner stops reprinting in every process.** It is kept for Debug and Trace, where someone asked for detail.
- **The resolved log level stops being echoed from every process.** Every record already carries its level; the detection trace is kept for Debug and Trace only.
- **Readiness probes stop provoking postgres `FATAL` lines.** The probes connected before the database and role existed, so a normal boot reported two fatal errors that were not errors.
- **uwsgi request records carry their own level** rather than the container's configured verbosity, so a request log line is not mislabelled as whatever `DISPATCHARR_LOG_LEVEL` happens to be.
- **Shell events and swallowed failures get a real severity.** The entrypoint gains a `log_line` helper so a warning from the boot script reads as a warning instead of undifferentiated stdout.
- **The access token stops being written to the browser console.**

### Merge order

**This PR requires [logging 1].** `dispatcharr/settings.py` here calls `startup_log()` eight times, and `dispatcharr/startup_log.py` does not exist on `dev` — it is introduced by [logging 1]. Rebased onto `dev` alone, Django raises at settings import in every process. The entrypoint's new `log_line` helper also emits the collector's canonical grammar, which nothing on `dev` parses.

It does **not** depend on [logging 2]; it touches no file that PR owns. It can land before or after [logging 2], in either order.

## Related Issue

Closes #

## How was it tested?

Built the image from this branch and booted it against a fresh volume (`TZ=Pacific/Auckland`), then compared its first-boot log against a container built from [logging 1] alone, booted identically:

| | [logging 1] alone | with this PR |
|---|---|---|
| boot lines | 533 | **457** |
| postgres `FATAL` | 2 | **0** |
| log-level echoes | 40 | **0** |
| celery banner lines | 4 | **0** |

The app serves (`/login` → 200) and the collector runs in both.

Every changed shell script passes `bash -n`; every changed Python file parses. The collector's own suite is unaffected by this branch and still passes.

## Checklist


SergeantPanda:
Hey @nagelm I see you have been keeping this branch up to date, curious, are you ready for this to be reviewed?

Issue #1361, opened 2026-06-17
[Feature]: VOD for standard web users


TubbyMuncher:
### Is your feature request related to a problem?

I didnt' see an open issue that specifies the allowance for web 'standard' users to view VOD. Apparently this is only available through SC.

### Describe the solution you'd like

Please provide a user access level that allows users to view both TV and VOD through the web interface.

### Additional context

Nothing Extra

PR #1607, opened 2026-08-25
feat(vod): add per-user movie and series access flags

### LINKED ISSUE #1188: [Feature]: Per-user VOD access control

**Feature Request: Per-user VOD access control**

**Summary**
Add the ability to enable or disable VOD access (movies and/or series) on a per-user basis, independently of live channel access.

**Motivation**
Currently, any user with XC credentials gets full VOD access alongside live channels. There's no way to grant a user live TV access only without also exposing the entire VOD catalog. This limits how safely and flexibly Dispatcharr can be used in shared/multi-user setups.

Note: user-level M3U account restrictions for VOD were removed in v0.10.0 (#402), so there's currently no native path to achieve this.

**Proposed solution**
Add a toggle (or permission flag) in the user edit form to allow/deny VOD access — ideally split between movies and series. When VOD is disabled for a user, the XC API endpoints `/get_vod_streams`, `/get_series`, `/movie/`, and `/series/` should return empty or 401 for that user.

**Use case**
Sharing live IPTV access with family/friends without exposing the full VOD library

Describe here.

### Additional context

Nothing Extra

brendongl:
## Description

Adds two per-user flags, `vod_movies_enabled` and `vod_series_enabled`, that turn movies and series on or off independently of each other and of live channel access. Both default to on.

The issue asks for a toggle "ideally split between movies and series", so these are two separate switches rather than one VOD flag — turning both off is the "no VOD at all" case.

**Why:** there is currently no way to give a user live TV without also handing them the entire VOD catalogue. The user-level M3U restrictions that used to do this were removed in `48a2f2da`, and nothing replaced them, so every authenticated XC user gets VOD from every active M3U account.

**How:** the flags follow the existing `catchup_enabled` pattern exactly — JSON booleans on `User.custom_properties` (so no migration), admin-managed only, read through a single helper.

`apps/vod/utils.py` (new):

- `is_vod_enabled(kind=..., user=...)` — one place decides what an absent flag means. Absent reads as enabled, so existing users are unaffected by the upgrade. An unknown `kind` raises `ValueError` rather than failing open.
- `vod_kind_for_content_type()` — maps the VOD proxy's `content_type` to a flag kind. Episodes are gated by the series flag; anything the flags don't cover returns `None` and is left alone.

For a disabled kind, that user gets:

| Surface | Behaviour |
| --- | --- |
| `get_vod_categories`, `get_vod_streams`, `get_series_categories`, `get_series` | `[]` |
| `get_vod_info`, `get_series_info` | 404 |
| `/movie/{u}/{p}/{id}.{ext}`, `/series/{u}/{p}/{id}.{ext}` | 403 |
| `/proxy/vod/{type}/{uuid}` | 403 |

Two decisions worth a maintainer's call:

**Empty lists rather than 401 on the catalog actions.** The issue suggested "empty or 401". Several XC clients treat a 401 on `get_vod_streams` as the whole account being dead rather than just VOD-less, which would break live TV for a user who only lost VOD. Happy to switch to 401 if you'd rather.

**`stream_vod()` gates too, not just the XC handlers.** Without that, a user with the flag off could still play by hitting the native `/proxy/vod/{type}/{uuid}` route with a JWT, so the XC gate alone would be decorative.

Frontend is two switches in the user Permissions tab, directly below Enable Catchup, plus the matching entries in `UserUtils.js`. Both flags are in `ADMIN_ONLY_PROPS` so a user cannot re-enable themselves through `PATCH /api/accounts/users/me/`.

**Scope deliberately left out:** `/api/vod/movies|series|episodes/` are not gated. A disabled user can still list VOD metadata over REST with a JWT — they just cannot play any of it. Standard-user web VOD is tracked separately in #1361, and gating those viewsets means touching `UnifiedContentViewSet`'s hand-rolled `list()`, so it felt like it belonged there rather than here. Say the word and I'll add it.

Also unchanged and pre-existing: `stream_vod` is `AllowAny` with no authentication requirement, so an anonymous caller who already knows a content UUID can stream. That is not introduced or made worse by this PR, but you may want an issue for it.

**The second commit** (`fix(vod): return 401 instead of 500 on bad XC VOD credentials`) is a one-line drive-by fix. `Response` is never imported in `apps/proxy/vod_proxy/views.py`, so the invalid-credential and blocked-network branches of `stream_xc_movie` / `stream_xc_episode` raise `NameError` and return HTTP 500 instead of 401/403. It is in the same two functions this PR modifies, and one of the new tests asserts the 401, which is why it is bundled. Glad to pull it into its own PR if you'd prefer.

## Related Issue

Closes #1188

## How was it tested?

Built and ran a real instance from this branch — `docker/docker-compose.dev.yml` against `dev` @ `f234f35` (v0.29.0) — seeded a movie, a series and an episode with `M3UMovieRelation` / `M3USeriesRelation` / `M3UEpisodeRelation` rows against an active M3U account, and exercised every flag state through the actual XC endpoints with `curl`, reloading uWSGI between changes:

| Flag state | movies catalog | `get_vod_info` | `/movie/` | series catalog | `get_series_info` | `/series/` | `get_live_streams` |
| --- | --- | --- | --- | --- | --- | --- | --- |
| flags absent (upgrade case) | 1 movie | 200 | 301 → stream | 1 series | 200 | 301 → stream | 200 |
| both `true` | 1 movie | 200 | 301 → stream | 1 series | 200 | 301 → stream | 200 |
| movies `false`, series `true` | `[]` | 404 | **403** | 1 series | 200 | 301 → stream | 200 |
| movies `true`, series `false` | 1 movie | 200 | 301 → stream | `[]` | 404 | **403** | 200 |
| both `false` | `[]` | 404 | **403** | `[]` | 404 | **403** | 200 |

Also verified by hand:

- `GET /proxy/vod/movie/{uuid}` with a JWT for a user whose movie flag is off → 403, confirming the XC gate cannot be walked around via the native route.
- The full round trip through the built frontend as an admin: Users → Permissions → toggled Enable Series off → Save → confirmed `vod_series_enabled: False` landed in `custom_properties`, that the XC surface changed accordingly, and that reopening the form shows the saved state (Movies on, Series off).
- `GET /movie/<user>/<wrong-password>/<id>.mp4` → 500 before the second commit, 401 after.

Automated:

- `python manage.py test` → **1903 tests, 0 failures.** 21 are new: 8 in `apps/output/tests/test_views.py` and 13 in a new `apps/proxy/vod_proxy/tests/test_vod_access_flag.py`, covering absent/true/false per flag, flag independence in both directions, detail-lookup 404s, playback 403s, that only a JSON `false` disables (so a truthy or missing value cannot lock a user out), that an unknown `kind` raises, and that live channels are unaffected.
- `npm run test` → **6227 tests, 0 failures**, 2 new in `UserUtils.test.js` for the payload mapping and the default-to-true behaviour.
- `npx eslint` on the three changed frontend files → no errors (one pre-existing `react-hooks/exhaustive-deps` warning at `User.jsx:105`, untouched by this PR).
- `npx prettier --check` → clean on every line this PR adds. Both touched JS files already had pre-existing Prettier diffs on unrelated lines; I left those alone rather than reformatting files outside the scope of this change.

## Checklist

Two checklist notes: no models were changed — the flags live in the existing `custom_properties` JSONField — so there is no migration to include; and no new API endpoints were added, so the OpenAPI schema is unchanged.

Disclosure per CONTRIBUTING's AI section: I used an AI assistant while writing this. It is not unreviewed output — I have read every line, the design deliberately mirrors the existing `catchup_enabled` implementation rather than inventing a new pattern, and I can explain any line and its tradeoffs in review.


SergeantPanda (approved):
Thank you for this, overall a solid implementation that closely mirrors how we handled catchup permissions.

I did change several of the function and variable names for consistency with the application.

I also added the same gate for head that we have for get.

Thanks again!

PR #1609, opened 2026-08-26
feat(dvr): opt-in "treat untagged episodes as new" toggle for series rule mode "new" (#1573 rework)

v8eta:
## Description

Rework of #1573 along the lines suggested there: the behaviour change is now a **per-rule opt-in**, not a new default.

A `mode="new"` series rule accepts a programme only when the EPG tagged it `<new/>`. XMLTV also lets a feed mark **repeats** with `<previously-shown/>` instead, and a feed using that convention exclusively never sets `<new/>` on anything — so on such a feed a `mode="new"` rule silently matches nothing at all (on the instance where this was found, a rule sat idle for months with zero recordings; measured there, all four feeds tag only repeats: 244/244 upcoming programmes carry neither tag).

The first version made "not previously-shown" count as new unconditionally, which — as the review pointed out — breaks the opposite kind of feed, one that tags only first runs with `<new/>` and leaves repeats untagged: that feed would start recording everything.

### The change

A new optional boolean on the rule, `untagged_is_new` (default `false`):

- **Off (default):** exactly the current behaviour — only `<new/>` counts. Feeds that tag first runs are unaffected.
- **On:** a programme counts as new when tagged `<new/>` **or** when carrying neither tag. A programme tagged `<previously-shown/>` is always excluded.

Applied identically in both places that hold a copy of the freshness filter — the scheduler (`tasks.py`) and the Edit Series Rule preview (`api_views.py`), including the per-row `is_new` flag, so the dialog always agrees with what would record.

The rule is a JSON dict in `CoreSettings`, so there is no model change and no migration. The flag is only persisted when `mode == "new"` and it is set, so existing rules and rules on well-tagged feeds are byte-identical to before.

### Frontend

The Edit Series Rule dialog shows a checkbox — *"Treat untagged episodes as new"* — only while **New only** is selected, hydrated from the rule and included in both the preview payload and the save payload, so the preview count reflects the toggle live.

## Related Issue

Follow-up to #1573 (closed); reworked per the maintainer feedback there.

## How was it tested?

Backend — 6 tests in `apps/channels/tests/test_series_rule_new_mode.py` on the existing `SeriesRuleDedupBaseTestCase` fixture, covering both sides of the toggle:

- default: `<new/>` records; a programme with neither tag is **not** assumed new (stock behaviour preserved)
- flag on: an untagged first run records; a marked repeat is still excluded; a mixed schedule (evening first runs + daytime repeat) records exactly the first runs; an explicit `<new/>` stays authoritative alongside a previously-shown date

Full run with both existing series-rule suites: **57 tests, OK** (6 new + `test_series_rule_dedup` + `test_series_rule_dedup_time_drift`, zero regressions).

Frontend — 5 new cases in `SeriesRuleEditorModal.test.jsx`: checkbox visible only for mode `new`, pre-checked from the rule, `untagged_is_new` present in the save payload when checked and absent when not. ESLint and Prettier pass on both touched files.

Also verified against the live EPG on the affected instance: the flag-on predicate is the same one that has been running there as a local patch since the first version of this change, where it selects exactly the airings not tagged `previously_shown` (checked again today: a week of repeats — 4 airings all tagged `previously_shown` — correctly yields 0 matches with the flag on). With the flag off, behaviour is byte-for-byte stock.

## Checklist

- [x] Backend: migrations are included if any models were changed (N/A — no model changes; series rules are JSON in CoreSettings)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (no new endpoints; the new optional field is declared on the existing inline serializer)


SergeantPanda (approved):
Thank you for the PR. I like this method much better.

I created a shared helper to avoid future issues if the code changes slightly. I also changed the checkbox to a switch and increased the text size of the description.

Otherwise, everything else looks good!

Issue #1391, opened 2026-06-26
[Feature]: Plex DVR parser bug work around


toomanydevils:
### Is your feature request related to a problem?

Plex's DVR epg data parsing is broken, and fails to correctly parse guide data from the xml file for airings that have identical titles, and copies the description to all airings. The best example is with multiple airings of sports with the same title, NFL Football, MLB Baseball, FIFA World Cup etc etc.

I came across a user mcarpenter (@MCHammer65 on git) on the plex forums asking for feedback on their m3u proxy app "tunaar", and I had made a comment about the epg issue, and they replied right away essentially and found a work around for their app by adding a toggle that "folds each programme’s sub-title into its title in the guide it hands Plex". [Original Plex Forum Reply](https://forums.plex.tv/t/tunaar-iptv/939933/5) 

I had issues with trying to spin it up on unraid, I'm not that savy with docker and linux all around lol. I usually use the templates. I was considering running in tandem to solve the guide data for plex, and just wanted to see if his workaround worked for me. 

Would it be possible to get a feature like this implemented? 

### Describe the solution you'd like

opt-in toggle enabling/disabling DVR title-subtitle merging for epg output from dispatcharr

### Additional context

Nothing Extra

MCHammer65:
Hi - I am in process of a) making the issue you mention less of an issue and b) building on Disatcharr's features - MediaHelm that has an array of other features

✅ See who's watching (direct play vs. transcode) ✅ "Why did this fail?" playback diagnostics ✅ One-click metadata & poster rollback ✅ Turn IPTV into Live TV — auto-discovered HDHomeRun tuner, one upstream shared across viewers, provider failover, FFmpeg profiles ✅ Secure by default · installable PWA · runs on any NAS (arm64 too)
One container. Open the box, follow setup, you're live 👇 docker run -d -p 5005:5005 -v ./data:/data ghcr.io/mchammer65/mediahelm:latest
#selfhosted #homelab #Plex #Jellyfin #IPTV

________________________________
A couple of things I couldn't do for you here: I can't set the GitHub About description or topics from this session (no repo-settings access) — paste the tagline/topics above into the repo's About panel. Also note the landing copy references the ghcr.io/mchammer65/mediahelm:latest image, which only exists once the publish workflow runs on main (the CI is on the branch but hasn't built an image yet) — until then, the "build from source" path is the working install.
[...]

Issue #1197, opened 2026-04-16
[Feature]: Save DVR Recording to a custom folder


guimello87:
### Is your feature request related to a problem?

It would be great to have the option to save the DVR recording to a specified folder. 

I use Dispatcharr with Unraid and it gets saved to the appdata folder, which is not ideal.

### Describe the solution you'd like

A simple option under Setting->DVR where we specify the folder DVR recording will be save to.

### Additional context

Nothing Extra

fastgriz:
You would have to set up a "Path" in unraid's doker GUI, which you can already do.    I have /data/recordings mapped to /mnt/user/PVR/Dispatcharr/

sethwv:
Per @fastgriz's note, customizable save paths are already a feature - you will need to add a new Path mapping in Unraid (or a volume mount in docker run / docker compose)

In my unraid installation, I map the `/data/recordings` container path to `/mnt/user/media/uncategorized/_dispatcharr/` on the host.
<img width="537" height="92" alt="Image" src="https://github.com/user-attachments/assets/7a7a0c1d-a302-4303-9934-da04424c3277" />

In Dispatcharr, I match the following paths to the container path that was set above.
<img width="831" height="773" alt="Image" src="https://github.com/user-attachments/assets/6e583387-cd7b-4ccc-9e86-594177257e5b" />

PR #1615, opened 2026-08-28
feat(dvr): broadcast-date placeholders for DVR path templates

### LINKED ISSUE #682: [Feature]: More options for naming of output files in DVR recordings

I am wanting to use the DVR function in dispatcharr and then view those recordings in plex.  Plex will only pickup the recordings however if they are named in a specific way.  I mostly record sports so things such as motogp and f1 do not have episode / series etc data.  Template used when an episode has no season/episode. Supports {show}, {start}, {end}, {channel}, {year} but i also found {sub_title} works.

It would be great if there were some other options, for my needs specifically it would be great if you could do a date say in the form of {yyyy-mm-dd} this would allow plex to pick it up as i would be able to use a string such as 

/recordings/{show}/Season {year}/{show} - {yyyy-mm-dd} - {sub_title}.mkv

### Additional context

Nothing Extra

v8eta:
## Description

A DVR path template can name a recording by season and episode, or fall back to the `{start}` timestamp (`20260115_203000`), but there is no way to write a plain calendar date. Content that carries no season/episode numbers — sports, news, one-off events — therefore cannot be named in the `Show - YYYY-MM-DD - Subtitle` form Plex and Jellyfin parse. That is what #682 asks for.

This adds four placeholders to the values every DVR template is formatted with, so they work in the **TV**, **TV fallback**, **movie**, and **movie fallback** templates alike:

| Placeholder | Type | Example |
| --- | --- | --- |
| `{start_date}` | string | `2026-01-16` |
| `{start_year}` | int | `2026` |
| `{start_month}` | int | `1` — `{start_month:02d}` gives `01` |
| `{start_day}` | int | `16` — `{start_day:02d}` gives `16` |

The example from #682 now works as written:

```
TV_Shows/{show}/Season {start_year}/{show} - {start_date} - {sub_title}.mkv
```

### Two decisions worth calling out

**1. The values are the broadcast date in the configured system time zone, not UTC.**

`start_time` reaches `_build_output_paths` as an aware UTC datetime, so `strftime` on it directly yields the *UTC* calendar day. For anyone not on UTC that is the wrong day for a good part of the schedule: a 20:30 UTC airing is already 16 January in `Asia/Singapore`, and a 02:30 UTC airing is still 14 January in `America/Chicago`. A date placeholder that is off by a day for evening recordings would be worse than none.

The zone comes from `CoreSettings.get_system_time_zone()` — the **Time zone** control under Settings → UI, and the same value `sync_recurring_rule_impl` already uses to compute `local_today`. So the day in the filename cannot disagree with the day the series rule matched. That resolution (with its existing invalid-zone warning and fallback to the server zone) is extracted into `_system_timezone()` so both call sites share one implementation instead of two copies; behaviour at the existing call site is unchanged.

`get_system_settings()` is Redis-cached and returns defaults when the group is absent, so this adds no Postgres round trip per recording and is safe before any settings have been written.

**2. `year` / `month` / `day` are integers, not preformatted strings.**

The TV template help already tells users to *"use format specifiers like `{season:02d}`"*, and real templates do exactly that with numeric parts. Were these strings, `{start_month:02d}` would raise `Unknown format code 'd' for object of type 'str'` — and since a template that fails to format is silently replaced by a fallback rather than reported, the user would see an inexplicably wrong filename instead of an error. Integers make both `{start_month}` and `{start_month:02d}` work, and `{start_date}` covers the zero-padded ISO case in a single token.

### Deliberately not changed

- **`{original_air_date}` keeps its current scope** — TV fallback template only, per #1518. These placeholders are the programme's *broadcast* date on this channel, which is a different value; the existing tests asserting `{original_air_date}` is unavailable to the normal TV and movie templates pass untouched.
- **`{start}` and `{end}` keep their UTC form**, so no existing template changes behaviour. There is a test pinning that.
- **`{year}` is untouched** — it stays EPG-derived (production year, falling back to the start year). `{start_year}` is the new, unambiguous broadcast year.
- The **Date format** preference (`DD/MM/YYYY` vs `MM/DD/YYYY`) under Settings → UI is a display setting and is not applied to filenames. `{start_date}` is always ISO `YYYY-MM-DD`, which is what media servers parse and what #682 asked for.

Adding keys to the format dict is backward compatible: no existing template can start failing, and an unknown placeholder still falls through to the same fallback it does today.

## Related Issue

Closes #682

Related, and intentionally not addressed here:

- #1518 — `{original_air_date}` for TV fallback templates. Its stated scope boundary is preserved rather than widened.
- #1307 — season/episode not picked up from the EPG. That bug pushes recordings onto the fallback template, which is part of why the fallback template needs these placeholders too, but the parsing bug itself is untouched.
- #1391 — folding sub-title into title for Plex's guide parser. Different mechanism (EPG output), same underlying frustration with how Plex names things.
- #1197 — custom DVR output folder (closed).

## How was it tested?

**Backend** — 10 new tests in `apps/channels/tests/test_recording_template_dates.py`, driving `_build_output_paths` with the four template getters and `get_system_time_zone` patched:

- `{start_date}` in `UTC`, `Asia/Singapore` and `America/Chicago` from one instant — the date rolls forward for `+08:00` and does not for `-06:00`
- a second instant where the date rolls *backward* for `-06:00`, so a pass cannot be explained by the local date coincidentally equalling the UTC one
- `{start_month:02d}` / `{start_day:02d}` pad; bare `{start_month}` / `{start_day}` render as unpadded integers
- the placeholders reach the TV fallback template (no season/episode) and the movie template
- `{start}` / `{end}` still render as UTC `%Y%m%d_%H%M%S` while the system zone is `Asia/Singapore`
- an unresolvable stored zone (`Not/AZone`) still produces a path, and `_system_timezone()` returns the server zone
- `_system_timezone()` returns the configured zone

**Mutation-checked**, so the tests are not passing vacuously:

- reverting the conversion to `local_start = start_time` → 5 failures, exactly the zone-dependent ones; the `UTC` and `America/Chicago` same-day cases correctly still pass
- removing the four keys entirely → 8 failures, each landing on the hardcoded last-resort path, which also confirms the graceful-fallback claim above
- running the new frontend test against the unmodified descriptions → fails; against the change → passes

**Full run:** `manage.py test apps.channels core` → **Ran 596 tests, OK** (10 new, zero regressions), on PostgreSQL 17 with `dispatcharr.settings_test`. `scripts/ci_backend_test_labels.py` resolves this diff to `["apps.channels.tests"]`, which is inside that run.

**Frontend** — 1 new test in `DvrSettingsForm.test.jsx` asserting all four placeholders are advertised on all four template descriptions; that file is **32 passed**. `npx prettier --check` is clean on both touched frontend files, and `npx eslint` reports **0 errors** (the 2 `react-hooks/exhaustive-deps` warnings are pre-existing — verified by linting the unmodified file from `HEAD`, which reports the same 2).

Note for anyone re-running the whole frontend suite: it is currently nondeterministic on a loaded machine (an unmodified checkout produced 22 failures in one run and the changed tree 3 in another), which is why the frontend evidence above is per-file rather than a suite-wide count.

## Checklist

- [x] Backend: migrations are included if any models were changed (N/A — no model changes; the placeholders are computed in the task)


SergeantPanda (approved):
This looks really good overall. 

I made one small change to use the actual program start time instead of the recording window start time (which uses offsets)

Thank you!

PR #1619, opened 2026-08-28
feat(dvr): capture recordings through a configurable output profile

### LINKED ISSUE #1618: [Feature]: DVR recordings ignore Output Profiles — a recording is always a raw copy of the source

A DVR recording is always a raw copy of whatever the source sends. There is no way to make recordings use an Output Profile, so the transcode settings that apply to live playback cannot be applied to the thing you keep.

Tracing it through `dev`:

1. The DVR builds its capture URL in `apps/channels/tasks.py` with no query string:

   ```python
   stream_url = f"{base}/proxy/ts/stream/{channel.uuid}"
   ```

2. `stream_ts` resolves a profile via `_resolve_output_profile(request, user)` in `apps/proxy/live_proxy/views.py`, which tries exactly three things:

   ```python
   param = request.GET.get('output_profile')      # 1. query parameter
   ...
   profile_id = custom.get('output_profile')      # 2. authenticated user's custom_properties
   ...
   return None                                    # 3. nothing
   ```

3. The DVR's ffmpeg fetches that URL with only a User-Agent — no token, no cookie, no auth header:

   ```python
   "-user_agent", dispatcharr_dvr_user_agent(recording_id),
   ...
   "-i", stream_url,
   "-c", "copy",
   ```

   so the request is anonymous, `user` is `None`, and branch 2 cannot apply to a recording either.

4. With `None`, no profile command is built at all — `views.py` only does this `if resolved_output_profile`:

   ```python
   if resolved_output_profile:
       cmd = resolved_output_profile.build_command()
       if not proxy_server.ensure_output_profile(channel_id, resolved_output_profile.id, cmd):
   ```

5. And the DVR's own ffmpeg is `-c copy`, so it cannot encode anything itself. The proxy's Output Profile is the only place encoding can happen for a recording.

There is also no system-wide fallback to close the gap: `CoreSettings` has `get_default_output_format()`, but that is a container format (mpegts/fmp4), not a profile. `hdhr_output_profile_id` exists but is Stream-Settings state consumed by `apps/hdhr/api_views.py` for the HDHR path, and the web-player profile is a UI setting for the browser player. Neither reaches the DVR.

The net effect is that someone who has configured a working Output Profile — say to normalise audio, deinterlace, or store recordings in HEVC instead of a large MPEG-2 original — cannot apply any of it to recordings. Related recording-quality reports such as #1422 and #1293 sit inside this same raw-copy path.

A DVR-specific output profile setting, defaulting to unset so current behaviour is unchanged:

- `dvr_output_profile_id` in DVR settings, with a `CoreSettings.get_dvr_output_profile_id()` accessor mirroring the existing `get_hdhr_output_profile_id()`.
- A **DVR Output Profile** select in the DVR settings form, alongside the comskip and path-template fields, listing active Output Profiles and clearable back to "None".
- When it is set, the DVR appends `?output_profile=<id>` to the capture URL. When it is unset, the URL is byte-identical to today's and recordings stay a raw copy.

No new plumbing is required in the proxy: `/proxy/ts/stream/<uuid>` already honours `?output_profile=`, and `_resolve_output_profile` already validates that the profile exists and is active, falling back to `None` if not. The only missing pieces are somewhere to store the choice and the line that passes it.

I have been running exactly that query-parameter approach locally against a HEVC VAAPI profile and recordings encode correctly through it, so the mechanism works as-is — what is missing is a supported way to configure it.

If a per-DVR setting is felt to be too narrow, an alternative would be a general "default output profile" in Stream Settings that `_resolve_output_profile` falls back to before returning `None`. That is a smaller change but a wider blast radius, since it would also alter every anonymous live request, so the DVR-scoped setting seems the safer shape. Happy to build whichever is preferred.

### Additional context

I searched for prior art before opening this and did not find it — #1304/#1305 covered the web player and TV Guide, #1230 introduced Output Profiles, and #1610 uses "output profiles" for VOD source selection, which is a different concept. If I missed an existing issue, point me at it and I will close this.

Happy to submit the PR if the approach is agreed.

(Replaces #1616 and #1617, both closed by the template bot because they were filed from the CLI and so carried no issue type. Same content, nothing else changed.)

v8eta:
## Description

Adds a **DVR Output Profile** setting so recordings can be captured through an Output Profile. Today a recording is always a raw copy of the source, and nothing in the current path can change that:

- The DVR builds its capture URL with no query string, so `_resolve_output_profile`'s first branch (`?output_profile=`) never fires.
- Its FFmpeg fetches that URL with only a User-Agent — no token, no cookie, no auth header — so the request is anonymous, `user` is `None`, and the second branch (the user's `custom_properties`) cannot fire either.
- `_resolve_output_profile` therefore returns `None`, and `views.py` only builds a profile command `if resolved_output_profile`.
- The DVR's own FFmpeg is `-c copy`, so it cannot encode anything itself.

The proxy's Output Profile is the only place a recording can be encoded, and there was no supported way to select one. There is no system-wide fallback either: `get_default_output_format()` is a container (mpegts/fmp4), not a profile.

### The change

| Piece | What it does |
| --- | --- |
| `dvr_output_profile_id` in DVR settings | Stores the choice; defaults to `None` |
| `CoreSettings.get_dvr_output_profile_id()` | Mirrors the existing `get_hdhr_output_profile_id()`; coerces whatever is stored and yields `None` for anything unusable |
| `_dvr_capture_url(base, uuid, profile_id)` | Builds the capture URL, appending `?output_profile=<id>` only when an id is set |
| **DVR Output Profile** select | Lists active profiles, clearable back to none, in the DVR settings form |

No proxy changes: `/proxy/ts/stream/<uuid>` already honours `?output_profile=`, and `_resolve_output_profile` already validates that the profile exists and is active.

### Behaviour when unset or misconfigured

- **Unset** — the capture URL is byte-identical to before and recordings stay a raw copy. This is the default, so existing installs are unaffected. There is a test asserting exactly this.
- **Missing or deactivated profile** — `_resolve_output_profile` returns `None` and the capture degrades to the same raw copy rather than failing the recording.
- **`0`** — treated as unset; it is not a valid pk and would resolve to no profile anyway.

### Two notes on the shape

**Why a DVR-scoped setting rather than a global default.** The alternative discussed in #1618 was a general default in Stream Settings that `_resolve_output_profile` falls back to before returning `None`. That is a smaller diff but a much wider blast radius — it would change every anonymous live request, not just recordings. Happy to switch if you'd prefer that shape.

**Why the key is `dvr_output_profile_id` and not `output_profile_id`.** It already lives in the DVR group, so the prefix looks redundant — but the settings form flattens every group into a single form-key namespace (`streamFields` / `dvrFields` / …), where a bare `output_profile_id` would sit ambiguously next to the existing `hdhr_output_profile_id`.

**Why `_dvr_capture_url` was extracted.** The URL was built inline inside `run_recording`, which is far too large to invoke from a test — the existing tests for that task resort to `inspect.getsource` string matching. Pulling out four lines makes the actual behaviour directly testable; the call site keeps one guard test for the wiring.

## Related Issue

Closes #1618

## How was it tested?

**Backend** — 9 new tests in `apps/channels/tests/test_dvr_output_profile.py`:

- unset and `None` leave the URL byte-identical; a set id appends the parameter; `0` is treated as unset
- `get_dvr_output_profile_id()` round-trips an int, coerces the numeric string the frontend Select stores, and yields `None` for `""`, `"abc"`, `[]` and `{}` — a bad value must disable the feature, never break scheduling
- the configured setting reaches the URL builder, and the call site inside `run_recording` passes it through

**Frontend** — 3 new tests:

- `DvrSettingsForm.test.jsx`: the select offers only **active** profiles, and a selection is stored as an integer
- `SettingsUtils.test.js`: `dvr_output_profile_id` routes to the `dvr_settings` group coerced to an integer, and a cleared value is stored as `null`

**Mutation-checked**, so none of it passes vacuously:

- dropping the `if output_profile_id:` guard → exactly the two "unchanged by default" tests fail
- removing the key from the id-coercion list → exactly the integer-routing test fails, while the null test still passes

**Full runs:** `manage.py test apps.channels core` → **595 passed, OK** on PostgreSQL 17 with `dispatcharr.settings_test`. Frontend: `DvrSettingsForm.test.jsx` 33, `SettingsUtils.test.js` 35, plus `StreamSettingsForm.test.jsx` 44 as a regression check on the shared settings plumbing — **112 passed**. `prettier --check` clean on all four touched frontend files; `eslint` reports 0 errors (the 2 `react-hooks/exhaustive-deps` warnings in `DvrSettingsForm.jsx` are pre-existing — verified by linting the unmodified file from `HEAD`, which reports the same 2).

Also verified that `auth.jsx` fetches output profiles at app bootstrap, so the new select is populated on the DVR tab rather than only after visiting Stream Settings.

One thing worth flagging for anyone writing further tests here: `CoreSettings._get_group` is Redis-cached, and that cache is **not** rolled back with the test transaction, so a value stored by one test is still readable in the next. The new setting tests delete the `dvr_settings` row in `setUp` to fire the post-delete invalidation.

## Checklist

- [x] Backend: migrations are included if any models were changed (N/A — no model changes; the setting is a key in the existing `dvr_settings` JSON group)


SergeantPanda (approved):
This looked good overall.

I made a couple of minor changes to it:
1) Renamed key to just output_profile_id
2) Added a cleanup for when output profiles are deleted, it will remove it from the DVR selection (this was added AFTER your PR for the HDHR path, nothing you missed)
3) Some minor verbiage changes.

Thank you for the PR!

Issue #970, opened 2026-02-12
Dummy EPG sources should create per-channel EPGData entries instead of one shared entry


xBOBxSAGETx:
## Problem

When you create a dummy EPG source, the signal in `apps/epg/signals.py` auto-creates a single shared EPGData entry (e.g. `dummy_sports__us_sports`). Every channel assigned to that source points to this one entry, which means they all share the same `tvg_id` in the XMLTV output.

This causes cross-group EPG contamination. If you have NBA, NCAAF, MLB, and ESPN+ channels all assigned to a single "US Sports" dummy source, they all get the same programme listings in the EPG output. Clients like TiviMate/Plex see identical EPG across completely different sports.

The dummy EPG generator itself is per-channel (it uses the channel's stream name to extract teams/times), but because the EPGData entry is shared, the XMLTV output collapses them all under one `<channel id="">`.

This will also affect XC API emulation since it pulls from the same EPGData.

## Current Workaround

I wrote a script that creates per-channel EPGData entries manually (`dummy-{sport}-{channel_id}` as tvg_id), so each channel gets its own unique entry on the source. Works great but has to be re-run after every M3U/XC refresh when new channels come in.

## Proposed Fix

When a channel is assigned to a dummy EPG source (either via the UI, auto-matching, or M3U import), create a dedicated EPGData entry for that channel instead of reusing the shared one. Something like:

```python
# Instead of pointing all channels to the single auto-created EPGData:
epg_data = EPGData.objects.get(epg_source=dummy_source)  # shared

# Create per-channel:
epg_data, _ = EPGData.objects.get_or_create(
    tvg_id=f"dummy-{source.id}-{channel.id}",
    epg_source=dummy_source,
    defaults={"name": channel.name}
)
```

The signal that auto-creates the shared entry on source creation could stay (it's harmless), but the channel assignment path should create per-channel entries.

This would eliminate cross-contamination permanently without needing external scripts.

SergeantPanda:
You are correct, if you use tvg-id as the source for tvg-id instead of the default of channel number this could be a problem.

PR #1622, opened 2026-08-28
fix(epg): skip dummy sources in bulk set-from-EPG channel actions

hdimer:
## Description

Creating a dummy EPG source auto-creates a single `EPGData` row (`apps/epg/signals.py` → `create_dummy_epg_data`), and every channel assigned to that source points at that one row. Its `tvg_id` (`dummy_<source_name>`) and its `name` describe the *source*, not any individual channel.

Two bulk actions copy fields off `channel.epg_data` onto the channel:

- **Set TVG-IDs from EPG** (`set_channels_tvg_ids_from_epg`, `apps/channels/tasks.py`)
- **Set names from EPG** (`set_channels_names_from_epg`, same file)

Run either over a selection of dummy-assigned channels and they all get overwritten with the same shared placeholder. For tvg_id that is #970: with `tvg_id_source=tvg_id`, the XMLTV export then emits one `<channel id="dummy_us_sports">` for what used to be NBA, MLB, NCAAF and ESPN+, so every client shows identical listings across them. The name task has the same defect and renames every affected channel to the source's own name.

Both tasks now exclude dummy-source channels from the queryset:

```python
channels = Channel.objects.filter(id__in=batch_ids).select_related(
    'epg_data'
).exclude(epg_data__epg_source__source_type='dummy')
```

This mirrors `apps/epg/api_views.py`, which already partitions channels with `epg_data__epg_source__source_type='dummy'` and deliberately uses `str(channel.uuid)` rather than the shared dummy `tvg_id`. Doing it in the queryset rather than in the loop means dummy channels are never fetched or hydrated at all, which matters for exactly the workload in #970 (a large dummy-assigned selection).

`EPGData.epg_source` and `Channel.epg_data` are both nullable. Django keeps NULL rows across a negated nullable join (`NOT (source_type = 'dummy' AND source_type IS NOT NULL)` over LEFT OUTER JOINs), so channels with no EPG data and `EPGData` rows with no source still behave as before. There's a test pinning that.

Two things this deliberately does **not** do:

- It does not implement #970's proposed per-channel `EPGData` design. This closes the corruption path; the shared-row design is a bigger call and yours to make.
- It cannot repair installs that already ran the action. The original per-channel tvg_ids were overwritten in place and aren't recoverable from the DB.

One thing I'd fix in a follow-up if you want it: the confirmation dialog in `ChannelBatch.jsx` still says "This will replace the current TVG-IDs with the TVG-IDs from their assigned EPG data", which is now slightly inaccurate for dummy-assigned channels. I left the frontend out of a backend bugfix rather than guess at the wording.

Happy to drop the `set_channels_names_from_epg` half into its own PR if you'd rather keep this pinned to #970 literally. It's the identical defect two functions apart, so it seemed worse to leave it.

I used an AI assistant while working on this (finding the twin in the name task, and the suggestion to move the filter into the queryset). I've read and verified every line, and I can explain the NULL-join semantics, the `select_related` choice, and each test if you want to dig in.

## Related Issue

Related to #970

## How was it tested?

New file `apps/channels/tests/test_set_channels_from_epg.py`, 5 tests:

| test | pins |
| --- | --- |
| `test_copies_tvg_id_from_a_real_epg_source` | normal tvg_id copy still works |
| `test_copies_tvg_id_when_epg_data_has_no_source` | `EPGData.epg_source = None` rows survive the `exclude` |
| `test_leaves_dummy_epg_channels_tvg_ids_alone` | two dummy-assigned channels keep their own distinct tvg_ids (#970) |
| `test_copies_name_from_a_real_epg_source` | normal name copy still works |
| `test_leaves_dummy_epg_channel_names_alone` | two dummy-assigned channels keep their own names |

Reproduced first, on `dev` before touching anything: two channels on one dummy source "US Sports", run the task, both come back `dummy_us_sports`.

Every branch was mutation-tested rather than just run green:

- Revert `tasks.py` only → both dummy tests fail (`'dummy_us_sports' != 'nba.us'`), the other three pass. The bug tests are red for the right reason.
- Swap `.exclude(...)` for an in-loop `if epg_source and epg_source.source_type == 'dummy': continue`, then drop the `epg_source and` half → `test_copies_tvg_id_when_epg_data_has_no_source` fails. That test is load-bearing, not decoration.
- Printed the generated SQL to confirm both joins are LEFT OUTER and the `IS NOT NULL` term sits inside the negation.

Suite: `apps.channels.tests` (which is what `scripts/ci_backend_test_labels.py` selects for these two paths) — 499 tests, same 8 errors as clean `dev` at 494 tests, so no new failures. Those 8 are `test_logo_path_jail` / `test_recording_output_path` / `test_catchup_utils` failing on `/data` not existing outside the container; they're identical with and without this change.

Local caveat, for transparency: I ran on sqlite (`TEST_USE_SQLITE=1`) with migrations disabled, because several older migrations use Postgres-only raw SQL (`jsonb ~`, `ON CONFLICT`) that won't run on sqlite even against an empty DB. Tables were built from current model state. CI's Postgres run is the authoritative one. Nothing in this change is backend-specific.

## Checklist

- [x] Backend: migrations are included if any models were changed — no model changes
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema — no new endpoints
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) — n/a, no frontend files touched

---
Used AI assistance on this; I reviewed and tested the change myself.

### DIFF
--- a/apps/channels/tasks.py
+++ b/apps/channels/tasks.py
@@ -4022,8 +4022,12 @@ def set_channels_names_from_epg(self, channel_ids):
             batch_ids = channel_ids[i:i + batch_size]
             batch_updates = []
 
-            # Get channels and their EPG data
-            channels = Channel.objects.filter(id__in=batch_ids).select_related('epg_data')
+            # Get channels and their EPG data. A dummy source shares one placeholder
+            # EPGData row across all its channels, so copying from it would collapse
+            # every channel on that source onto the same value.
+            channels = Channel.objects.filter(id__in=batch_ids).select_related(
+                'epg_data'
+            ).exclude(epg_data__epg_source__source_type='dummy')
 
             for channel in channels:
                 try:
@@ -4285,8 +4289,12 @@ def set_channels_tvg_ids_from_epg(self, channel_ids):
             batch_ids = channel_ids[i:i + batch_size]
             batch_updates = []
 
-            # Get channels and their EPG data
-            channels = Channel.objects.filter(id__in=batch_ids).select_related('epg_data')
+            # Get channels and their EPG data. A dummy source shares one placeholder
+            # EPGData row across all its channels, so copying from it would collapse
+            # every channel on that source onto the same value.
+            channels = Channel.objects.filter(id__in=batch_ids).select_related(
+                'epg_data'
+            ).exclude(epg_data__epg_source__source_type='dummy')
 
             for channel in channels:
                 try:
--- a/apps/channels/tests/test_set_channels_from_epg.py
+++ b/apps/channels/tests/test_set_channels_from_epg.py
@@ -1,0 +1,108 @@
+"""Tests for the bulk "set channel TVG-IDs / names from EPG" tasks."""
+from django.test import TestCase
+
+from apps.channels.models import Channel
+from apps.channels.tasks import (
+    set_channels_names_from_epg,
+    set_channels_tvg_ids_from_epg,
+)
+from apps.epg.models import EPGData, EPGSource
+
+
+class SetChannelsFromEpgTests(TestCase):
+    def setUp(self):
+        self.source = EPGSource.objects.create(
+            name="XML EPG",
+            source_type="xmltv",
+            url="http://example.com/epg.xml",
+        )
+        # Creating a dummy source auto-creates a shared EPGData row
+        # (apps/epg/signals.py: create_dummy_epg_data) whose tvg_id and name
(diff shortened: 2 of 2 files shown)


SergeantPanda (approved):
This looks good. I went back and forth a bit on this but ultimately, until dummy epgs get reworked, this is an OK stop-gap.

Thank you!

PR #1624, opened 2026-08-28
fix(vod): keep VOD movie/series IDs stable across M3U refresh (#961)

### LINKED ISSUE #961: VOD movie IDs change on every M3U refresh, breaking XC-compat STRM URLs

rarf:
## Summary
Fixes #961 — VOD movie/series IDs change on every M3U refresh, breaking XC-compat STRM URLs and any tmdb-keyed grouping.

When a provider does **not** supply a stable `tmdb_id`/`imdb_id` in its list payload, `process_movie_batch` / `process_series_batch` re-pointed an existing `M3UMovieRelation`/`M3USeriesRelation` to a freshly built object on every refresh. Because the incoming stream had no external id, the existing-record lookup missed, a brand-new `Movie`/`Series` row (new auto-increment id) was created, and the relation was repointed to it. Over repeated refreshes this spawned duplicate VOD rows and churned every consumer keyed on `Movie.id`/`Series.id`.

## Changes
- Never re-point an existing relation to the freshly built object — keep the relation's current linked movie/series (the stable record).
- Back-fill `tmdb_id`/`imdb_id` onto the **linked** (stable) record when the incoming data now carries them and the record lacks them (root cause #1: ids in `basic_data` but never persisted on initial import).
- Mark the freshly built object as superseded so the bulk-create step does not insert a duplicate row.

This does not trust that a `stream_id` was not reused by the provider: the linked record is unchanged, so a reused `stream_id` simply updates its relation rather than orphaning/duplicating the movie.

## Relation to previous work
This is a re-submission of the root-cause fix from #973 (closed by inactivity). The maintainer's only objection there — blindly trusting `stream_id` — is addressed here because the linked record is never replaced. Targets `dev` per #973's base.

## Test plan
- New regression tests: `apps/vod/tests/test_vod_stable_ids_on_refresh.py` (3 tests).
- Verified failing on unpatched `main` (`AssertionError: 2 != 1` — duplicate movie created) and passing with the fix.
- Real-world validation: a 112k-movie / 570k-relation instance showed 721 duplicate `(name, year)` Movie rows and 99% of movies with `null` `tmdb_id` — the exact signature this fix prevents.

## Evidence
- Issue #961 (closed as "completed" via the read-side fallback #1315 only; the import-time root fix was never merged).
- This fix targets the import path, complementing #1315.

### DIFF
--- a/.github/workflows/issue-template-check.yml
+++ b/.github/workflows/issue-template-check.yml
@@ -26,6 +26,6 @@ jobs:
             ## Issue not opened from a template
 
 
-            This issue was closed because it was not opened using one of the available issue templates.
+            This issue was closed because it was not opened using one of the available issue templates or was opened from the CLI instead of GitHub's web interface.
 
             Please [open a new issue]({new-issue-url}) and select the appropriate template. This helps us triage and address issues efficiently.
--- a/.github/workflows/pr-freshness.yml
+++ b/.github/workflows/pr-freshness.yml
@@ -1,5 +1,3 @@
-name: PR freshness
-
 on:
   schedule:
     - cron: '0 16 */2 * *'
--- a/.github/workflows/pr-freshness.yml
+++ b/.github/workflows/pr-freshness.yml
@@ -1,0 +1,43 @@
+name: PR freshness
+
+on:
+  schedule:
+    - cron: '0 16 */2 * *'
+  workflow_dispatch:
+
+jobs:
+  freshness:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/create-github-app-token@v1
+        id: app-token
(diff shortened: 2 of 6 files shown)


SergeantPanda (changes requested):
Please rebase this on dev and accept the CLA for this to be looked at.

SergeantPanda:
Closing as stale due to no response, no cla signed and needing a rebase.

You may open another PR scoped to dev if you'd like to see this looked at.

PR #1627, opened 2026-08-29
feat(users): add sort and search to the Users table

### LINKED ISSUE #1626: [Feature]: Sort and search on the Users page

The Users page renders every account in a fixed `id` order, with no way to reorder or narrow the list. There is no search box, and the column headers are plain text with no sort control.

On a small install that's fine. On one with a few dozen accounts it stops being usable: finding one user means scrolling the whole table and reading it manually, and questions like "who has never logged in?", "who did I add most recently?" or "which accounts are Admins?" can't be answered from the page at all — the data is all there in the columns, it just can't be ordered.

Every other data-heavy table in the app (M3U Accounts, EPGs, Streams, Channels, Logos) already offers sorting and/or filtering. Users is the outlier.

Client-side sort and search on the Users table — no API changes, since `/api/accounts/users/` already returns the full list the page renders.

- **Sort** on User Level, Username, Name, Email, Date Joined and Last Login, using the same header sort control the M3U and EPG tables use (click cycles ascending → descending → unsorted). XC Password, Channel Profiles and Actions stay unsorted.
- **Search** — a debounced box in the existing toolbar, matching username, full name, email and the user level label, case-insensitively. Sorting applies within the search results.
- Users with no value for the sorted column (blank name, never logged in) sort last in both directions rather than clumping at one end.
- An empty-state message when a search matches nothing.

Two deliberate choices worth flagging for review:

- Search does **not** match `custom_properties.xc_password`. The column masks it behind `••••••••`, so making it searchable would let someone confirm a password without ever revealing it.
- `user_level` sorts on its numeric value (0 Streamer / 1 Standard / 10 Admin) so the order is by privilege. Sorting the labels alphabetically would give Admin → Standard User → Streamer, which reads as arbitrary.

### Additional context

Happy to open the PR — I have this implemented and tested against `dev` and can push it as soon as the approach is agreed. It's frontend-only and follows the existing table patterns (`sortable` column flag, `makeHeaderCellRenderer` / `makeSortingChangeHandler`, a `UsersTableUtils.js` alongside the other `utils/tables/*Utils.js` files).

One small shared-code change is needed: `makeHeaderCellRenderer`'s `Group` currently wraps, and the Date Joined column ends up a pixel or two too narrow for label + sort icon, which drops the icon onto a second row and doubles the header height. Setting the group to `nowrap` with a smaller gap fixes it; that also affects the M3U and EPG header rows, which I've checked render correctly with it.

If you'd rather this were server-side ordering/filtering on the users endpoint instead of client-side, say so and I'll rework it — client-side seemed right given the page already fetches every user in one call and installs don't have huge user counts.

brendongl:
## Description

Adds client-side sorting and search to the Users page. Frontend only — no API, model or migration changes.

- **Sort** on User Level, Username, Name, Email, Date Joined and Last Login. Clicking a header cycles ascending → descending → unsorted, matching the M3U and EPG tables. XC Password, Channel Profiles and Actions are not sortable.
- **Search** — a debounced (300 ms) box in the existing toolbar matching username, full name, email and the user level label, case-insensitively. Sorting applies within the search results.
- Empty-state message when a search matches nothing.

The page previously rendered every account in fixed `id` order with no way to reorder or narrow it, which stops scaling at a few dozen users. Every other data-heavy table in the app already sorts and/or filters; Users was the outlier.

### How it works

**Sorting is done on the data, not by TanStack.** `useTable` registers only `getCoreRowModel`, and nothing in the frontend registers `getSortedRowModel` or `getFilteredRowModel` — so every table that sorts does it by sorting its own array and passing `manualSorting: true`. This follows that: rows are derived in one `useMemo` (filter first, since it shrinks what has to be sorted, then sort) and the table is told not to reorder or re-filter them.

Columns opt in with the existing custom `sortable` flag, and the header cells reuse `makeHeaderCellRenderer` / `makeSortingChangeHandler` from `M3uTableUtils` rather than reimplementing the click cycle.

Non-obvious decisions:

- **`getSortedUsers` copies before sorting.** The previous `data` memo called `users.sort(...)`, which sorts the Zustand store's own array in place. That was pre-existing, but the new code sorts far more often, so the copy matters.
- **`user_level` is compared on its raw numeric value** (0 / 1 / 10), so the order is by privilege. Sorting the labels alphabetically would give Admin → Standard User → Streamer.
- **The Name column has no model field** — it's `first_name + last_name`. The cell renderer, the sort and the search all go through `getUserFullName` so they can't drift apart.
- **Empty values sort last in both directions**, rather than leading in one. A blank name or a "Never" last-login isn't meaningfully smallest-or-largest, and 25 of the 84 accounts on my own install have no name set — having them swap between the top and bottom of the list on every direction toggle is noise.
- **Search deliberately excludes `custom_properties.xc_password`.** The column masks it behind `••••••••`; making it searchable would let someone confirm a password without ever revealing it.

### Changes outside `UsersTable.jsx`

Two, both in `M3uTableUtils.jsx`, both required by this feature rather than incidental cleanup — calling them out explicitly since they touch the M3U and EPG header rows:

1. `makeSortingChangeHandler`'s `onDataSort` callback is now optional. M3Us/EPGs keep their rows in state and re-sort them in that callback; this page derives rows from `sorting`, so it has nothing to pass. Making the argument optional was preferable to duplicating the three-click cycle logic. Existing callers are unaffected.
2. `makeHeaderCellRenderer`'s `Group` is now `wrap="nowrap"` with `gap="xs"`. With the default wrap and 16 px gap, the Date Joined header needed 105 px (75 label + 16 gap + 14 icon) in a 104 px column and dropped the sort icon onto a second row, doubling the header height. I checked the M3U and EPG header rows render correctly with the tighter grouping. Date Joined's own width also went 90 → 120 since it now carries a control.

If you'd rather these were not in this PR, I can inline a Users-specific header renderer instead — but that duplicates the cycle logic, so I went with the shared change.

## Related Issue

Closes #1626

## How was it tested?

**Automated**

- `npm run test` — **6305 tests pass across 207 files**, including the existing M3U and EPG table suites that share the changed renderer.
- New `frontend/src/utils/tables/__tests__/UsersTableUtils.test.js` (21 tests): sort direction, case-insensitive collation, that sorting does not mutate the input array, the virtual name column, numeric level ordering, null/blank-last in both directions, and that the XC password is not searchable.
- `frontend/src/components/tables/__tests__/UsersTable.test.jsx` extended 46 → 59 tests: search filtering, `allRowIds` staying in step with the filtered rows, the debounce, the empty state, the clear button, the sort cycle, sort replacing rather than stacking, non-sortable columns having no control, and sorting within search results.

**Manual, in a browser**

Ran a dev container built from this branch (`DISPATCHARR_ENV=dev`, vite + HMR) and drove the page with Playwright:

- Sort cycle asc → desc → off on all six sortable columns; verified the rendered row order each time.
- Search matching username, surname-only, email domain and user level label; verified case-insensitivity in both directions (a lowercase query matching a capitalised username and vice versa); verified `xc_password` values return no matches.
- Empty state, and the clear (×) button restoring the full list.
- Measured the header layout in-page to confirm the Date Joined sort icon no longer wraps to a second row (all header groups 20 px tall, previously 50 px for that one column).

Then loaded a **real 84-account user list** (25 with no name set, 82 never logged in) into the container to exercise the null/blank paths on non-synthetic data: blanks and "Never" trail correctly in both sort directions, level sorts Streamer → Standard User → Admin, and a re-sort of 84 rows takes ~30 ms measured click-to-paint.

## Checklist

Two notes so those boxes aren't overstated:

- **Backend items are vacuous here** — no models and no endpoints were touched, so there is nothing to migrate or to appear in the OpenAPI schema.
- **On lint/format:** `npm run lint` reports 164 problems (108 errors, 56 warnings) on `dev`, and reports *exactly the same* 164 on this branch — this PR introduces none. My five touched files report a single warning, a `useMemo` dependency on `profileIdToName` in `UsersTable.jsx` that is pre-existing on `dev` and outside this change. There is no `format` script in `frontend/package.json` (the template and CONTRIBUTING both reference `npm run format`), so I verified formatting with `npx prettier --check` on the changed files instead; all new files are clean. Happy to add a `format` script in a separate PR if that would be useful.

---

*Disclosure: this PR was written with AI assistance (Claude). I have reviewed the diff and stand behind it — the reasoning behind each non-obvious decision is written up above, and I'm happy to talk through any line in review or rework anything you'd prefer done differently.*

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Users table sort and search.** The Users page can sort by User Level, Username, Name, Email, Date Joined, and Last Login (header click cycles ascending, descending, unsorted, matching M3U/EPG), and filter with a debounced search over username, full name, email, and user level label. Empty values sort last in both directions; user level sorts by privilege rather than label. - Thanks [@brendongl](https://github.com/brendongl)
+
+### Changed
+
+- **Shared table sort header helpers moved out of `M3uTableUtils`.** `makeHeaderCellRenderer` and `makeSortingChangeHandler` now live in `frontend/src/components/tables/tableSortingUtils.jsx`, since M3U, EPG, and Users all reuse them. Imports updated accordingly.
+
 ## [0.30.0] - 2026-08-29
 
 ### Added
--- a/frontend/src/components/tables/EPGsTable.jsx
+++ b/frontend/src/components/tables/EPGsTable.jsx
@@ -49,7 +49,7 @@ import {
 import {
   makeHeaderCellRenderer,
   makeSortingChangeHandler,
-} from './M3uTableUtils.jsx';
+} from './tableSortingUtils.jsx';
 
 // Helper function to get status text color
 const getStatusColor = (status) => {
--- a/frontend/src/components/tables/M3UsTable.jsx
+++ b/frontend/src/components/tables/M3UsTable.jsx
@@ -60,7 +60,7 @@ import {
 import {
   makeHeaderCellRenderer,
   makeSortingChangeHandler,
-} from './M3uTableUtils.jsx';
+} from './tableSortingUtils.jsx';
 
 const ALL_ACCOUNT_TYPES = ['STD', 'XC'];
 
--- a/frontend/src/components/tables/M3uTableUtils.jsx
+++ b/frontend/src/components/tables/M3uTableUtils.jsx
@@ -14,7 +14,10 @@ export const makeHeaderCellRenderer = (sorting, onSortingChange) => (header) => 
   }
 
   return (
-    <Group>
+    // nowrap keeps the sort control on the header's line: with the default
+    // wrap, a column only a pixel or two too narrow drops the icon onto a
+    // second row and doubles the header height.
+    <Group gap="xs" wrap="nowrap">
       <Text size="sm" name={header.id}>
         {header.column.columnDef.header}
       </Text>
@@ -46,7 +49,9 @@ export const makeSortingChangeHandler =
     }
 
     setSorting(newSorting);
-    if (newSorting.length > 0) {
+    // Optional: tables that keep their rows in state re-sort them here. Tables
+    // that derive their rows from `sorting` don't pass a callback.
+    if (onDataSort && newSorting.length > 0) {
       onDataSort(newSorting[0].id, newSorting[0].desc);
     }
   };
--- a/frontend/src/components/tables/UsersTable.jsx
+++ b/frontend/src/components/tables/UsersTable.jsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useCallback, useEffect, useState } from 'react';
+import React, { useMemo, useCallback, useState } from 'react';
 import API from '../../api';
 import UserForm from '../forms/User';
 import useUsersStore from '../../store/users';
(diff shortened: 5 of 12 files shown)


SergeantPanda (approved):
This looks good overall. I did take this as an opportunity to move the utils out of m3u and into a correctly named sort utils file.

I also cleaned up the comments a bit as well as used a shared useDebouce.

Thank you!

PR #1636, opened 2026-09-01
DVR: seamless HLS→MKV finalize (fix ~40 ms audio blips every 4 s in F…

torhve:
## Description

Use simpler concat to splice the HLS segments to avoid muxer timing problems with ffmpeg 

## Related Issue

Sorry, I have not opened an issue for this as I believed it to be a small fix, but with the gating and tests it grew much larger than I first anticipated.

## How was it tested?

I ran the new tests - and also ran ffmpeg simulations to verify both the problem and the fix before writing the code.

## Checklist


SergeantPanda:
Really good job diagnosing this, thank you!
I was able to reproduce what you’re seeing as well. The concat demuxer really is the wrong tool here.
I’d rather make the primary finalize path a straight remux from `index.m3u8` (`ffmpeg -i index.m3u8 -c copy …`) instead of bit-concatenating the segments. Same seamless timeline in my testing, no third on-disk copy of the recording, and a bit faster in benches.

I’ll close this PR and open a follow-up; you’ll get credit for the diagnostic work.

This is the commit: 64c61f31b311b9066722ead92c8792ff64f9bdc0

Issue #1256, opened 2026-05-21
[Bug]: Stale profile_connections:* Redis counter causes false “maximum connection limits” and 503 stream failures


mwongj:
## Environment
Dispatcharr Docker image: ghcr/dispatcharr/dispatcharr:dev
App version: 0.24.0
Build timestamp: 20260503204152
Redis: 8.6.2
Deployment: AIO Docker container on Unraid
Container started: 2026-05-06T03:22:19Z

##  Summary
Dispatcharr rejected IPTV stream requests with:

`All active M3U profiles have reached maximum connection limits`

but the profile was not actually at capacity. Redis had a stale profile_connections:<profile_id> value of 5 while only one active live stream reservation existed for that profile.

Relevant profile configuration

|account_id | account       | account_max | profile_id | profile               | profile_max | active|
|--|--|--|--|--|--|--|
|2          | iptv | 5           | 2          | iptv Default | 5           | true|
|3          | HDHR5-4US     | 4           | 3          | HDHR5-4US Default     | 4           | true|
|4          | HDFX-4K       | 4           | 4          | HDFX-4K Default       | 4           | true|

## Failure logs

2026-05-21 00:43:51,954 INFO ts_proxy.views Requested stream for channel 6f32f89f-c297-4beb-a67e-c08003c96ae1
2026-05-21 00:43:51,962 ERROR ts_proxy.url_utils No stream available for channel 6f32f89f-c297-4beb-a67e-c08003c96ae1: All active M3U profiles have reached maximum connection limits
2026-05-21 00:43:54,607 INFO ts_proxy.views Failed to obtain stream after 13 attempts over 2s: All active M3U profiles have reached maximum connection limits
2026-05-21 00:43:54,608 ERROR django.request Service Unavailable: /proxy/ts/stream/6f32f89f-c297-4beb-a67e-c08003c96ae1
2026-05-21 00:43:51,000 INFO uwsgi.requests Worker ID: 4 GET 503 /proxy/ts/stream/6f32f89f-c297-4beb-a67e-c08003c96ae1 2657ms

##  Repeated again:

2026-05-21 00:44:16,427 INFO ts_proxy.views Requested stream for channel 6f32f89f-c297-4beb-a67e-c08003c96ae1
2026-05-21 00:44:16,436 ERROR ts_proxy.url_utils No stream available for channel 6f32f89f-c297-4beb-a67e-c08003c96ae1: All active M3U profiles have reached maximum connection limits
[...]

Issue #1577, opened 2026-08-20
[Bug]: VOD profile_connections counter is never reclaimed at runtime, so one ended playback permanently consumes an account with max_streams=1


PiratesIRC:
### Describe the bug

On 0.29.0, a VOD playback that ends leaves `profile_connections:<profile_id>`
incremented with nothing behind it, and nothing in the running application ever
reclaims it. With `max_streams=1` on the account, that single leaked count makes
the account permanently unusable for both VOD and live until the container is
restarted. Every later request gets HTTP 503, and the client shows a generic
playback error.

I believe this is the specific code path a maintainer predicted but never
located. In pull request 1290, SergeantPanda wrote: "Dispatcharr flushes Redis
on startup already, which would clear these counters. I do not believe this is
the fix for the linked issue here. More than likely there is a bug with not
releasing a connection. My money would be on VOD somewhere." That pull request
was closed without merging, and issue 1256 was closed as completed, but the
release bug itself was never identified. What follows is that bug.

## The code path

All line numbers are from the 0.29.0 image
`ghcr.io/dispatcharr/dispatcharr:latest`, file
`/app/apps/proxy/vod_proxy/multi_worker_connection_manager.py`.

1. The session hash EXPIRES, the counter does NOT.
   `RedisBackedVODConnection._save_connection_state` sets a one-hour expiry on
   the session hash at line 333:
   `self.redis_client.expire(self.connection_key, 3600)`.
   The counter written by `_get_profile_connections_key` has no expiry and no
   session id embedded in it. Once the hash expires, or was never written
   because the client closed during setup, the counter is orphaned: there is no
   longer any record that says which session it belonged to.

2. The only code that scans for stale sessions is NEVER CALLED.
   `cleanup_stale_persistent_connections` is defined at line 1490. I searched
   every Python file in `/app` (505 files) for that name. It appears exactly
   once: its own definition. Nothing calls it, no Celery beat schedule
   references it, and there is no other periodic reaper.

3. Even if it were called, it could not fix this.
   It scans `vod_persistent_connection:*` and cleans up sessions it finds
   there. An orphaned counter has no such key, so the scan finds nothing to
   clean and the counter survives.

4. There is no startup reconciliation either.
   The string `reconcile` does not appear anywhere in `/app`. Pull request 1290
   would have added startup reconciliation and was closed unmerged.

The result is that `profile_connections:<id>` is a monotonically increasing
[...]

PR #1637, opened 2026-09-01
fix(vod): strict-client XC payloads, session cleanup race, self-termination and capacity races around VOD playback

cedric-marcoux:
## Description

Five stacked fixes around XC VOD/series playback, found while debugging why iPlayTV (Apple TV) could not play series episodes. They compound: strict clients first fail on the JSON payload; once past that, rapid range/probe request patterns (VLC-based players, and TiviMate/Emby-style players at start-up) trip races and the user's own stream-limit enforcement, producing black screens, `profile_connections` slot leaks, and provider request storms.

1. **`custom_sid: null` in XC VOD/series payloads** (`apps/output/views.py`) — strict XC clients (iPlayTV) reject JSON `null` where a string is expected; real panels emit `""`. The live-channel path already sends `""`; the three VOD paths (`xc_get_vod_streams`, `xc_get_series_info`, `xc_get_vod_info`) still sent `null`. Also guarded `stream_icon`/`cover` (`_xc_cover_or_logo` returns `None` when a movie/series has no artwork and no logo).
2. **Episode ids emitted as JSON numbers** (`apps/output/views.py`) — real XC panels emit episode ids as strings; iPlayTV's decoder rejects numbers there and never even requests the stream. Now `str(episode.id)` for `id` and `info.id` in `get_series_info`.
3. **Delayed-cleanup race destroys session state under a live request** (`multi_worker_connection_manager.py`) — the 1s-delayed cleanup after each completed range request hard-deletes the shared session state. A client chaining rapid ranges can be mid-flight between profile-slot reserve and stream INCR on that same session when it fires: the in-flight request dies (`INCR-AS failed: no state`) and its reserved profile slot is never decremented — a permanent `profile_connections` ghost that blocks the whole account until Redis is reset. Fix: skip the hard delete while the session saw activity in the last 20s; the existing key TTL still reaps abandoned sessions.
4. **User stream-limit enforcement terminates the user's own playback** (`apps/proxy/utils.py`) — `check_user_stream_limits` has a same-channel exemption for live and a sibling exemption for timeshift (a9087207), but nothing for VOD. With `stream_limit=1`, the viewer's own parked same-content session counts against their next range request, so enforcement terminates their own stream every ~800KB (the stop-key poll interval), forcing an endless reconnect loop that also hammers the provider (we observed provider-side 429 rate limiting as a direct consequence). Fix: same-content VOD sibling exemption, mirroring the existing live/timeshift precedents.
5. **Seek/start-up burst gets an instant 503 while the viewer's own stream holds the slot** (`apps/proxy/vod_proxy/views.py`) — FF/RW opens a new connection before the old one tears down; players also open several parallel probe connections at start-up which hold the slot for well under a second. With `max_streams=1`, selection failed instantly → 503 → players abort playback entirely. Fix: on capacity failure, signal the viewer's own same-content stream to stop (timeshift-scrub-style takeover), then retry selection for up to 3s before giving up; displacement signals are consumed afterwards so a reused session id never reads a stale kill order.

No behavior change for other clients: fixes 1–2 align the payload with what real panels emit (every client already handles that), and fixes 3–5 only act on paths that were already failures (instant 503 / killed stream / leaked slot). Three existing assertions in `apps/output/tests/test_views.py` that locked in the `null` payload values were updated to the new `""` values.

Happy to split this into separate PRs if you prefer — the five changes are in distinct code paths but were found and validated together.

## Related Issue

Related: #1256 — fix 3 is the mechanism behind the stuck `profile_connections` counters reported there (reproduced live: reserve at capacity → delayed cleanup deletes session state under the in-flight request → its slot is never decremented). Not marking it as closed since that issue was already closed, but this removes the leak path we could reproduce.

## How was it tested?

- Built a fresh AIO image from this branch (`docker build -f docker/Dockerfile` succeeds) and ran the backend suites in a disposable container with its own database:
  `python -m django test apps.timeshift.tests.test_helpers apps.timeshift.tests.test_sessions apps.timeshift.tests.test_stats apps.timeshift.tests.test_views apps.output.tests.test_xc_vod_info apps.output.tests.test_views apps.output.tests.test_xc_series_info apps.proxy.tests.test_stream_limits` → **424 tests, OK**.
- 4 new regression tests included: strict-client series payload (walks the whole episodes payload asserting no `null` anywhere, string `id`/`info.id`, `custom_sid == ""`) and `VodSiblingLimitExemptionTests` (same-content VOD sibling allowed without termination; different content still enforced; live connections unaffected).
- Live-validated the equivalent patches on a production 0.30.0 setup with two XC provider accounts (`max_streams=1` each):
  - iPlayTV/Apple TV: series episode now starts (previously the app crashed parsing the payload and never requested the stream), and 4 consecutive FF jumps (21.9MB → 677.8MB → 1.59GB → 2.03GB byte offsets) were each served at the exact requested position in ~0.6s on the reused session.
  - TiviMate/Android: series start-up burst (3-4 parallel probe connections) no longer gets an instant 503; playback starts and runs on a single long-lived provider request.
  - Verified the `profile_connections` leak no longer reproduces under rapid range churn (counter returns to 0 after playback, no ghost slots, no stale stop keys), where before we could reproduce a stuck counter within seconds.

## Checklist

### DIFF
--- a/apps/output/views.py
+++ b/apps/output/views.py
@@ -1335,7 +1335,7 @@ def xc_get_vod_streams(request, user, category_id=None):
                 logo_id=row['movie__logo_id'],
                 logo_url_parts=_logo_url_parts,
                 url_parts=_movie_image_parts,
-            ),
+            ) or "",
             "rating": rating or "0",
             "rating_5based": round(float(rating or 0) / 2, 2) if rating else 0,
             "added": str(int(row['movie__created_at'].timestamp())),
@@ -1352,7 +1352,7 @@ def xc_get_vod_streams(request, user, category_id=None):
             "category_id": category_id_str,
             "category_ids": category_id_list,
             "container_extension": row['container_extension'] or "mp4",
-            "custom_sid": None,
+            "custom_sid": "",
             "direct_source": "",
         })
 
@@ -1429,7 +1429,7 @@ def xc_get_series(request, user, category_id=None):
                 logo_id=row['series__logo_id'],
                 logo_url_parts=_logo_url_parts,
                 url_parts=_series_image_parts,
-            ),
+            ) or "",
             "plot": row['series__description'] or "",
             "cast": custom_props.get('cast', ''),
             "director": custom_props.get('director', ''),
@@ -1569,16 +1569,18 @@ def xc_get_series_info(request, user, series_id):
         )
 
         seasons[season_num].append({
-            "id": episode.id,
+            # Real XC panels emit episode ids as strings; strict clients
+            # (e.g. iPlayTV) reject JSON numbers where they expect strings.
+            "id": str(episode.id),
             "season": season_num,
             "episode_num": episode.episode_number or 0,
             "title": episode.name,
             "container_extension": container_extension,
             "added": added_timestamp,
-            "custom_sid": None,
+            "custom_sid": "",
             "direct_source": "",
             "info": {
-                "id": int(episode.id),
+                "id": str(episode.id),
                 "name": episode.name,
                 "overview": episode.description or "",
                 "crew": str(episode.custom_properties.get('crew', "") if episode.custom_properties else ""),
@@ -1882,7 +1884,7 @@ def xc_get_vod_info(request, user, vod_id):
             "category_id": str(movie_relation.category.id) if movie_relation.category else "0",
             "category_ids": [int(movie_relation.category.id)] if movie_relation.category else [],
             "container_extension": movie_relation.container_extension or "mp4",
-            "custom_sid": None,
+            "custom_sid": "",
             "direct_source": "",
         }
     }
--- a/apps/output/tests/test_views.py
+++ b/apps/output/tests/test_views.py
@@ -689,7 +689,7 @@ class XcVodSeriesRegressionTests(TestCase):
         self.assertEqual(stream["stream_type"], "movie")
         self.assertEqual(stream["stream_id"], movie.id)
         self.assertEqual(stream["rating_5based"], 5.0)
-        self.assertEqual(stream["custom_sid"], None)
+        self.assertEqual(stream["custom_sid"], "")
         self.assertEqual(stream["direct_source"], "")
 
     def test_vod_streams_null_optional_fields(self):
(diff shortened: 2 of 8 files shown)


apps/output/views.py
```diff
@@ -1352,7 +1352,7 @@ def xc_get_vod_streams(request, user, category_id=None):
             "category_id": category_id_str,
             "category_ids": category_id_list,
             "container_extension": row['container_extension'] or "mp4",
-            "custom_sid": None,
+            "custom_sid": "",
```
SergeantPanda, line 1355: I tested several of my providers and every single one of them return none for custom_sid and not an empty string. I do not believe this change is correct.

Do you have more evidence to show a string is correct for `get_vod_streams`?

apps/output/views.py
```diff
@@ -1569,16 +1569,18 @@ def xc_get_series_info(request, user, series_id):
...
             "episode_num": episode.episode_number or 0,
             "title": episode.name,
             "container_extension": container_extension,
             "added": added_timestamp,
-            "custom_sid": None,
+            "custom_sid": "",
             "direct_source": "",
             "info": {
-                "id": int(episode.id),
+                "id": str(episode.id),
```
SergeantPanda, line 1583: I believe this is probably a correct change.

SergeantPanda (changes requested):
Yes please split these into separate PRs. I want to review each on its own and may not take all of them.

On **`custom_sid`**: I'd like confirmation that the player you tested actually rejects JSON `null` there. A lot of real panels (and the original XC `get_vod_streams` path) emit `null`, so if that client can't handle it, it already fails against many providers. If it only broke on episode `id` type / other fields, I'd rather leave `custom_sid` as `null` for panel fidelity. If you can show a hard failure on `null` specifically, great, include that evidence in the PR.

**`#3` (cleanup grace)** almost certainly needs to be reworked or dropped. There was a large `multi_worker_connection_manager` refactor on `dev` after this branch was cut (`#1577` / related session seeding). Please re-validate against current `dev` before spending more time on it.

Grouping I'd be happy with:
- **`#1` + `#2`** (XC payload / string episode ids) in one PR, once `custom_sid` is verified or omitted
- **`#4`** (VOD sibling stream-limit exemption) on its own
- **`#5`** (capacity displace / retry) on its own, rebased onto current `dev`
- **`#3`** only if it still reproduces after that rebase

Thanks!

PR #1642, opened 2026-09-02
Fix VOD category filter mis-splitting names that contain '|' (#1603)

### LINKED ISSUE #1603: [Bug]: VOD category filter (filter_category) mis-splits category names containing "|", causing all VOD results to return empty
### Describe the bug

`MovieFilter.filter_category` (and the equivalent series filter) in `apps/vod/api_views.py` incorrectly parses the `category` query param on `/api/vod/movies/` and `/api/vod/series/` when the category *name itself* contains a pipe character.

The filter is designed to accept an optional `name|type` suffix so callers can disambiguate duplicate category names (e.g. `Comedy|movie` vs `Comedy|series`). It does this with:

    if '|' in value:
        category_name, category_type = value.rsplit('|', 1)
        return queryset.filter(
            m3u_relations__category__name=category_name,
            m3u_relations__category__category_type=category_type
        )
    else:
        return queryset.filter(m3u_relations__category__name=value)

The problem: it only checks whether `|` appears *anywhere* in the value, not whether a type suffix was actually appended. Some IPTV providers name categories with an embedded pipe as part of the name itself, e.g. `|EN| 4K CLASSIC MOVIES`. When a client passes that literal category name with no type suffix, `rsplit('|', 1)` still splits on the last pipe in the string — which is part of the category name, not a real type separator — producing a `category_name`/`category_type` pair that matches nothing. The query silently returns `count: 0` even though matching content exists.

This affects any client using the native REST API directly (confirmed via Dispatcharr's own AerioTV "Direct Dispatcharr Integration" mode). Dispatcharr's Xtream-Codes-emulation endpoints (`apps/output/views.py`, e.g. `xc_get_vod_streams`) are unaffected — they filter by integer `category_id` and don't touch this function.

Expected: filtering by a category name that happens to contain "|" as part of the name (with no explicit type suffix) should match that category correctly, the same way it does for names without a pipe.

1. Have (or create) a VOD category whose name contains a literal "|" character, e.g. "|EN| 4K CLASSIC MOVIES" — common with some IPTV providers' locale-tagged category naming.
2. Confirm content exists: GET /api/vod/movies/?page_size=5 (no category filter) returns a nonzero count including movies belonging to that category.
3. Query with the category filter, using the exact category name and no explicit type suffix:
   GET /api/vod/movies/?category=%7CEN%7C+4K+CLASSIC+MOVIES
   Result: {"count": 0, "results": []}
4. Try the numeric category ID instead (e.g. category=18) — still {"count": 0} (rules out "needs an ID").
5. Now append an explicit, correctly-placed type suffix so rsplit lands on the intended pipe:
   GET /api/vod/movies/?category=%7CEN%7C+4K+CLASSIC+MOVIES%7Cmovie
   Result: {"count": 139, ...} — real matching movies returned.

Step 5 vs step 3 demonstrates the filter works once you (incorrectly) force it to split on the pipe you intend, and fails when the category name's own embedded pipe gets picked up instead.

### Dispatcharr Version

0.29.0

### Docker Version

29.7.2

### Client Information

AerioTV (Android TV app, https://github.com/jonzey231/AerioTV), version 0.4.20 (48) github, connection mode: "Dispatcharr Direct Connect" (native REST API + API key auth). Tested on NVIDIA SHIELD Android TV. Also reproduced directly via raw curl against the Dispatcharr REST API (bypassing AerioTV entirely), confirming this is server-side and not app-specific.

thejdubb02:
## Description

The VOD category filter returns no results for categories whose name contains `|`, for example `|EN| 4K CLASSIC MOVIES`.

The filter parses a `name|type` value with `value.rsplit('|', 1)` whenever a `|` is present, so a name that itself contains `|` (and has no type suffix) is split into a bogus name and type. In `MovieFilter` and the series filter that matches nothing; in `UnifiedContentViewSet.list` (`/api/vod/all/`) the resulting type matches neither `movie` nor `series`, and because there is no `else`, the category filter is silently dropped and everything is returned unfiltered.

The fix only treats the trailing token as a category type when it is one of `VODCategory.CATEGORY_TYPE_CHOICES` (`movie`, `series`); otherwise the whole value is the category name. So `|EN| 4K CLASSIC MOVIES` matches as a name, while `Action|movie` and `|EN| 4K|series` still split correctly. The parsing is extracted into `parse_category_filter_value()` in the existing Django-free `apps/vod/utils.py`, and all three category-parse sites use it.

## Related Issue

Closes #1603

## How was it tested?

- Added `apps/vod/tests/test_category_filter.py` (a `unittest.TestCase`, so `python manage.py test` collects it) covering: a plain name, `name|type`, a name containing `|` with no type suffix, and a name containing `|` with a type suffix. Run: `python manage.py test apps.vod.tests.test_category_filter`.
- Compared the new parser against the previous `value.rsplit('|', 1)` on the reported input and confirmed only the broken case changes: `|EN| 4K CLASSIC MOVIES` now stays a single name, while `Action|movie`, `|EN| 4K|series`, and `Documentaries` parse identically to before, so existing behaviour is preserved.
- `python -m py_compile` is clean on the changed files. The full celery/redis-backed backend suite runs in CI.

## Checklist

- [x] Backend: migrations are included if any models were changed (no models were changed)
- [x] Frontend: ESLint and Prettier pass cleanly (N/A: backend-only change)


SergeantPanda:
This looks good, thank you!

PR #1651, opened 2026-09-05
Prevent EPG edits from corrupting concurrent XMLTV exports

### LINKED ISSUE #1650: [Feature]: Safe EPG updates for plugins and concurrent XMLTV exports

I'd like to propose a supported way for plugins to update EPG data without corrupting concurrent XMLTV exports. The same mechanism could serve built-in imports, refreshes, and administrative edits.

In testing against **Dispatcharr v0.30.0** (`cd3f6c509d74ad7618d870e5673476928d528de9`), calling `invalidate_epg_chunk_cache()` after a committed programme update could leave malformed XMLTV in the cache. Waiting until the database transaction commits is necessary, but it does not coordinate invalidation with active exporters.

The race occurs as follows:

1. An XMLTV exporter begins generating a feed and appends chunks to Redis.
2. An EPG update commits and calls the invalidation helper.
3. Invalidation deletes the exporter's chunk list, status, and build lock.
4. The original exporter continues appending its remaining chunks, recreating an incomplete list.
5. The exporter marks that list ready.
6. Subsequent requests receive HTTP 200 with truncated XML, potentially missing the document header, opening element, and earlier programmes.

A controlled reproduction confirmed that the original request could finish with valid older data while subsequent requests received malformed cached XML. Valid output returned after the native cache TTL expired, approximately 300 seconds in that measurement. The TTL starts when the producer finishes, so this is not a universal bound from the database update.

This affects output integrity, beyond delayed visibility of an EPG update. XC XMLTV uses the same exporter and cache machinery; separate request keys can mask the problem when comparing endpoints.

**Reproduction outline**

1. Populate an EPG large enough to require multiple export chunks.
2. Start an HTTP XMLTV export and pause its producer after at least one programme-bearing chunk has been cached.
3. Commit a programme metadata update and call the existing invalidation helper.
4. Resume the exporter and request XMLTV again using the same cache key, including the same origin, profile, and relevant query parameters.
5. Parse both responses and compare programme counts.

The expected behavior would be complete, valid XML in both responses. An already-started request may finish with its earlier snapshot; later requests should eventually reflect the committed update.

Would a host-owned EPG mutation contract, backed by concurrency-safe cache invalidation, make sense here? One possible approach would be:

- **A documented mutation entry point.** Provide a service or transaction helper for bounded programme updates, including supported fields, conflict handling, and behavior when a refresh replaces rows. Plugins would not need to manipulate private Redis keys or patch exporters.
- **Commit-aware revisions.** Successful batches advance an EPG revision through a defined commit-safe mechanism. Rollbacks, previews, and no-op updates do not invalidate exports. Bulk updates participate explicitly rather than relying solely on model signals.
- **Separate identities for cache builds.** Store chunks under a generation and unique build identifier so consumers cannot observe chunks combined from different builds.
- **Atomic publication of completed builds.** A producer publishes only if it still owns the build and its revision remains eligible. An invalidated producer cannot make an older or incomplete build current.
- **Retention for active readers.** Invalidation retires a generation from selection by new requests without deleting data still needed by existing readers. Reader leases or another explicit retention mechanism would also protect active responses from cleanup and expiry.
- **Defined failure behavior.** Redis failures or unsuccessful publication should not promote partial output to a successful cached response. Cache-refresh failures should be visible for diagnosis and recovery, with a documented fallback that preserves XML integrity.

Cache generations address stream integrity; database snapshot consistency is a separate concern. It would also help to document whether an export represents one consistent EPG revision or permits bounded staleness across updates.

Existing import, refresh, and invalidation callers could share the same implementation. The immediate concern is the cache-integrity race; a documented plugin mutation API could follow separately if that makes the changes easier to review. These are implementation suggestions, and an alternative that provides the same safety guarantees would be welcome.

### Additional context

Relevant pinned implementation:

- [Cache producer and reader implementation](https://github.com/Dispatcharr/Dispatcharr/blob/cd3f6c509d74ad7618d870e5673476928d528de9/apps/output/streaming_chunk_cache.py#L96-L141)
- [EPG cache invalidation](https://github.com/Dispatcharr/Dispatcharr/blob/cd3f6c509d74ad7618d870e5673476928d528de9/apps/output/streaming_chunk_cache.py#L245-L263)
- [XMLTV exporter cache integration](https://github.com/Dispatcharr/Dispatcharr/blob/cd3f6c509d74ad7618d870e5673476928d528de9/apps/output/epg.py#L654-L659)

**Suggested validation**

Some scenarios that may be useful when evaluating a fix:

- Concurrent EPG updates during generation and while reading an already-cached feed, through both real XMLTV HTTP endpoints.
- Invalidated or expired producers attempting to publish after a newer generation exists.
- Rollbacks, previews, unchanged updates, and bulk operations.
- Concurrent row replacement without resurrecting deleted programmes.
- Worker interruption, Redis failure, and cache expiry without malformed output.
- Preservation of channel mappings, schedules, profile restrictions, hidden channels, and output permissions.
- A minimal documented plugin example explaining commit, conflict, and cache-visibility behavior.

This could give plugin authors a stable extension boundary while addressing a cache race that can affect any EPG writer.

dskvr:
## Description

Prevent committed EPG edits from corrupting concurrent XMLTV exports. Export selection now includes a durable database revision, and each Redis build has a unique identity with atomic ownership, chunk-count and publication checks. Invalidation retires a revision without deleting data used by active readers or producers.

Cache-miss producers still stream immediately. Concurrent followers wait up to five seconds for a completed build, then generate their own response if needed. Completed cache replays carry an exact Content-Length, so expired or missing data causes a detectable incomplete transfer rather than a normally completed truncated XML document. Memory remains bounded to one chunk.

Also adds a documented Python service for up to 500 expected-value programme metadata updates per batch, preserving unrelated properties and reporting conflicts/missing rows. Metadata changes and their revision commit together. Existing programme API writes advance the same revision atomically; stale PATCH requests cannot recreate rows deleted by an EPG refresh.

No new dependencies, HTTP endpoints, or frontend changes. The migration creates the EPG revision row. The implementation and tests were AI-assisted. Author has manually reviewed and tested these changes, ready for review. 

## Related Issue

resolves #1650

## How was it tested?

- `python manage.py test apps.output.tests apps.epg.tests apps.channels.tests apps.m3u.tests --keepdb -v1`: **1,230 tests passed** on PostgreSQL 17 and real Redis using the project CI bootstrap in a disposable container.
- Added failing regressions before the cache and API fixes. Real HTTP tests exercise both standard and XC XMLTV, active producers/readers during mutation, preview/no-op/rollback, API freshness, missing-row conflicts, and preservation of profile/user/hidden-channel/override behavior.
- Real Redis expiry/deletion tests verify incomplete cached transfers are detected. HTTP fault tests use Django's WSGI LiveServer with `Connection: close`; this is not a claim of production nginx/uWSGI deployment testing.
- `python manage.py makemigrations epg --check --dry-run`: no EPG migration drift. Python compilation and `git diff --check` passed. Independent review found no remaining high-severity correctness or security findings.

Existing requests do not become global database snapshots. Retired generations expire naturally; readers paused beyond retention can fail detectably. Followers may regenerate after the bounded wait.

## Checklist

- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (not applicable: no new endpoints)
- [x] Frontend: ESLint and Prettier pass cleanly (`npm run lint`, `npm run format`) (not applicable: no frontend changes)
- [x] Existing tests still pass (the four affected/backend caller suites above)

### DIFF
--- a/Plugins.md
+++ b/Plugins.md
@@ -528,3 +528,93 @@ class Plugin:
 - Model: `apps/plugins/models.py`
 - Frontend page: `frontend/src/pages/Plugins.jsx`
 - Sidebar entry: `frontend/src/components/Sidebar.jsx`
+
+## Updating EPG programme metadata
+
+Plugins running inside Dispatcharr can use `apps.epg.services.update_programmes`
+for bounded, transactional metadata updates. This is a Python service, not an
+HTTP endpoint. It does not grant additional permissions: plugins already execute
+with application privileges and must select the intended programmes themselves.
+
+```python
+from apps.epg.models import ProgramData
+from apps.epg.services import MISSING, ProgrammeUpdate, update_programmes
+
+programme = ProgramData.objects.get(pk=programme_id)
+if isinstance(programme.custom_properties, dict):
+    result = update_programmes([
+        ProgrammeUpdate(
+            programme_id=programme.pk,
+            expected={
+                "epg_id": programme.epg_id,
+                "start_time": programme.start_time,
+                "end_time": programme.end_time,
+                "title": programme.title,
+                "custom_properties": {
+                    "language": programme.custom_properties.get("language", MISSING),
+                },
+            },
+            values={"custom_properties": {"language": "en"}},
+        ),
+    ])
+    # result.changed, .unchanged, .conflicts, and .missing contain programme IDs.
+```
+
+The service accepts at most 500 updates with distinct positive integer IDs. It
+validates the entire batch before writing, then locks existing rows in ID order
+and compares each update's `expected` values with the current database row. A
+mismatch skips that programme and returns its ID in `conflicts`. A deleted or
+replaced row is returned in `missing`; the service never recreates it. Re-read
+conflicting or replaced programmes and reconsider the update rather than retrying
+with stale expectations. A source refresh can still replace metadata after a
+successful update; the API does not make plugin changes permanent overrides.
+
+Writable fields are `title`, `sub_title`, `description`, and `custom_properties`.
+Every writable scalar or property key included in `values` must also appear in
+`expected`. Additional expected conditions can include `epg_id`, `start_time`,
+`end_time`, `tvg_id`, and `program_id`. Timestamps must be timezone-aware Python
+`datetime` objects. Schedule times, source identity, and channel mappings cannot
+be changed through this service. Invalid requests raise Django `ValidationError`
+without applying any part of the batch.
+
+`custom_properties` is an explicit **top-level merge**. Unmentioned keys are
+preserved; a nested object supplied for a key replaces that key's whole value.
+Use `MISSING` in `expected` to require an absent key, and in `values` to delete a
+key. `None` means JSON null, which is distinct from absence. Property values must
+be JSON values with string object keys and finite numbers. Stored properties
+that are not an object cause a conflict when the update references properties;
+they are not silently converted or discarded.
+
+`preview=True` performs validation and conflict detection without changing data
+or the EPG revision. In previews, `changed` lists IDs that **would** change. No-op
+updates also leave the revision unchanged. Results classify IDs within this
+batch; they are not a durable audit log.
+
+A changed batch advances a durable database EPG revision once, in the same
+transaction as its metadata writes. If the caller wraps the operation in an
+outer `transaction.atomic()`, both the data and revision become visible on that
+outer commit, and both roll back together. Redis availability is not required
+for the update to commit. Plugins must not additionally delete XMLTV cache keys
+or call the output invalidation helper after this service.
+
+XMLTV requests select their cache generation using the committed revision.
+Previously started requests may finish with older data, while subsequent
+requests select the new generation. This protects cache and stream integrity;
+it does not provide a global database snapshot across multiple programme
+queries. Large jobs should use bounded batches and expect exports to observe
+intermediate committed batches. Built-in import and refresh paths continue to
+use their own ingestion logic with the host's revision invalidation mechanism.
+
+The cache never deletes an active build or reader's data when the revision
+changes. A cache-miss producer streams immediately; concurrent requests wait
+up to five seconds for a completed build before generating their own response.
+Only completed cache builds are replayed, with an exact byte `Content-Length`.
+Memory remains bounded to a chunk rather than a complete guide.
+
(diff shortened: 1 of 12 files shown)


dskvr:
https://github.com/sandwichfarm/covarr is presently blocked as it depends on this functionality. Alternative implementations were fragile, relied on hacks, had expensive resource consumption and resulted in uglier, less maintainable code.

SergeantPanda (changes requested):
Thanks for digging into this and for the solid writeup. The truncated-cache race is real, and the tests around invalidate-during-produce are appreciated.

I think the right core idea is already in here: put a generation/revision in the XMLTV cache key, and on invalidate advance that instead of deleting live Redis keys so in-flight readers/producers aren’t torn down. Letting the old builds expire on TTL is fine.

Where I’m hesitant is the overall size/shape. For that bug I don’t think we need the full immutable-build / Lua ownership rewrite, or changing follower behavior so concurrent requests wait and then may regenerate on their own after a few seconds. Large XMLTV exports often take longer than that, and losing live single-flight sharing feels like a real regression risk under concurrent XC/clients. Same for the plugin `update_programmes` CAS path and wiring it through the programme API: useful maybe, but I’d rather see that as a follow-up than bundled with the cache integrity fix.

My preference would be a much smaller change:

- generation/revision in the cache key via Redis (no new DB model/migration)
- `invalidate_epg_chunk_cache()` advances that and stops the SCAN/DELETE of active builds
- keep the existing chunk cache / single-flight behavior
- regression test for the invalidate-while-producing case

I’d like to land that minimal integrity fix first and not take on a schema change, the broader cache redesign, or the plugin mutation API in the same PR. Does that direction work for you?

aresthegodofwar:
Ran into this on my setup (v0.31.0). Plex pulled /output/epg/<profile> while some channel and EPG updates were running, got a malformed guide, and crashed. The crash is Plex's own bug, but the broken XML came from Dispatcharr:

parser error : Extra content at the end of the document
  <programme start="20260920210000 +0000" stop="20260921010000 +0000" channel="7

By the time I looked, the guide was fine again. Some channels also showed "Unknown Airing" in Plex until I restarted it, which I suspect is the same thing.

It still reproduces on current main. Clearing the cache mid-build leaves a broken document cached for later requests. If a second request starts a build at the same time, both builds write into the same cached output and it comes out garbled.

Issue #1652, opened 2026-09-05
[Bug]: Logo/M3U/EPG URLs bake in internal container port even with NO reverse proxy present (plain Docker port-remap)


dillardblom:
## Bug description

Absolute URLs generated by Dispatcharr (logo cache URLs, M3U/EPG endpoints, etc.) bake in the
**internal container port** instead of the externally reachable port — but **without any reverse
proxy in the path at all**. This reproduces with a plain Docker host-port remap, e.g.:

```yaml
services:
  dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:latest
    ports:
      - "80:9191"
```

I'm aware of #618 / #644 / #1226 / #704, but all of those describe misconfigured
`X-Forwarded-*` headers **behind a reverse proxy**. This report is a distinct trigger: there is
**no proxy of any kind** between the client and Dispatcharr in the affected path — a client hits
`http://dvb.example.com/` (host port 80), Docker NATs that straight to the container's port 9191,
and Dispatcharr's own process never sees `http://dvb.example.com/`, it just sees a raw connection
on port 9191. Confirmed still present on **v0.30.0** (well past the #618 fix), so this is a gap
in `get_host_and_port()`'s fallback chain (`core/utils.py`), not a regression.

## Root cause (traced in `core/utils.py::get_host_and_port`)

The fallback chain is, in order: `X-Forwarded-Host` → `Host` header → `X-Forwarded-Port` →
(if any `X-Forwarded-Proto`/`X-Forwarded-For` present, assume standard port) → **`SERVER_PORT`**
→ dev fallback. When there is no reverse proxy at all, none of the `X-Forwarded-*` headers exist,
so the function falls through to `SERVER_PORT` — which is the port the WSGI/ASGI server is bound
to **inside the container** (9191), not whatever host port Docker mapped it to. There is no way
for the process to introspect Docker's NAT mapping from inside the container, so this fallback is
structurally unable to be correct whenever host-port ≠ container-port and no proxy is present.

## Reproduction

1. `docker compose up -d` with the port mapping above (no reverse proxy in front).
2. Add channels, let a logo get cached.
3. `curl http://dvb.example.com/api/channels/channels/` (or the M3U output) → `tvg-logo`
   attribute is `http://dvb.example.com:9191/api/channels/logos/<id>/cache/`.
4. That URL is unreachable from any client outside the container (nothing listens on 9191 at the
[...]

PR #1653, opened 2026-09-05
fix(core): add public_port override for no-proxy Docker port remaps

### LINKED ISSUE #1654: [Bug]: Logo/M3U/EPG URLs bake in internal container port even with NO reverse proxy present (plain Docker port-remap)
### Describe the bug

Absolute URLs Dispatcharr generates for itself (logo cache URLs, M3U/EPG endpoints, VOD posters, catch-up) bake in the internal container port instead of the externally reachable port, but WITHOUT any reverse proxy in the path at all. This reproduces with a plain Docker host-port remap, e.g. `ports: ["80:9191"]`.

I'm aware of #618 / #644 / #1226 / #704, but all of those describe misconfigured X-Forwarded-* headers behind a reverse proxy. This is a distinct trigger: there is no proxy of any kind between the client and Dispatcharr. A client hits http://dvb.example.com/ (host port 80), Docker NATs that straight to the container's port 9191, and Dispatcharr's own process never sees the external hostname/port, it just sees a raw connection on 9191. Confirmed still present on v0.30.0, so this is a gap in get_host_and_port()'s fallback chain (core/utils.py), not a regression.

Root cause: the fallback chain checks X-Forwarded-Host -> Host header -> X-Forwarded-Port -> (if any X-Forwarded-Proto/X-Forwarded-For present, assume standard port) -> SERVER_PORT -> dev fallback. With no reverse proxy at all, none of the X-Forwarded-* headers exist, so it falls through to SERVER_PORT, the port the process is bound to inside the container (9191), not whatever host port Docker mapped it to. The process has no way to introspect Docker's NAT mapping from inside the container, so this fallback is structurally unable to be correct whenever host-port <> container-port and no proxy is present.

  1. docker compose up -d with a port mapping like ports: ["80:9191"] and NO reverse proxy in front
  2. Add channels, let a logo get cached
  3. curl http://<host>/api/channels/channels/ (or the M3U output): the tvg-logo attribute is http://<host>:9191/api/channels/logos/<id>/cache/
  4. That URL is unreachable from any client outside the container (nothing listens on 9191 at the host/network level), so broken logos for every consumer that isn't literally inside the container's network namespace

### Dispatcharr Version

 0.30.0

### Docker Version

29.7.0

    dispatcharr:                                                                                                                                                                              
      ports:                                                                                                                                                                                  
        - "80:9191"                                                                                                                                                                           
      environment:                                                                                                                                                                            
        - DISPATCHARR_ENV=aio 

### Client Information

N/A. This is a server-side absolute-URL-generation bug, not client-specific. Reproduces with any HTTP client (curl, browser) hitting the M3U/logo/EPG endpoints directly; affects every client equally since the broken URLs are baked in server-side before they ever reach a client.

dillardblom:
## Description

Absolute URLs Dispatcharr generates for itself (logo cache URLs, M3U/EPG endpoints, VOD posters, catch-up) bake in the **internal container port** instead of the externally reachable one when there is **no reverse proxy at all** in front of Dispatcharr, e.g. a plain Docker host-port remap like `ports: ["80:9191"]`, nothing else. With no `X-Forwarded-*` signals present, `_resolve_host_port_scheme()` falls through to `SERVER_PORT`, which reflects whatever port the process is bound to *inside* the container (9191), not whatever host port Docker actually mapped it to. The process has no way to introspect this from inside the container.

Adds an explicit **Public Port** setting under Settings > System (`CoreSettings.get_public_port()` / `set_public_port()`) that `_resolve_host_port_scheme()` consults right after the existing reverse-proxy header checks and right before the `SERVER_PORT` fallback. Default (empty) preserves today's behavior exactly. Also adds a `DISPATCHARR_PUBLIC_PORT` env var (`CoreConfig.ready()` -> `_seed_public_port_from_env`) that pre-fills this setting **once**, at first boot only, for Docker deployments that want to set it at compose-file-write time next to their `ports:` mapping instead of clicking through the admin UI post-deploy. Once anything has configured the setting (env var or UI), the DB value is the permanent source of truth and the env var is ignored on every later startup.

## Related Issue

Closes #1654 (re-filed through the web form after #1652 was auto-closed by the repo's bot for being CLI-submitted)

## How was it tested?

**End-to-end, on a real test instance** (own VM, Postgres+Redis, not a sandbox mock):
- Built this branch as a Docker image and ran it standalone, confirmed HTTP 200 and that a logo endpoint actually served `image/png`.
- Ran the **full Django suite**: `python manage.py test` -> 2192 tests, **1 failure**. Reproduced that exact same failure (`test_m3u_profile_test_catastrophic_regex_returns_quickly`, a ReDoS-timeout assertion in `tests/test_websocket_consumer_filter.py`) against a **freshly built, completely unmodified upstream/dev image**, confirmed pre-existing and unrelated to this change.
- Along the way, the full suite caught two real regressions this PR itself introduced (SimpleTestCase classes in `core/tests/test_core.py` and `apps/output/tests/test_views.py` that broke once `build_absolute_uri_with_port()`/`get_host_and_port()` started reading `CoreSettings.get_public_port()`, a DB query where there used to be none). Fixed by switching those three test classes to `TestCase` and clearing Django's cache in `setUp`/`tearDown` (Redis isn't covered by per-test transaction rollback).
- Ran the **full frontend suite**: `npx vitest run` -> 208 files, **6357 tests, all passing**.
- Ran `npm run lint` (ESLint + Prettier via `eslint-plugin-prettier`) -> 168 pre-existing problems repo-wide, **zero** in any file this PR touches (verified by name: my three changed frontend files don't appear in the output at all, and `SystemSettingsForm.jsx` itself only carries the same `react-hooks/exhaustive-deps` warning every sibling settings-form component already has).
- Manually verified in a Django shell against the running instance: an unconfigured Public Port falls back to the old `SERVER_PORT` behavior unchanged; setting it makes `get_host_and_port()` return that value instead; setting it equal to the scheme's standard port omits it from the URL like every other path in the function; a real reverse proxy's own `X-Forwarded-*` headers still take priority over a stale override; an **untrusted peer sending a spoofed `X-Forwarded-Host`** still gets the correct fallback instead of the spoofed value being honored.
- `DISPATCHARR_PUBLIC_PORT` seeding verified for its three cases: seeds an unset value, does **not** override an already-configured value even when the env var later disagrees, no-ops when absent.

## Checklist

- [x] I understand every change in this PR, line by line, and can explain it if asked
- [x] Backend: migrations are included if any models were changed (n/a: `CoreSettings.value` is a `JSONField`, no schema change)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (n/a: no new endpoints; `system_settings` is a free-form settings blob, not a rigidly-typed serializer)

### DIFF
--- a/apps/output/tests/test_views.py
+++ b/apps/output/tests/test_views.py
@@ -1,4 +1,4 @@
-from django.test import TestCase, Client, SimpleTestCase, RequestFactory
+from django.test import TestCase, Client, RequestFactory
 from django.http import Http404
 from django.urls import reverse
 from unittest import skipUnless
@@ -1418,11 +1418,24 @@ class XcGetEpgDummyTests(TestCase):
         self.assertIn("Unnumbered", content)
 
 
-class GenerateEpgPrevDaysTests(SimpleTestCase):
-    """Profile EPG keeps legacy prev_days=0 unless URL or user setting says otherwise."""
+class GenerateEpgPrevDaysTests(TestCase):
+    """Profile EPG keeps legacy prev_days=0 unless URL or user setting says otherwise.
+
+    TestCase (not SimpleTestCase) because generate_epg() -> build_absolute_uri_with_port()
+    now reads CoreSettings.get_public_port() on the no-forwarded-header path - a real
+    DB query where there used to be none.
+    """
 
     def setUp(self):
+        from django.core.cache import cache as django_cache
+
         self.factory = RequestFactory()
+        django_cache.clear()
+
+    def tearDown(self):
+        from django.core.cache import cache as django_cache
+
+        django_cache.clear()
 
     @patch("apps.output.epg.stream_cached_response")
     @patch("apps.output.epg.Channel.objects")
@@ -1465,11 +1478,24 @@ class GenerateEpgPrevDaysTests(SimpleTestCase):
         self.assertEqual(lan_key, same_lan_key)
 
 
-class GenerateM3UCacheKeyTests(SimpleTestCase):
-    """M3U shared cache must not reuse absolute URLs built for a different Host."""
+class GenerateM3UCacheKeyTests(TestCase):
+    """M3U shared cache must not reuse absolute URLs built for a different Host.
+
+    TestCase (not SimpleTestCase) because generate_m3u() -> build_absolute_uri_with_port()
+    now reads CoreSettings.get_public_port() on the no-forwarded-header path - a real
+    DB query where there used to be none.
+    """
 
     def setUp(self):
+        from django.core.cache import cache as django_cache
+
         self.factory = RequestFactory()
+        django_cache.clear()
+
+    def tearDown(self):
+        from django.core.cache import cache as django_cache
+
+        django_cache.clear()
 
     @patch("django.core.cache.cache")
     def test_m3u_cache_key_includes_request_origin(self, mock_cache):
--- a/core/apps.py
+++ b/core/apps.py
@@ -39,12 +39,40 @@ class CoreConfig(AppConfig):
             # Database not migrated yet: the collector keeps its defaults.
             pass
 
+        try:
+            self._seed_public_port_from_env()
+        except Exception:
+            # Database not migrated yet: seeding is retried on next startup.
+            pass
+
         # Sync developer notifications and check for version updates on startup
         # Only run in the main process (not in management commands, migrations, or workers)
         if should_skip_initialization():
             return
 
         self._sync_developer_notifications()
+
+    def _seed_public_port_from_env(self):
+        """One-time seed of Settings > System > Public Port from
+        DISPATCHARR_PUBLIC_PORT, for Docker deployments that map a
+        non-default host port (e.g. `8080:9191`) and want it set at deploy
+        time instead of clicking through the admin UI after first boot.
+
+        Only applies while the setting has never been configured. Once set -
+        by this seed or manually via the UI - the DB value is the source of
+        truth and this env var is ignored on every later startup, so a UI
+        correction always sticks even if the env var is later removed or
+        left at a stale value.
+        """
+        import os
+        from core.models import CoreSettings
+
+        env_value = os.environ.get("DISPATCHARR_PUBLIC_PORT")
+        if not env_value:
+            return
+        if CoreSettings.get_public_port() is not None:
+            return
+        CoreSettings.set_public_port(env_value)
 
     def _sync_developer_notifications(self):
         """Sync developer notifications from JSON file to database."""
--- a/core/models.py
+++ b/core/models.py
@@ -775,7 +775,25 @@ class CoreSettings(models.Model):
             "log_max_mb": 10,
             "log_keep": 5,
             "log_persist": True,
+            "public_port": None,
         })
+
+    @classmethod
+    def get_public_port(cls):
+        """Explicit override for the port baked into generated absolute URLs
+        (logos, M3U/EPG endpoints, VOD posters, catch-up, ...) when no reverse
+        proxy is present to supply X-Forwarded-* headers - e.g. a single
+        Docker host-port remap (`8080:9191`). None (default) preserves the
+        existing auto-detection behavior. See core.utils.get_host_and_port.
+        """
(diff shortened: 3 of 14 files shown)


dillardblom:
Pushed a follow-up commit: `DISPATCHARR_PUBLIC_PORT` env var that pre-fills the Public Port setting at deploy time.

Context: a pure UI setting has a gap. On a fresh instance nobody's visited Settings > System yet, so it's a manual post-deploy step even when the operator already knows their custom port mapping while writing the compose file. A bare env var alone (no UI counterpart) was the wrong call for the reason already discussed: easy to forget, and would then silently break one resource type without any indication why.

So: env var only ever seeds the setting once, while it's unset. Once anything has configured it (the env var on a previous boot, or the UI by hand), that's the source of truth on every later startup, and the env var is ignored. Set it once next to your `ports:` mapping for the common case; the UI still works standalone if you forget, change your mind, or never used Docker env vars to begin with.

Documented in `docker/docker-compose.aio.yml` next to the port mapping it corresponds to, and in the field's own description in the Settings UI.

SergeantPanda:
I'm not sure a separate public-port setting is needed. You can set DISPATCHARR_PORT to the port you want (e.g. 6393) and map it 1:1 in compose (`6393:6393`). That already makes SERVER_PORT correct for generated URLs with no reverse proxy.

Even if we did want an explicit override for asymmetric remaps (`8080:9191`), this PR is a lot more surface than that needs. A DISPATCHARR_PUBLIC_PORT env var consulted in the no-proxy fallback (before SERVER_PORT) would cover it without a DB setting, UI field, or one-time seed.

dillardblom:
Agreed, that's simpler and covers the actual case. I'll cut this down to just `DISPATCHARR_PUBLIC_PORT`, read directly in the no-proxy fallback ahead of `SERVER_PORT`, and drop the DB setting, UI field, and one-time seed logic entirely.

Will push the reduced version shortly.

SergeantPanda:
That would be more likely to be merged in but I'm still curious why not just do a direct map with docker? What is the use-case for you to not be able to do that?

Seems like you created a solution in search of a problem currently.

dillardblom:
Dug into this properly instead of arguing back, and you're right that a code change isn't the fix. Here's what I actually tested:

**Direct/no-proxy access without `DISPATCHARR_PORT` set (the bug):** `80:9191` mapping, hit the container on port 80. `get.php` bakes `:9191` into every URL (`tvg-logo`, `xmltv.php`, stream URLs). Nothing listens on 9191 externally, so all of it is dead.

**Same setup, but through a reverse proxy (https, NPMplus in front):** works correctly out of the box, no port at all in the generated URLs, matches the incoming scheme. So the proxied path was never broken, only the direct one.

**`DISPATCHARR_PORT` on a fresh container, symmetric mapping:** `-e DISPATCHARR_PORT=80 -p 80:80` and `-e DISPATCHARR_PORT=7575 -p 7575:7575`. Both self-correct completely: no port baked in on 80 (correct, default HTTP port), `:7575` baked in on the non-standard port (also correct, has to be explicit). Logos load in both cases. So this isn't limited to a couple of known ports, it works generically.

So the actual gap isn't missing functionality, it's that `DISPATCHARR_PORT`'s effect on external URL generation is undocumented for anyone running the AIO image. It's only mentioned in `docker-compose.yml` (modular), and even there it's framed purely as "must match for Celery-to-web DVR communication," nothing about it also fixing generated URLs when there's no reverse proxy. `docker-compose.aio.yml` doesn't reference it at all.

I'd like to withdraw the code change and turn this into a docs-only PR instead: add `DISPATCHARR_PORT` to `docker-compose.aio.yml` with a comment explaining when you need it (asymmetric external port, no reverse proxy in front), mirroring how `DISPATCHARR_PUBLIC_PORT` is already documented there. Happy to push that as a new commit on this branch, or close this PR and open a fresh one if you'd rather keep them separate. Let me know which you prefer.

SergeantPanda:
That makes sense to me, I'd rather a new PR to be honest.

A short explanation for DISPATCHARR_PORT would not hurt one bit. It would also be good to have it in our official docs repo as well.

PR #1657, opened 2026-09-07
fix(vod): guard against a short provider response wiping a series' episodes

jstevenscl:
## Summary

Real-world symptom this fixes: a user mid-episode, a routine playlist/VOD refresh happens, and the show goes from a full season list to showing empty in the client (TiviMate, Dispatcharr's own guide) -- with nothing having actually changed upstream. This has been reported as happening roughly daily for some users.

**Root cause chain:**

1. `apps/output/views.py::xc_get_series_info` (the handler for a client's `get_series_info` XC request) synchronously, in-request, calls `refresh_series_episodes(...)` whenever `series_relation.last_episode_refresh` is more than 24 hours old. This explains the "daily" cadence (a natural resume-after-a-day pattern) and the "mid-watching" trigger (it runs live, in the same request the client is waiting on).
2. `refresh_series_episodes` -> `batch_process_episodes` (`apps/vod/tasks.py`) then deletes every `M3UEpisodeRelation` not present in *that one fetch*, unconditionally, with no grace period: `stale_qs.exclude(stream_id__in=episode_ids).delete()`.
3. There is no status code or header that distinguishes a genuinely short/incomplete provider response (slow upstream, transient network hiccup, provider mid-update) from a real removal -- `core/xtream_codes.py`'s XC client only raises on outright HTTP/JSON/XC-error failures; a syntactically valid but truncated episode list passes through untouched.

**Fix:** `batch_process_episodes` now snapshots the existing/to-be-deleted relation counts for a series *before* any of this call's own creates/updates run, and skips the stale-relation delete for that pass when it would remove more than half of the series' on-file relations (and there are at least 5 on file to begin with, so genuinely tiny/ending series still clean up normally). Every episode actually present in the response is still created/updated regardless of this guard -- a later, complete refresh reconciles any real removal normally.

Two tunable constants: `EPISODE_STALE_DELETE_MIN_EXISTING = 5`, `EPISODE_STALE_DELETE_MAX_FRACTION = 0.5`.

## Test plan

- [x] New test file `apps/vod/tests/test_batch_process_episodes_stale_guard.py`, 15 cases: short/incomplete response vs. legitimate small removal, exact 50%/just-over-50% boundary, `MIN_EXISTING` boundary (at and just below), season growth (never guarded), proportional season removal (a real, intentional removal that's still under threshold), repeated short responses (guard doesn't wear down), reconciliation once a complete response arrives, multi-account scoping isolation, the legacy `series_relation=None` fallback path, and a documented trade-off case (a provider re-issuing brand-new stream_ids for the same episodes).
- [x] Full `apps.vod` suite (71 tests) passes.
- [x] An earlier version of the guard computed its existing-count *after* this call's own new relations were already created in the same transaction, which let a fetch's own new stream_ids skew the threshold; caught by the expanded test suite and fixed by snapshotting the decision before any writes.

https://claude.ai/code/session_01Vpd7EZap8kRiVNwXR7iy9P


SergeantPanda:
I think you are on the right track with this but I don't really want arbitrary metrics determining whether we decide it's good output or not. The problem I see with this approach is if the provider legitimately does decrease the number of episodes they have, they will never be cleaned up from the database.

I suspect the real issue is going to come from empty responses from the series endpoints (as well as movie endpoints). Recently we added a check for empty categories and skipping the refresh if they came back empty. We should look at doing that with movies and series as well (except MAYBE looking at the last_seen time to determine if this is a legitimate empty response or not) (categories didn't have a last_seen).

I am going to close this but if you want to look at handling empty responses from series/movies I'd look at it for sure. Just please follow the contributing.md file and accept the CLA so we can review it.

PR #1664, opened 2026-09-08
fix(vod): keep a blank provider name from discarding the whole import batch (#1586)

### LINKED ISSUE #1586: [Bug]: A single blank-name VOD entry discards its whole import batch (1000 movies)
### Describe the bug

If a provider returns a VOD entry whose `name` is an explicit `null`, the entire batch of up to 1000 movies is rolled back — and the refresh still reports success.

On my install, 25 blank entries out of 10,936 movies killed 8 of 11 batches: **2,866 movies imported instead of 10,936**. Same for series (2 blank entries, 2 of 4 batches lost). From the UI it just looked like my provider had no recent films.

### Root cause

`apps/vod/tasks.py`, `process_movie_batch` (~line 435):

```python
name = movie_data.get('name', 'Unknown')
```

The `'Unknown'` default only fires when the key is **absent**. `"name": null` returns `None`, which reaches `bulk_create` and violates the `NOT NULL` constraint on `vod_movie.name`. Since the batch runs in `transaction.atomic()`, one bad row discards every movie and relation in it.

`process_series_batch` (~line 812) has the same line.

The failure is silent: the `except` logs the error but returns a string instead of re-raising, so the caller still logs `Completed processing all 10936 movies in 11 chunks`.

1. Use an XC provider with at least one `"name": null` entry in `get_vod_streams`.
2. Run a VOD refresh — every batch containing such an entry fails.
3. Compare the provider's stream count against Dispatcharr's XC output.

```
ERROR Movie batch processing failed: null value in column "name" of relation "vod_movie" violates not-null constraint
ERROR Series batch processing failed: null value in column "name" of relation "vod_series" violates not-null constraint
INFO  Completed processing all 10936 movies in 11 chunks
```

Series are affected identically — `process_series_batch` has the same
`series_data.get('name', 'Unknown')` line. In my case 2 blank series
entries killed 2 of 4 series batches.

### Suggested fix

Tested locally by patching the container. All 11 movie batches and all 4
series batches then complete: 10,936 movies imported, matching the provider
exactly.

`process_movie_batch` (~line 435):

```python
stream_id = str(movie_data.get('stream_id'))
name = (movie_data.get('name') or '').strip()
if not name:
    logger.warning(f"Skipping movie with blank name (stream_id={stream_id})")
    continue
```

`process_series_batch` (~line 812):

```python
series_id = str(series_data.get('series_id'))
name = (series_data.get('name') or '').strip()
if not name:
    logger.warning(f"Skipping series with blank name (series_id={series_id})")
    continue
```

This skips blank entries. The smaller alternative is to honour the original
placeholder intent with `name = (... or '').strip() or 'Unknown'` — one line,
no `continue`. The trade-off: entries without TMDB/IMDB IDs key on
`f"name_{name}_{year or 'None'}"`, so every titleless entry becomes
`name_Unknown_None` and they all collapse into one shared record. Your call.

If skipping, the check has to sit at the top of the per-item loop —
by the time `bulk_create` runs, `relations_to_create` already references
the movie objects, so dropping one later would leave a relation pointing
at an unsaved record.

Worth considering separately: a batch that loses 1000 records shouldn't
report success.

### Dispatcharr Version

v0.29.0

### Docker Version

Docker version 28.1.1, build 4eba377

```yaml
services:
  gluetun:
    image: qmcgaw/gluetun
    container_name: gluetun
    restart: unless-stopped
    cap_add:
      - NET_ADMIN
    devices:
      - /dev/net/tun:/dev/net/tun
    environment:
      VPN_SERVICE_PROVIDER: <provider>
      VPN_TYPE: wireguard
      WIREGUARD_PRIVATE_KEY: <redacted>
      WIREGUARD_ADDRESSES: <redacted>
      SERVER_COUNTRIES: <country>
      FIREWALL_OUTBOUND_SUBNETS: "172.16.0.0/12,192.168.0.0/16"
      TZ: Europe/Vienna
    ports:
      - "9191:9191"          # Dispatcharr UI + M3U/Xtream/HDHR output

  dispatcharr:
    container_name: dispatcharr
    restart: unless-stopped
    shm_size: "512mb"
    network_mode: "service:gluetun"
    depends_on:
      gluetun:
        condition: service_healthy
    environment:
      DISPATCHARR_ENV: aio
      REDIS_HOST: localhost
      CELERY_BROKER_URL: redis://localhost:6379/0
      DISPATCHARR_LOG_LEVEL: info
      TZ: Europe/Vienna
    volumes:
      - dispatcharr_data:/data

volumes:
  dispatcharr_data:
```

### Client Information

No IPTV client involved. Reproduced entirely server-side — the failure
happens during VOD import, before any client is in the picture.

Verified with curl against the XC API:
  Dispatcharr get_vod_streams  →  2,866 movies
  provider get_vod_streams     →  10,936 movies

Dispatcharr web UI accessed via Chrome 151.0.7922.169.

thejdubb02:
## Description

A VOD entry whose name is an explicit null (or blank/whitespace) discarded its entire import batch. `process_movie_batch` and `process_series_batch` build each batch inside one `transaction.atomic()` block, so when a null name reached `Movie.name` / `Series.name` (both NOT NULL), `bulk_create` raised `IntegrityError` and rolled back every other row in the batch (up to 1000). The outer `except` logged the error but returned a string instead of re-raising, so the refresh still reported success and the missing titles looked like a provider gap.

Root cause: `name = movie_data.get('name', 'Unknown')` only falls back to `'Unknown'` when the key is absent. An explicit `"name": null` returns `None`, which flows into `bulk_create`.

Fix: extend the existing `'Unknown'` fallback to cover null, empty, and whitespace-only names (and stringify non-string values so a numeric name does not break `.strip()`):

```python
name = str(movie_data.get('name') or '').strip() or 'Unknown'
```

Applied to both `process_movie_batch` and `process_series_batch`. This is the smallest change that keeps the batch intact: it does not alter behavior for rows that already have a usable name, and absent names keep importing as `'Unknown'` exactly as before.

Scope: this targets the reported `name` cause. Making a batch resilient to any other bad column (catch `IntegrityError` and retry row by row) would be a larger change and is left out to keep this focused.

## Related Issue

Closes #1586

## How was it tested?

Ran the `apps/vod/tests/test_vod_sync_preserve_details` suite (16 tests) against a real Postgres database (the project's `settings_test` config), all passing. Added `VODBlankNameBatchTests` covering null, whitespace-only, absent, and non-string names for both movies and series. Each test asserts the valid row in the same batch still imports (the regression this fixes) and that blank names become `'Unknown'`, never NULL. Both new tests fail without the fix with the reported not-null `IntegrityError`.

## Checklist

- [x] Backend: migrations are included if any models were changed (no models changed, so none needed)
- [x] Frontend: ESLint and Prettier pass cleanly (no frontend changes in this PR)

### DIFF
--- a/apps/vod/tasks.py
+++ b/apps/vod/tasks.py
@@ -433,7 +433,11 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
     for movie_data in batch:
         try:
             stream_id = str(movie_data.get('stream_id'))
-            name = movie_data.get('name', 'Unknown')
+            # A null or blank name reaches Movie.name (NOT NULL) and, because the
+            # batch is created in one transaction.atomic() block, one such row rolls
+            # back the whole batch. Extend the existing 'Unknown' default (which only
+            # fired for a missing key) to cover null/blank names too. (#1586)
+            name = str(movie_data.get('name') or '').strip() or 'Unknown'
 
             # Get category with proper error handling
             category = None
@@ -806,7 +810,10 @@ def process_series_batch(account, batch, categories, relations, scan_start_time=
     for series_data in batch:
         try:
             series_id = str(series_data.get('series_id'))
-            name = series_data.get('name', 'Unknown')
+            # See process_movie_batch: coerce a null/blank name to the existing
+            # 'Unknown' default so one bad row cannot roll back the whole atomic
+            # batch via the NOT NULL constraint on Series.name. (#1586)
+            name = str(series_data.get('name') or '').strip() or 'Unknown'
 
             # Get category with proper error handling
             category = None
--- a/apps/vod/tasks.py
+++ b/apps/vod/tasks.py
@@ -433,11 +433,16 @@ def process_movie_batch(account, batch, categories, relations, scan_start_time=N
     for movie_data in batch:
         try:
             stream_id = str(movie_data.get('stream_id'))
-            # A null or blank name reaches Movie.name (NOT NULL) and, because the
-            # batch is created in one transaction.atomic() block, one such row rolls
-            # back the whole batch. Extend the existing 'Unknown' default (which only
-            # fired for a missing key) to cover null/blank names too. (#1586)
-            name = str(movie_data.get('name') or '').strip() or 'Unknown'
+            # Skip blank names: Movie.name is NOT NULL, and one null in this
+            # atomic batch would roll back every other row.
+            name = str(movie_data.get('name') or '').strip()
+            if not name:
+                logger.warning(
+                    "Skipping movie with blank name (stream_id=%s, account=%s)",
+                    stream_id,
+                    account.id,
+                )
+                continue
 
             # Get category with proper error handling
             category = None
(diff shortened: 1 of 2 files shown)


SergeantPanda (approved):
Thank you for the PR. 

I made a slight change to it, your fix was OK but then we would have a bunch of duplicate 'unknown' movies/series in the database with little value and only increasing the chance of bad merges of data. 

So, I opted instead to just skip any movie or series whose name is empty.

PR #1665, opened 2026-09-08
fix(logos): serve the logo cache to image/* clients instead of 406 (#1541)

### LINKED ISSUE #1541: [Bug]: Logo cache endpoint returns 406 for Accept: image/* - channel logos broken in Xtream API clients (Tivimax/Apple TV)
### Describe the bug

The channel logo cache endpoint (`/api/channels/logos/<id>/cache/`) returns **406 Not Acceptable** for any request whose `Accept` header asks for images - e.g. `Accept: image/*` - even though the endpoint exists to serve images.

Native image-loading libraries send exactly such headers. Tivimax on Apple TV (Alamofire) is how I hit it: channels and streams work fine over the Xtream API, but every logo request fails, so no channel icons render. Browsers and curl send `Accept: */*` and get 200, which makes the endpoint look healthy when tested by hand - my access log had 871 consecutive 406s from the Apple TV before I spotted the header difference.

### Root cause
The `cache` action (`apps/channels/api_views.py`, `def cache`, ~line 2861) is a DRF viewset action, so DRF content negotiation runs before the action body. With only JSON renderers registered, any image-specific `Accept` header fails negotiation and DRF returns 406 before `serve_local_or_remote_image()` is ever called. The response content type is decided inside the action, not by a renderer, so negotiation adds nothing here. The VOD logo cache endpoint (`/api/vod/vodlogos/<id>/cache/`) has the same issue.

### Suggested fix
Bypass content negotiation on the image-serving actions, e.g.:

```python
from rest_framework.negotiation import BaseContentNegotiation

class IgnoreClientContentNegotiation(BaseContentNegotiation):
    def select_parser(self, request, parsers):
        return parsers[0]
    def select_renderer(self, request, renderers, format_suffix):
        return (renderers[0], renderers[0].media_type)
```

applied via `content_negotiation_class` on the logo/VOD-logo cache actions (or serve these routes from a plain Django view).

### Workaround
Overriding the header in the bundled nginx config fixes all clients - added `proxy_set_header Accept "*/*";` to the logo cache `location` blocks in `/etc/nginx/sites-enabled/default` (bind-mounted over the container's copy).

### Environment
- Docker AIO on Ubuntu, direct LAN access (no external reverse proxy)

```console
$ curl -s -o /dev/null -w '%{http_code}\n' -H 'Accept: */*' http://SERVER:9191/api/channels/logos/1/cache/
200
$ curl -s -o /dev/null -w '%{http_code}\n' -H 'Accept: image/*' http://SERVER:9191/api/channels/logos/1/cache/
406
$ curl -s -H 'Accept: image/*' http://SERVER:9191/api/channels/logos/1/cache/
{"detail":"Could not satisfy the request Accept header."}
```

Real client example from nginx access.log:
```
192.168.1.x - - [13/Aug/2026:19:44:48 +0200] "GET /api/channels/logos/3/cache/ HTTP/1.1" 406 57 "-" "TiviMaxTV/3.6.8 (com.tivimax.premium; build:60807.1465; tvOS 26.6.0) Alamofire/5.9.1"
```

### Dispatcharr Version

0.29.0 (AIO Docker image, `ghcr.io/dispatcharr/dispatcharr:latest`)

### Docker Version

Docker version 29.7.2, build a7dcaa6

  dispatcharr:
    container_name: dispatcharr
    environment:
      - TZ=Europe/Zagreb
      - DISPATCHARR_ENV=aio
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=info
    volumes:
      - /media/appdata/dispatcharr/data:/data
    ports:
      - 9191:9191
    restart: unless-stopped

### Client Information

Apple TV 4K, Tivimax Premium 3.6.8, tvOS 26.6.0

thejdubb02:
## Description

`GET /api/channels/logos//cache/` (and `/api/vod/vodlogos//cache/`) returns `406 Not Acceptable` for a request whose `Accept` header asks for an image, for example `Accept: image/*`. Native image loaders send exactly that (the reporter hit it with Tivimax on Apple TV via Alamofire), so channel and VOD logos never load for those clients. Browsers and curl send `Accept: */*` and get 200, which is why the endpoint looks healthy when tested by hand.

Root cause: the `cache` action streams the image itself (`serve_local_or_remote_image` / `serve_vod_image` build a `StreamingHttpResponse`/`HttpResponse` and set the content type), but DRF runs content negotiation before the action body. With only JSON renderers registered, an image-specific `Accept` fails negotiation and DRF returns 406 before the action ever runs.

Fix: a small `RawImageContentNegotiationMixin` (in `core/image_proxy.py`, the shared logo/art module) makes the viewset skip the `Accept` check for the `cache` action only. `LogoViewSet` and `VODLogoViewSet` mix it in; every other action keeps normal JSON negotiation. This is scoped to the one action that returns a raw response, so it does not change how the JSON API endpoints behave.

Chose overriding `get_content_negotiator` (via the mixin) rather than a custom renderer because the action returns a raw response and never invokes a renderer, so there is nothing for a renderer to do; the only thing to fix is the pre-action negotiation.

## Related Issue

Closes #1541

## How was it tested?

Ran the full logo test suites (13 tests) against a real Postgres database (the project's `settings_test` config), all passing. Added tests that:
- assert the channel and VOD cache endpoints return 200 with an `image/...` content type for `Accept: image/*`, and fail without the fix with 406.
- assert a non-cache action (`retrieve`) still returns 406 for `Accept: image/*`, so the negotiation skip stays scoped to the `cache` action.

## Checklist

- [x] Backend: migrations are included if any models were changed (no models changed, so none needed)
- [x] Backend: new API endpoints appear correctly in the OpenAPI schema (no new endpoints, negotiation change on existing routes)
- [x] Frontend: ESLint and Prettier pass cleanly (no frontend changes in this PR)


SergeantPanda:
Thank you for this. It looked pretty good overall.

Made a couple of changes to reuse the same idea for the vod image endpoints. I cleaned up some of the overly verbose comments.

PR #1666, opened 2026-09-08
fix(vod): create a relation for every category/stream a movie or series appears under in a batch (#1511)

### LINKED ISSUE #1511: [Bug]: Multiple versions of movies/series disappear when they happen to be in the same batch
### Describe the bug

When the same movie/series exists in multiple enabled XC Stream VOD categories, the provider might return it multiple times (with different `category_id`s or `stream_id`s). If these duplicate entries happen to fall into the same sync batch of 1000 items (`chunk_size`), Dispatcharr outputs only one of them.

This happens because of the deduplication condition in `apps/vod/tasks.py` inside the `process_movie_batch` function (around line 485):

```
# Skip duplicates in this batch
if movie_key in movie_keys:
    continue
```

Since the key (movie_key or series_key) is based on the metadata (TMDB/IMDB/Name+Year) rather than the actual stream_id, the code completely skips creating the relation (M3UMovieRelation / M3USeriesRelation) for any subsequent appearances of that item in the same batch.

Interestingly, this bug is chunk-dependent: if the duplicate stream happens to fall into a different batch of 1000, it processes correctly because the keys dictionary is reset for each chunk.

(Note: The exact same logic flaw also exists in process_series_batch around line 850).

```
# Skip duplicates in this batch
if series_key in series_keys:
    continue
```

1. Add an Xtream Codes account.

2. Enable two or more VOD categories containing the exact same movies.

3. Wait for the VOD synchronization to complete.

4. Check the XC Stream API output or use any IPTV player. You will notice that overlapping movies are missing from some categories.

### Dispatcharr Version

v0.28.2

### Docker Version

29.7.1

```
services:
  dispatcharr:
    image: ghcr.io/dispatcharr/dispatcharr:base
    container_name: dispatcharr_dev
    restart: unless-stopped
    ports:
      - "5656:5656"
      - "8001:8001"
      - "9191:9191"
    environment:
      - DISPATCHARR_ENV=dev
      - REDIS_HOST=localhost
      - DISPATCHARR_LOG_LEVEL=debug
      - TZ=Europe/Istanbul
    volumes:
      - ./dispatcharr-dev:/app
      - ./dispatcharr-data:/data
      - ./dispatcharr-vods:/VODS
```

### Client Information

Devices: iPhone, Mac
App: IPTV+ app, Python

northernpowerhouse:
## Description

When a provider offers the same movie or series under more than one enabled XC VOD category (or with more than one stream_id/series_id) and both occurrences land in the same sync batch (chunk_size = 1000), only the first one got an M3UMovieRelation/M3USeriesRelation. The rest were silently dropped from the output because the movie/series itself was matched correctly by TMDB/IMDB/name+year, but every category/stream after the first was lost.

Root cause: process_movie_batch/process_series_batch build a movie_keys/series_keys dict keyed by canonical identity (movie_key/series_key), and used if movie_key in movie_keys: continue to skip recomputing props for anything already seen but that same skip also discarded the relation for that occurrence, because relations were built from the single stored stream_id/category per key. Since M3UMovieRelation and M3USeriesRelation are uniquely keyed on (m3u_account, stream_id) / (m3u_account, external_series_id), not on the movie/series identity, collapsing them to "one per canonical movie" was the wrong axis.

Fix: keep the identity dedup (movie/series props are still only computed once per canonical key, so no duplicate Movie/Series rows), but now record every provider occurrence (stream_id/series_id + its category) under that key in an occurrences dict, and build one relation per occurrence. A literal duplicate row (same stream_id appearing twice) is still coalesced to a single relation via dict.setdefault, so that case behaves exactly as before.

## Related Issue

Closes #1511

## How was it tested?

Added VODDuplicateAcrossCategoriesBatchTests to apps/vod/tests/test_vod_sync_preserve_details.py:

a movie in two enabled categories in one batch gets a relation for each category, both pointing at the same canonical Movie
a literal duplicate row (same stream_id twice) still produces exactly one relation
same two tests mirrored for series/M3USeriesRelation
Verified the new tests fail against unmerged dev (1 != 2 relations) and pass with the fix. Ran the full apps/vod/tests/test_vod_sync_preserve_details.py suite (17 tests) twice: once against a throwaway PostgreSQL 16 container, and again inside a real deployed build of this branch (see below). both clean, no regressions in existing coverage (blank-name handling, is_adult sync, detail-preservation on list sync).

Checked compatibility with 
Dispatcharr/Dispatcharr#1664 (separate fix for the null-name batch-rollback bug, touching the same two functions): merges without conflict in tasks.py, and a combined regression scenario (null name + cross-category duplicate in the same batch) passes with both fixes applied together.

Live test against real production data. Built this branch into a full Dispatcharr image and ran it as a drop-in replacement for my own production instance (same /data, network, and proxy config), then triggered refresh_vod_content against two real Xtream Codes accounts (~46k series, tens of thousands of movies):

Movie relations went from 8,235 to 9,041 (+806 relations recovered on the first sync) on one account; movies with more than one relation went from 474 to 1,050.
Concrete example: a movie previously stuck with a single relation now correctly carries 4: standard, Dolby Audio, 4K HDR Dolby Vision, and 4K 3840p Dolby Audio, instead of losing 3 of them to the batch dedup bug.
Re-ran the refresh: counts stayed exactly stable (idempotent, no duplicate-relation growth).
Ran a second account against the same underlying catalog: converged to the identical final state, as expected.
Zero errors, zero batch failures, across the full run.

## Checklist


SergeantPanda (approved):
Thank you for the PR.

This looked good and solves the issue correctly. I did make an extra change here to save the extra props that may exist in the duplicates but not in the first occurance. 

I also cleaned up some of the verbose commments a bit (prefer just non obvious code comments)

northernpowerhouse:
Noted about the comments, cheers!

Issue #1644, opened 2026-09-04
[Feature]: Optional web UI sign-in from a trusted reverse proxy identity header


benjitobz:
### Is your feature request related to a problem?

Dispatcharr running behind an authenticating reverse proxy (Cloudflare Access, Authelia, oauth2-proxy, authentik) makes users log in twice. The proxy authenticates them, and then Dispatcharr presents its own login form to a user whose identity the proxy has already established and passed along.

Session expiry makes it worse. Access tokens last 30 minutes and refresh tokens a day, so a tab left open lands back on the login screen even though the proxy session is still perfectly valid.

To be clear about scope: this is about the **web UI login only**. It does not ask to change how Xtream Codes or M3U/EPG client endpoints authenticate.

### Describe the solution you'd like

An optional setting — off by default — where an admin names an identity header that a trusted reverse proxy sets, and Dispatcharr signs in the matching user instead of showing the login form.

Behaviour:

- Two fields: an enable toggle and the header name (for example `Remote-User`, `X-Forwarded-User`, or `Cf-Access-Authenticated-User-Email`).
- **The header is only honored when the connecting peer is a trusted proxy**, reusing the `DISPATCHARR_TRUSTED_PROXIES` mechanism that already exists from #1410. This is the part that answers the "implicit trust" objection: the header is never taken at face value from an arbitrary client, only from a peer the admin has already designated as their proxy. A spoofed header from anywhere else is ignored and logged.
- The identity resolves to a username first, then to an email matching exactly one account, so proxies that assert an email work without renaming accounts. An ambiguous email matches nobody.
- The header name is validated to letters, digits and dashes, since an underscore would be ambiguous once Django normalises it.
- The UI probes for proxy sign-in on load and falls back silently to the password form when it is unavailable, so nothing changes for anyone who has not enabled it. It also retries after a refresh token expires, which is what otherwise strands the SPA mid-session.

### Additional context

**Relationship to #1553 (Native OIDC SSO):** that request targets the Xtream API path for IPTV clients and wants full JWT verification against an IdP's JWKS. This one is narrower — web UI only — and complementary rather than competing; it would not preclude or complicate native OIDC later.

[...]

PR #1667, opened 2026-09-09
feat(auth): optional reverse proxy header authentication

benjitobz:
## Description

Adds an opt-in "Reverse Proxy Auth" setting under Settings -> System that signs a user in from an identity header asserted by an authenticating reverse proxy, so instances already sitting behind Cloudflare Access, oauth2-proxy, Authelia or similar do not present a second login form.

The header is only honored when an admin has enabled the setting AND the connecting peer is a trusted proxy (DISPATCHARR_TRUSTED_PROXIES, which already defaults to private and loopback ranges), because the header is otherwise client-suppliable. Identities resolve to a username first, then to an email matching exactly one account, so proxies that assert an email work without renaming accounts.

POST /api/accounts/auth/proxy-login/ exchanges the header for the same JWT pair as password login and returns 401 whenever proxy auth is unavailable, so the web UI probes it on load and silently falls back to the password form. initializeAuth retries it when a refresh token has expired, which otherwise leaves the SPA stranded until a manual reload. An explicit logout opts out until the page is reloaded so the logout button still does something visible.

## Related Issue
Closes https://github.com/Dispatcharr/Dispatcharr/issues/1644

### Backend — `apps/accounts/tests/test_reverse_proxy_auth.py` (11 new tests)

| Test | Asserts |
| --- | --- |
| `test_disabled_by_default_even_with_the_header_present` | 401 — the feature is inert until an admin enables it |
| `test_enabled_trusted_peer_returns_tokens` | 200 with `access` + `refresh` |
| `test_header_from_untrusted_peer_is_ignored` | 401 when `REMOTE_ADDR` is outside the trusted set |
| `test_explicitly_trusted_peer_is_honored` | 200 once that peer is added to `DISPATCHARR_TRUSTED_PROXIES` |
| `test_missing_header_does_not_sign_anyone_in` | 401 when the header is absent |
| `test_unknown_identity_is_rejected` | 401 for an identity with no account |
| `test_inactive_user_is_rejected` | 401 for `is_active=False` |
| `test_identity_matches_email_for_proxies_that_assert_one` | case-insensitive email match |
| `test_email_shared_by_two_accounts_matches_nobody` | 401 — an ambiguous email never resolves |
| `test_only_the_configured_header_is_read` | a different header is ignored; the configured one works |
| `test_returned_tokens_authenticate_api_calls` | the issued access token authenticates `GET /api/accounts/users/me/` |

### Frontend — 18 new tests

`store/__tests__/auth.test.jsx` (6): tokens stored for a proxy-asserted identity · failure reported when the proxy asserts nothing · no sign-in after an explicit logout · sign-in again once the page reloads · still signs in after a session-expiry logout · attempted by `initializeAuth` when no refresh token exists.

`components/forms/settings/__tests__/ReverseProxyAuthForm.test.jsx` (12):

- **Rendering:** enable switch and header input present; the trusted-proxy warning is shown; the form seeds from the stored setting; falls back to disabled/empty when the setting row is missing.
- **Validation:** a header is required when enabling; empty is allowed while disabled; a conventional name is accepted; a name containing an underscore is rejected.
- **Saving:** updates an existing row with a trimmed header; creates the row when none exists; refreshes settings and shows the success banner; surfaces the backend's error message on failure.

### Manually, on a live instance behind an authenticating proxy
Enabled the setting and confirmed the browser lands straight in the UI with no login form; a request without the header returns 401; a spoofed header from an untrusted peer is rejected and logs `Ignoring <header> from untrusted peer`; and after narrowing `DISPATCHARR_TRUSTED_PROXIES` to the single proxy address, sign-in still works while a spoof from any other container is refused.

## UI Changes
image
image

## Checklist

### DIFF
--- a/apps/accounts/api_views.py
+++ b/apps/accounts/api_views.py
@@ -16,11 +16,18 @@ from dispatcharr.utils import (
     SETUP_ALLOWED_IP_ENV,
     get_client_ip,
     network_access_allowed,
+    proxy_auth_identity,
     setup_ip_allowed,
 )
 
 from .models import User
-from .serializers import UserSerializer, GroupSerializer, PermissionSerializer
+from .serializers import (
+    UserSerializer,
+    GroupSerializer,
+    PermissionSerializer,
+    ProxyLoginResponseSerializer,
+)
+from rest_framework_simplejwt.tokens import RefreshToken
 from rest_framework_simplejwt.views import TokenObtainPairView, TokenRefreshView
 
 logger = logging.getLogger(__name__)
@@ -51,6 +58,20 @@ def _setup_forbidden_response(client_ip):
         },
         status=403,
     )
+
+
+def _resolve_proxy_auth_user(identity):
+    """Match a proxy-asserted identity by username, then by unambiguous email."""
+    user = User.objects.filter(username__iexact=identity).first()
+    if user is not None:
+        return user
+
+    if "@" in identity:
+        matches = list(User.objects.filter(email__iexact=identity)[:2])
+        if len(matches) == 1:
+            return matches[0]
+
+    return None
 
 
 class TokenObtainPairView(TokenObtainPairView):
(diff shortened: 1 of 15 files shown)


dispatcharr/utils.py
```diff
@@ -145,6 +140,44 @@ def get_client_ip(request):
...
+    """
+    config = CoreSettings.get_reverse_proxy_auth_settings()
+    if not config.get("enabled"):
+        return None
+
+    header = (config.get("header") or "").strip()
+    if not validate_proxy_auth_header(header):
+        return None
+
+    peer = _normalize_ip(request.META.get("REMOTE_ADDR") or "")
```
sethwv: This turns the generic private-network forwarded-header default into a JWT issuer trust boundary. With `DISPATCHARR_TRUSTED_PROXIES` unset, any LAN host or container that can reach Dispatcharr directly can send the configured header and impersonate an account. Please require an explicitly configured, dedicated proxy-auth allowlist rather than trusting every private address by default.

apps/accounts/api_views.py
```diff
@@ -53,6 +60,20 @@ def _setup_forbidden_response(client_ip):
     )
 
 
+def _resolve_proxy_auth_user(identity):
+    """Match a proxy-asserted identity by username, then by unambiguous email."""
+    user = User.objects.filter(username__iexact=identity).first()
```
sethwv: `username__iexact(...).first()` can select either account when case-distinct usernames exist. Please use an exact username match or reject ambiguous matches, as the email lookup does.

apps/accounts/api_views.py
```diff
@@ -215,6 +236,82 @@ def login(self, request):
...
+        description=(
+            "Exchange a reverse-proxy-asserted identity for JWT tokens. "
+            "Returns 401 unless reverse proxy auth is enabled, the request "
+            "carries the configured header, and the connecting peer is a "
+            "trusted proxy. Takes no request body."
+        ),
+        request=None,
+        responses={200: ProxyLoginResponseSerializer},
+    )
+    def proxy_login(self, request):
```
sethwv: Password and Django admin sign-in already share the `login` throttle. This endpoint also mints tokens, updates `last_login`, and writes audit events, but bypasses that limit. Please apply `LoginRateThrottle` here as well.

sethwv (changes requested):
This PR & feature seem to be in a really decent spot, serves as a good addition especially while we're still discussing oidc/auth internally, and it's appropriately tucked away and marked to be for advanced users only which is definitely the direction the team is thinking for something like this, given some of the risks involved with improper config.

Just a couple of changes I'd like to see to make sure we're a little more locked down. Please take a look at my inline notes & the merge conflict on `core/serializers.py`. Once those are addressed I'll give this another pass.

Thanks for submitting!

benjitobz:
Applied the recommended changes lmk if there's anything else

PR #1677, opened 2026-09-11
feat(dvr): stamp capture playlists with EXT-X-PROGRAM-DATE-TIME

v8eta:
## Description

Adds `program_date_time` to the DVR capture's `-hls_flags`, so every segment in the capture's working `index.m3u8` carries `EXT-X-PROGRAM-DATE-TIME`.

**Why.** The capture and the final playlist remux both run `-avoid_negative_ts make_zero`, so a finished recording starts at 0 whatever the source's timestamps were, and a resumed attempt restarts at 0. The recording therefore carries no broadcast timing at all. With this flag the playlist records when media time 0 of each attempt happened (from the muxer's start clock), which is what a post-processor needs to align a recording with the EPG, account for an outage, or place a break on the broadcast clock. The recording file itself is unchanged; only the working playlist gains tags.

**How.** The flag string moves into a module constant next to the command builder so the test can pin it. A restart-marking flag (`discont_start`) was tried and dropped: with `append_list` the HLS muxer already writes `EXT-X-DISCONTINUITY` ahead of an appended attempt, and the playlist remux produced the same monotonic output with and without the extra flag.

## Related Issue

Small, self-contained change to one ffmpeg argument; no issue opened.

## How was it tested?

- `python manage.py test apps.channels.tests.test_dvr_hls_flags apps.channels.tests.test_dvr_output_profile` in a container of the current `dev` image: 18 tests, OK. The new tests pin the flag string and its presence in the built command for a first and a resumed attempt; they do not establish that the muxer honours the flag.
- Two-attempt synthetic capture (3 s + 3 s appended into one playlist, then the playlist remux command from `_dvr_build_hls_playlist_remux_cmd`) on ffmpeg 6.1 and 8.1: every segment carries `EXT-X-PROGRAM-DATE-TIME`, the splice carries the `EXT-X-DISCONTINUITY` that `append_list` writes, and the remuxed output has no backwards video timestamps, identical with and without `discont_start`.

## Checklist


SergeantPanda:
I don't see any value in this. Do you have any evidence that this actually helps in any way?

From what I can tell this will write to the manifest while it's being generated but nothing will actually utilize it during remux. 

FFmpeg during remux will already see the discontinuities and the output mkv will work as intended.

PR #1679, opened 2026-09-11
fix(dvr): enrich recording_end with outcome/file details instead of new events

### LINKED ISSUE #1681: [Feature]: recording_finalized / recording_failed events after DVR post-processing

`recording_end` is emitted when the capture's ffmpeg stops, before the HLS playlist is remuxed into the final file and before the final status (`completed` / `interrupted` / `stopped`) is written. A Connections webhook or a plugin that subscribes to it and tries to act on the recording (rename, index, post-process, notify on failure) finds no finished file yet and has to poll `custom_properties` until `remux_success` appears.

Two new events, dispatched after the final metadata save at the end of `run_recording`:

- `recording_finalized`: always, with `status`, `file_path`, `file_url`, `file_size`, `remux_success`, `has_file`, `bytes_written`, `interrupted_reason` (plus the usual `channel_name` / `recording_id`).
- `recording_failed`: alongside it when there is no playable file (remux failed, or the file is missing or empty), so a failure alert needs no payload template.

Both registered in `SUPPORTED_EVENTS`, `SystemEvent.EVENT_TYPES` (with migrations) and the frontend labels. `recording_end` stays as it is.

Alternatives considered: moving `recording_end` later would change its meaning for existing subscribers; polling `custom_properties` works but every consumer reinvents it.

### Additional context

A branch is ready: #1679. Re-opened from #1678 and #1680, which were closed for not being created from the web form.

v8eta:
## Description

Enriches the existing `recording_end` event instead of adding new terminal event types.

**What.** `recording_end` now fires after the HLS→MKV remux and final metadata save (previously it fired at ffmpeg stop, before either) and carries `outcome` (`success`/`failed`), `has_file`, `failure_reason` (`remux_failed`/`missing_file`/`empty_file`/`null`), `status`, `interrupted_reason`, `file_path`, `file_name`, `file_url`, `file_size`, `remux_success`, `bytes_written`, `start_time` and `end_time`. It's also now emitted from the crash-recovery finalize path in `recover_recordings_on_startup`, so a remux completed during startup recovery isn't silent.

**Why.** `recording_end` previously fired when the capture's ffmpeg stopped, before the playlist remux and before the final status was written, so a Connections webhook or plugin acting on it found no finished file and had to poll `custom_properties` for `remux_success`. Per review, this reshapes the original PR (which added `recording_finalized`/`recording_failed`) into a single enriched terminal event instead.

**How.** `_dvr_recording_end_payload(cp, final_path, remux_success, start_time=None, end_time=None)` replaces the old `_dvr_finalized_events` two-event builder with a single pure function that builds the enriched payload from the saved `custom_properties` and the on-disk file. It tolerates a missing or malformed `custom_properties` and never raises. `failure_reason` priority: `remux_failed` (remux itself failed) > `missing_file` (file absent/unreadable despite a claimed-success remux) > `empty_file` (zero bytes) > `None` (success). The early, pre-remux `recording_end` emit is removed; the one at the end of `run_recording` and the new one in `recover_recordings_on_startup` both call the same helper. `recording_finalized`/`recording_failed` (models, migrations, frontend labels/icons) are reverted.

## Related Issue

Closes #1681

## How was it tested?

- `DJANGO_SECRET_KEY=x TEST_USE_SQLITE=1 uv run python manage.py test apps.channels.tests.test_dvr_finalized_events`: 11 tests, OK. Covers success, interrupted-with-a-file (still success), failed remux (even with a real file on disk), missing file (even with `remux_success=True`), empty file, malformed inputs, JSON serialisability, `file_name` fallback to `os.path.basename`, and `start_time`/`end_time` passthrough/defaulting.
- Stepped through all four `failure_reason` branches under a live debugger (mcp-debugger) against a standalone copy of the payload builder, confirming the priority order (remux failure beats file presence) and the exact `file_size`/`path` values at each branch.
- Repo-wide search confirms no remaining references to `_dvr_finalized_events`, `recording_finalized`, or `recording_failed`.
- They do not establish that `run_recording` or `recover_recordings_on_startup` reach the emit site on every exit path in a real deployment, or that a subscriber receives the delivery; that needs a real recording end to end.

## Checklist

- [x] Backend: migrations are included if any models were changed (none needed here — the two new-choice migrations from the previous shape are reverted, not replaced)


SergeantPanda (changes requested):
Thanks for this. The core problem is real: `recording_end` currently fires when capture stops, before remux and final metadata, so a webhook or plugin can’t reliably act on a finished file. The helper and tests look solid.

I’d like to take a different shape though, instead of adding `recording_finalized` / `recording_failed`.

**Reuse / fix `recording_end`**

For Connect users, “recording ended” already sounds like “the recording is done.” Firing it before the file is ready feels more like a timing gap than intentional design. DVR and events aren’t widely adopted yet, so we’re fine calling this out in the changelog as a behavior change rather than adding more terminal events.

What we’re asking for:

1. Move the existing `recording_end` emit to after remux + final metadata save (where the new events are now).
2. Remove the early `recording_end` at ffmpeg/capture stop.
3. Drop `recording_finalized` / `recording_failed` (models, migrations for those two, frontend labels/colours, related tests). Keep any migration extras you need for the other unmerged PR.
4. Enrich the `recording_end` payload so scripts can branch on success/failure without a second event, something like:
   - `outcome`: `"success"` | `"failed"` (string preferred so Connect’s falsy-key stripping doesn’t drop it)
   - `has_file`, `remux_success`, `failure_reason` (`remux_failed` / `missing_file` / `empty_file` / null)
   - `file_path`, `file_url`, `file_name`, `file_size`
   - `bytes_written`, `status`, `interrupted_reason`
   - nice-to-haves if easy: `title` / season / episode, scheduled `start_time` / `end_time`, `started_at` / `ended_at`
5. Also emit from the expired-recording path in `recover_recordings_on_startup`, using the same helper, so crash-recovery remuxes aren’t silent.
6. Keep the pure helper that never raises. Just have it build the `recording_end` payload and decide `outcome` / `has_file` the same way you do now.

Changelog should note that `recording_end` now fires after post-processing with file/outcome details, and may fire later than before (after remux).

Happy to tweak field names if you want. Main point is one terminal event at the right time, with enough payload, not new event names. Thanks for digging into this.

v8eta:
Reshaped per your review — pushed 89676fe9.

- Dropped `recording_finalized`/`recording_failed` entirely (models, migrations, frontend labels/icons).
- Moved the existing `recording_end` emit from ffmpeg-stop to after the remux + final metadata save; removed the early emit.
- Enriched `recording_end`'s payload with `outcome` (`success`/`failed`), `has_file`, `failure_reason` (`remux_failed`/`missing_file`/`empty_file`/`null`), `status`, `interrupted_reason`, `file_path`, `file_name`, `file_url`, `file_size`, `remux_success`, `bytes_written`, `start_time`, `end_time`. (Skipped title/season/episode from the nice-to-haves — nothing in `run_recording` has those without an extra query, so I left them out rather than bolt one on.)
- Also now emitted from the expired-recording finalize path in `recover_recordings_on_startup`, via the same payload builder.
- Updated the changelog entry to describe it as a behavior change to `recording_end` rather than new events.

PR title/description updated to match. Let me know if you'd want different field names or the title/season/episode fields added after all.

SergeantPanda (changes requested):
Thanks for reshaping this around `recording_end`. The helper and the failure-reason order look right. A few call-site gaps should be fixed before this merges.

1. **Use the recording’s scheduled times.** `run_recording` currently puts the task arguments into `start_time` and `end_time`. Startup recovery of a still-open recording dispatches that task with `str(now)` as the start, so the event says the recording started at boot. Both emit sites should use `Recording.start_time` and `Recording.end_time`.

2. **Still emit if the final metadata save fails.** The new `recording_end` sits inside the metadata `try`. If `_db_retry` raises, the exception is logged at debug and the event is skipped, even when the file is already on disk. Emit from the in-memory `custom_properties` when the save fails, or log that failure at error.

3. **Emit `recording_end` on cancel.** Cancelling an in-progress recording deletes the row and sends the `recording_cancelled` WebSocket message, then `run_recording` cleans up and returns before the new emit. `recording_start` has already fired, so Connect and plugins never see a matching end. `recording_cancelled` is UI-only. Please emit `recording_end` from `destroy()`, before or as the row is deleted:
   - `status` and `outcome`: `"cancelled"` (not `"failed"`, so a failure alert does not fire on a user delete)
   - `has_file`: false
   - `failure_reason`: null
   - `cancelled_by`: `request.user.username`, plus the user id. `destroy()` already runs as an authenticated admin or DVR manager, for both JWT and API-key callers, so this is on `request.user` before the row is deleted.

   Stop should stay as it is: keep the row, remux, and emit with `status: "stopped"`.

4. **Changelog.** Move this entry to **Changed**. Note that `recording_end` may arrive later than before, that the old `interrupted` boolean is gone, and that cancel now emits `outcome: "cancelled"`.

Connect still strips falsy keys, so `has_file: false`, `remux_success: false`, and `file_size: 0` will not show up on webhooks or plugins. `outcome` and `failure_reason` are the fields subscribers should branch on. A test that the resume path uses the model times, and that cancel emits `cancelled` with `cancelled_by`, would lock this in.

Issue #259, opened 2025-07-23
[Bug]: teletext and subtiles service are remove from stream


www2000:
### Describe the bug

Streams that have subtitles stream and/or teletext
This come due to what i think a ffmpeg bug or -c COPY is only copy all the audio and video streams and remove subtitle/teletext streams 

The fix is simple to add the foaling command line flag to the ffmpeg command:
-c:s copy

### How can we recreate this bug?

1. add a tv channel that include an teletext and/or subtitle stream to distatcharr
2. open the channel in VLC
3. you see a list of subtitle in the subtitle menu and a teletext interface just above the media controls. 

See image how vlc looks like if a subtitle stream and teletext stream are include in the video stream.

<img width="1854" height="1048" alt="Image" src="https://github.com/user-attachments/assets/0a1b2e27-8506-4e91-b960-0e5dc4f3c1c9" />

### Dispatcharr Version

v0.7.0

### Docker Version

Docker version 27.5.1, build 27.5.1-0ubuntu3~24.04.2

### What's in your Docker Compose file?

docker run -d \
  -p 9191:9191 \
  --name dispatcharr \
  -v dispatcharr_data:/data \
  ghcr.io/dispatcharr/dispatcharr:latest

### Client Information

Tested with VLC, kodi and mpv
Teletext are only supported in VLC and kodi, vlc is the best option for testing.

SergeantPanda:
Yeah you are correct from my understanding with FFmpeg if you do not specify it only pulls the first subtitle track. As far as I know, it doesn't remove all subtitle tracks though.

krog:
The default ffmpeg behavior without the map option is to keep only one stream of each type.

You have to use the map option. The easiest way is to use '-map 0', which will select all streams of all types (video, audio, subtitles). 

documentation: https://trac.ffmpeg.org/wiki/Map.

v8eta:
Proposing a shape for this before writing code, since the obvious fix does not work.

**What happens today.** DVB teletext/subtitle streams never reach a recording: the DVR capture maps `0:v?` / `0:a?` only, because the HLS muxer refuses to copy `dvb_teletext` / `dvb_subtitle` (`[webvtt] Exactly one WebVTT stream is needed`) and the Matroska remux refuses them too (`Tag [6][0][0][0] incompatible with codec id`), so mapping them through fails the recording rather than degrading it (that is #1502).

[...]

PR #1689, opened 2026-09-13
fix: add `-c:s copy` to seeded default output profiles

### LINKED ISSUE #1502: Default FFmpeg profile crashes on streams with DVB Subtitles due to '-map 0'
### Describe the bug

When attempting to stream a channel whose input contains a DVB Subtitle track (dvbsub / dvb_subtitle), FFmpeg fails instantly on startup when using standard streaming profiles. This happens because default profiles use -map 0, forcing FFmpeg to attempt mapping all streams. Since no subtitle encoder or passthrough rule is defined for mpegts output, FFmpeg throws an Encoder not found error, causing a broken pipe and leading to a 500 Internal Server Error in Dispatcharr.

What I expected to happen:
Default streaming profiles should either strip subtitle/data tracks (e.g., using -map 0:v -map 0:a? or -sn) or handle subtitle mapping gracefully so unrecognized/unencoded subtitle streams do not crash the FFmpeg process.

Suggested Fix:
Update default FFmpeg profile parameter templates to explicitly map video and audio only (-map 0:v -map 0:a?) or append -sn to drop subtitle streams.

Relevant FFmpeg Log Extract:

Plaintext
Stream #0:0[0x100]: Video: h264
Stream #0:1[0x101]: Audio: ac3
Stream #0:2[0x102](eng): Subtitle: dvb_subtitle (dvbsub)

[sost#0:2] Automatic encoder selection failed. Default encoder for format mpegts (codec none) is probably disabled. Please choose an encoder manually.
[sost#0:2] Error selecting an encoder
Error opening output file pipe:1.
Error opening output files: Encoder not found
Stdin error: [Errno 32] Broken pipe

1. Assign user to output profile = Media Server 
2. Connect an M3U stream that contains video, audio, and a DVB subtitle track (e.g., UK DVB-T/IPTV feeds containing dvb_subtitle).
3. Tune into the channel from a client device.
4. The stream will fail to load, and container logs will reflect the FFmpeg Encoder not found crash on stream #0:2.

### Dispatcharr Version

0.28.2

### Docker Version

Unknown (seedbox provided)

Not using

### Client Information

Tivimate , IPTV Zero Player

v8eta:
## Description

Fixes #1502. Addresses the live-path half of #259 (both already open and discussed at length in #259's thread — this PR is the concrete result of that discussion, not new-issue code).

`core/migrations/0028_add_subtitle_copy_to_default_profiles.py` — a reversible `RunPython` data migration adding `-c:s copy` to both seeded `OutputProfile` rows (`Media Server (AC3 Audio)`, `Web Player (AAC Audio)`). No schema change, no new dependency, one ffmpeg flag.

Neither seeded profile maps a subtitle codec today. Without one, ffmpeg falls back to automatic stream selection, which fails to open the output entirely on any channel carrying a subtitle PID mpegts has no default encoder for — `dvb_teletext`, `dvb_subtitle`, `webvtt`, `subrip`. That's #1502's crash, and the measurements below show it isn't DVB-specific.

## Related Issue

Closes #1502
Part of #259

## How was it tested?

Both hazards measured on a genuinely unpatched instance, 2026-09-13 (full methodology in the linked #259 comment):

**Live-path A/B, real `dvb_teletext` broadcast (AU FTA, shipped default vs. same + `-c:s copy`):**

| Profile | With teletext PID | Subtitle-free control |
|---|---|---|
| shipped default | crash, `Encoder not found` | `h264, ac3` |
| + `-c:s copy` | `h264, ac3, dvb_teletext` | `h264, ac3` — identical |

**Non-DVB subtitle codecs (synthetic sources, mpegts output) — the case this change needed to not make worse:**

| Source subtitle codec | shipped default | + `-c:s copy` |
|---|---|---|
| `subrip` | crash, `Encoder not found` | succeeds — `Stream 2, codec subrip, is muxed as a private data stream and may not be recognized upon reading` (`subtitle:0KiB`) |
| `webvtt` | crash, `Encoder not found` | same — succeeds, unrecognised private data, never a dead channel |

Mechanism, from ffmpeg's own docs (Subtitle Formats table): `DVB teletext` has no muxing support anywhere in ffmpeg — `dvb_subtitle` is fully native (mux/demux/encode/decode). mpegts writes an unrecognised codec as opaque private data; Matroska does a strict codec→tag lookup and refuses. That's the real reason `-c:s copy` is safe here into mpegts but not into Matroska — not a teletext-specific limitation.

tvheadend fixes the identical problem the identical way: their subtitle-passthrough docs say to set "Subtitle codec" to "Copy codec type" on the stream profile — independent validation from a decade-old broadcast PVR, not a novel approach.

**Automated tests:** `core/tests/test_output_profile_subtitle_copy.py`, 5 tests. Forward migration asserts both profiles carry `-c:s copy` with the exact expected parameter string; `locked` is unaffected; reverse migration restores the original string exactly. Written first, watched fail (RED) against a scratch throwaway Postgres before the migration existed, then watched pass (GREEN) after — real Postgres, not SQLite, since an unrelated pre-existing migration (`apps/channels/migrations/0038_add_catchup_fields.py`) uses Postgres-only raw SQL that SQLite's runner can't parse, so `TEST_USE_SQLITE=1` doesn't currently work for anything touching `core`. Not something this PR should fix, flagging in case it's news.

**Manual/production verification:** running live in my own production instance as a local patch since 2026-09-13 (not just tested on a branch) — screenshots of the actual effect are in the #259 comment.

## Checklist

- [ ] Backend: new API endpoints appear correctly in the OpenAPI schema (N/A — no new endpoint)
- [ ] Frontend: ESLint and Prettier pass cleanly (N/A — no frontend change)


Jeremy-McCrindle:
Heads-up on a blocker that won't surface as a merge conflict.

This adds `core/migrations/0028_add_subtitle_copy_to_default_profiles.py`, but `core/migrations/0028_alter_systemevent_event_type.py` already exists on both `main` and `dev` — and both declare `dependencies = [('core', '0027_vlc_play_and_exit')]`. GitHub reports the PR mergeable because the filenames differ, so this is easy to miss, but Django ends up with two leaf nodes on `core`.

Reproduced on Django 6.1.1 using this PR's migration file verbatim:

```
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph:
(0028_add_subtitle_copy_to_default_profiles, 0028_alter_systemevent_event_type in core).
To fix them run 'python manage.py makemigrations --merge'
```

Renaming to `0029_add_subtitle_copy_to_default_profiles` and changing the dependency to `('core', '0028_alter_systemevent_event_type')` is sufficient — `migrate` then applies cleanly. I also checked the data operation itself behaves, forward and reverse:

| profile | after 0029 | after reversing to 0028 |
|---|---|---|
| Media Server (AC3 Audio), `locked=True` | `-c:s copy` present | removed |
| Web Player (AAC Audio), `locked=True` | `-c:s copy` present | removed |
| a user-made `locked=False` profile | untouched | untouched |

Happy to push a rebased branch if that's useful, but it's a one-line change and this is your PR — I didn't want to open a competing one.

One data point for the "is it worth it for the 98% of streams without subtitles" question from #259: the cost is nil. Running the current seeded `Media Server (AC3 Audio)` parameter string against a subtitle-free MPEG-TS, with and without `-c:s copy`, produces byte-identical output (`cmp` clean, 565880 bytes both, ffmpeg 9.0.1). So it's inert where there are no subtitles, and it turns #1502's hard `Encoder not found` failure into a working stream where there are.

v8eta:
Good catch, and thanks for testing it rather than just flagging it.

Pushed — the migration is now `0029_add_subtitle_copy_to_default_profiles` depending on `('core', '0028_alter_systemevent_event_type')`. I rebased onto current `dev` while I was in there, so the branch is a single commit again rather than 14 behind.

You're right that this doesn't surface as a conflict, and it bites in practice. I run this patch against the rolling `:dev` image on my own box, and on 14 Sep `0028_alter_systemevent_event_type` arrived and every `migrate` started failing with exactly the error you quoted. I papered over it locally with an empty merge migration whose header note says to delete it once this lands upstream — so your fix is the one I should have made to begin with. One note for anyone else running `:dev` with this applied: `migrate core` with no target is the safe invocation, since pinning `core 0028` is ambiguous while two of them exist.

On the cost question, your `cmp` result matches what I measured on 13 Sep and posted in #259 — `-c:s copy` is inert with no subtitle PID, and it's the difference between a working stream and a hard `Encoder not found` when there is one. I also checked it doesn't create a *new* failure mode for non-DVB codecs: synthetic `subrip` and `webvtt` sources both crash without it and both mux cleanly with it. They survive only as `bin_data` rather than as readable subtitle tracks, so that half isn't a feature — it just isn't a dead channel any more, which is what matters for a default.

Appreciate you checking forward and reverse, and confirming `locked=False` profiles are untouched — that matches my testing.

core/migrations/0029_add_subtitle_copy_to_default_profiles.py
```diff
@@ -0,0 +1,99 @@
...
+            '-c:a aac '
+            '-b:a 192k '
+            '-ac 2 '
+            '-c:s copy '
+            '-max_muxing_queue_size 4096 '
+            '-flush_packets 1 '
+            '-mpegts_flags +pat_pmt_at_frames+resend_headers+initial_discontinuity '
+            '-f mpegts pipe:1'
+        ),
+    )
```
sethwv: The migration overwrites customized profile parameters and cannot restore them on rollback. Please consider adding `-c:s copy` to the existing value instead of replacing the full string.

It's very unlikely that a user has a customized profile here, but not impossible since admin tokens can do this via the API. Mostly a nit, but logging regardless just in case.

sethwv (commented):
I think in general this looks positive, and in most cases this addition should be benign. Adding one note for your consideration before I give this one any sign-off.

Thanks for submitting!

PR #1710, opened 2026-09-15
Preserve radio/TV stream type through to XC and M3U output

### LINKED ISSUE #1683: [Bug]: RADIO channels are not properly processed and lose their radio property tags
### Describe the bug

Radio streams lose their type on the way to any client. At least one of my XC providers correctly tags radio entries upstream with stream_type: "radio_streams", and this survives into Dispatcharr's own storage (Stream.custom_properties). But both output paths discard it:
·	Classic Xtream API (apps/output/views.py, _xc_channel_entry, line 806): hardcodes "stream_type": "live" for every entry in xc_get_live_streams, regardless of what's in the underlying stream's custom_properties.
·	M3U output (apps/output/views.py, generate_m3u, line 379): the EXTINF line only ever sets group-title, never a radio="true" attribute.

So even though the raw signal exists and is already stored, no client can see it. This matters for Kodi's PVR IPTV Simple Client, which has native Radio support and switches an entry into its Radio section specifically based on the M3U radio="true" attribute, and for XC-based clients (e.g. IPTV Smarters) that read stream_type directly from get_live_streams. Right now radio channels can only be told apart by group-title text convention (e.g. "RADIO" in the category name), which isn't a structural signal any client acts on automatically.

1.	Sync an XC-type M3U account whose provider tags radio streams with stream_type: radio_streams (confirmed present for at least one of mine).
2.	Check Stream.custom_properties for one of those streams in the Django shell/DB, it's already there.
3.	Hit /player_api.php?...&action=get_live_streams for that same channel, stream_type comes back "live".
4.	Fetch the M3U output for the same channel, no radio="true" anywhere on the EXTINF line.

### Dispatcharr Version

0.30

### Docker Version

Docker version 29.7.0, build c1eba93

Standard AIO config, not relevant for this bug

### Client Information

Kodi (PVR IPTV Simple Client, uses M3U radio="true") and any Xtream-API client reading stream_type from get_live_streams (e.g. IPTV Smarters style clients).

dillardblom:
## Description

Radio streams are currently exposed to clients as regular TV channels: the M3U output never sets the `radio="true"` EXTINF attribute Kodi's IPTV Simple Client (and others) rely on to route entries into a Radio section instead of TV, and the XC API always hardcodes `"stream_type": "live"` regardless of the actual content.

Adds an `is_radio` field on both `Stream` and `Channel`, populated at both M3U import paths (`process_xc_category_direct` and `process_m3u_batch_direct` in `apps/m3u/tasks.py`) from the real Xtream `stream_type` field (`radio_streams` vs `live`/anything else), backfilled via migration for existing data, and rolled up from Stream to Channel the same way `is_catchup` already is (`rollup_channel_catchup_fields` pattern, plus a `ChannelStream` signal for incremental updates). M3U output now sets `radio="true"` when the channel is radio, and the XC API returns `"stream_type": "radio_streams"` instead of hardcoding `"live"`.

Verified this isn't just a made-up field: pulled the real `custom_properties` for actual PROVIDER streams and confirmed `stream_type: "radio_streams"` is a genuine provider-populated value, distinct from `created_live` (a provider-side bookkeeping value for self-assembled loop channels, unrelated to radio, confirmed by checking actual category breakdowns — a group literally named "MUSIC & RADIO" had zero `radio_streams` entries and was excluded from the fix on that basis).

## Related Issue

Closes #1683

## How was it tested?

- Restored a full production database dump (real M3U accounts, real Xtream `stream_type` values from 3 independent commercial providers) into an isolated throwaway container to verify the migration backfill against real data, not fixtures: `Stream.is_radio` landed at 399 true / 13634 false, `Channel.is_radio` at 90 true / 531 false, matching the expected real counts exactly.
- Confirmed M3U output includes `radio="true"` and XC API (`get_live_streams`) returns `"stream_type": "radio_streams"` for a known real radio channel on that restored instance.
- End-to-end tested against a real Kodi install (PVR IPTV Simple Client) pointed at that same restored instance via the **M3U URL** path: radio channels now show up in Kodi's Radio section and play back correctly as audio.
- Not yet tested end-to-end with a client logged in via the **XC-API** login mode specifically (only verified the raw JSON field value with curl, not a real XC client) — don't have a stable client on hand right now that supports both XC login and radio. Flagging this honestly rather than implying full coverage; the underlying code path (`_xc_channel_entry`) is the same one that already serves `tv_archive`/`stream_id` etc. correctly for every other field, so I'd expect it to work, but it hasn't been confirmed with a real client the way the M3U path has.
- Full backend test suite (132 tests, including the new radio-specific tests) run against the branch rebased on current `dev`, using the project's own CI recipe (`base-dev` image + `ci_bootstrap_backend.sh`, real Postgres/Redis, not a mock): all passing.

## Checklist


dillardblom:
Ran an independent review of this branch (Codex) before it goes further, both technically and functionally. Found and fixed three real issues, pushed just now:

- **Migration backfill didn't match the runtime rollup's active-account rule.** `rollup_channel_radio_flag` (the ongoing rollup) deliberately excludes streams on disabled M3U accounts, but the one-time migration backfill for existing installs didn't apply the same filter. A channel backed only by a disabled account would come out of the upgrade incorrectly flagged as radio, and stay that way indefinitely since disabled accounts are never refreshed. Fixed to join and filter on `is_active` the same way. Also added direct tests for both backfill functions, since migration 0039 had none at all before this.

- **Standard M3U `radio` attribute was read case-sensitively.** Every other EXTINF attribute in this parser (`tvg-id`, `group-title`, etc.) goes through a case-insensitive lookup helper; the new `radio` attribute used a plain dict `.get()` instead, so a playlist spelled `RADIO="true"` would have silently imported as TV. Fixed to use the same helper as everything else.

- **Flagged but not changed:** `is_radio` is writable via `StreamSerializer` without triggering a channel-level rollup on a direct field edit (the rollup signal only fires on `ChannelStream` link changes, not on the `Stream` fields themselves). Checked, and this is an existing gap `is_catchup` already has too, not something new here. Fixing only `is_radio` would make it behave inconsistently with the field it's directly modeled on, so leaving both as-is for now and tracking a combined fix separately rather than bundling it into this PR.

Full suite (886 tests across `apps.channels`, `apps.m3u`, `apps.output`) green after the fixes.

